PR #1, opened 2020-11-17
Reuse task registration

Twixes:
Since apparently both base and EE task handling is the same, this ensures consistency and removes code redundancy by using the calling `worker.register` in a loop based on the only difference: task names.


Twixes:
PR/issue #1, how glorious, let there be many more

Twixes:
#4

PR #3, opened 2020-11-17
Allow for optimized passing on unchanged event

Twixes:
This allows plugins to not have to pass on the entire event object if they want to leave it unchanged – they can just return undefined (or in other words not return anything), avoiding all the extra assignment/memory work, possibly slightly improving performance.

Would be nice to do: some benchmarking.


mariusandra:
We should probably benchmark this first... and if the performance difference is minimal, maybe pass on this? I'm saying this only because it's a breaking change that unfortunately won't make it into the next release... and we have this line in the docs since today: 
https://github.com/PostHog/posthog.com/pull/677/files#diff-a4573da2e409b5a9fee5e55915df6d80579a66534a2dce6c24a9bfd22fe7119dR36

That said, I agree that if we do breaking changes, now is the time to do them. We might have more than one breaking change for the next release (such as plugin.json -> package.json, etc)...

mariusandra:
skip for now, vm gets a reference to the actual object, so no performance penalty

PR #14, opened 2020-11-18
Multiprocess with PM2

### LINKED ISSUE #6: Enable multiprocessing
Currently the server can only use 1 thread, even if the machine has for instance 8 cores and could handle 8 server threads at once, greatly improving performance under load.

Twixes:
This enables straightforward multiprocessing based on `WEB_CONCURRENCY` or number of CPU cores with [PM2](https://pm2.keymetrics.io/). The processes are completely separated which ensures safety, but on the other hand increases memory usage. Resolves #6.


mariusandra:
Hey, cool! Perhaps it would make sense to extract this from plugin-server and to add it to the posthog repo directly? 

Currently we run [this script](https://github.com/PostHog/posthog/blob/master/bin/plugin-server), which runs "yarn start" in this [package.json](https://github.com/PostHog/posthog/blob/master/plugins/package.json) to get the server started. 

We could add `pm2` to that package.json and change the start script there to run via pm2 --no-daemon, while making sure all the config params are passed properly.

Twixes:
Moved changes to PostHog/posthog#2433.

PR #18, opened 2020-11-18
Base plugin config on package.json and add PostHog version compat check

### LINKED ISSUE #12: Minimum PostHog version for plugins
I'd like to specify a minimum posthog version in my plugins to make sure all the required functionality is there. For example to say "my plugin requires attachments, so I need at least PostHog 1.17.0". It might also be good to pin to an exact version of posthog, since we might still have some slightly breaking changes.

The `engines` field in package.json seems to be a great place for this.

https://docs.npmjs.com/cli/v6/configuring-npm/package-json#engines

### LINKED ISSUE #13: Move from `plugin.json` to `package.json`
Now that you can install plugins from npm, a `package.json` file is required. It duplicates most of the fields from `plugin.json` (`name`, `description`, `main` as is, `url` --> `homepage`), with the exception of `lib` and `config`.

We can just remove `lib` and keep all code in `main` / `index.js`. Most plugins will be compiled down to one `.js` file anyway, and magically including all code in `lib.js` inside `index.js` doesn't lead to a good experience while developing (no autocomplete for functions in lib.js).

For `config`, we can just use [config](https://docs.npmjs.com/cli/v6/configuring-npm/package-json#config) in package.json.

Twixes:
This makes `package.json` the base for plugin config, loading `plugin.json` on top of it. Also adds a PostHog semver compatibility check (getting version range from `pluginConfig.engine.posthog`). Resolves #12, resolves #13.


Twixes:
Since we haven't yet decided how to handle compatibility, and this got somewhat stale, I'll close it.

PR #24, opened 2020-11-26
Add integrated ingestion server

Twixes:
#23 should be merged before reviewing this!

This is the base event ingestion server (no request processing in this PR). It listens on all paths the same (since we'll defer to nginx/whatever to actually restrict it to `/e`, `/capture` etc.). Purposefully vanilla Node instead of Express for maximum performance and no bloat.


mariusandra:
Hey, cool!

Few points to add:
- I'd rename the server from "ingestion" to just "http" or something like that. Very soon (in the cosmic timescale) we will add support for webhooks in plugins and we should reuse this server then. I guess we'll have hooks like `/w/:plugin_id/:hook_uuid` or something like that, which will get directed to the right service.
- Same for the name of the CLI vars, etc.
- I'd start the HTTP server by default, and add a `--no-http` option to disable it. Maybe also a `--no-worker` option to disable the celery part?
- For extra performance, perhaps check out [fastify](https://www.fastify.io/) - it seems to be [faster than nodejs](https://medium.com/javascript-in-plain-english/fastify-express-benchmark-4c4aebb726d6) (medium link, open incognito, [other benchmarks](https://github.com/the-benchmarker/web-frameworks)) and provides an easy API.

Twixes:
Damn, Fastify actually looks pretty great! Around built-in `http` performance _and_ it has official [Postgres](https://github.com/fastify/fastify-postgres) and [Kafka](https://github.com/fastify/fastify-kafka) integrations.

mariusandra:
That's cool! Both of these adapters seem pretty practical!

Somewhat related and possibly useful context: In the beginning of the year I took a deep dive into the nodejs postgres ORM world... and its a mess. I think nothing has really changed there, so best to save time and stick to raw SQL for now.

Twixes:
Alright, very fair points, addressed all of them (besides `--no-worker`). @mariusandra

mariusandra:
I refactored the CLI params from camelCase to dash-case, as this is how I remember most other CLI apps working.

PR #25, opened 2020-11-26
Port Python get_event endpoint over to plugin server

Twixes:
This is a (WIP) port of the `get_event` view function from Python to TypeScript, more precisely only its `ee` branch. The original runs our capture endpoint in the Django app but isn't compatible with plugin processing, this should allow for compatibility.


mariusandra:
With regard to naming things, there are two parts of the stack that can both be called "ingestion":
- `/e/` --> some queue (and then before plugins) .... aka "capture"
- some queue after plugins --> database ... aka "process_event"

The JS server should be able to do both ends, yet I think the `process_event` step is more important for now, since the first step is handled by Django. Even in EE we write the events to a Kafka topic in `capture`, which the plugin server could just read from. There's however no way to get events after plugin processing back into Django for the final `process_event_ee` step...

(That said the `capture` step as started in this PR is also very important to get done. If for nothing else, we can at least do better benchmarks when that's in place)

Twixes:
This is super stale now.

PR #26, opened 2020-11-26
Add eslint rule for curly brackets

mariusandra:
Hey, similar to the rule that went in to `posthog` yesterday, this just makes code more readable IMO


mariusandra:
I had it like this before, but got an error like this:
image

I removed the [.clinic](https://github.com/clinicjs/node-clinic) folder and will merge, since I agreed that `.` is better than `src`.

PR #27, opened 2020-11-26
Slight vm global refactor

mariusandra:
Refactors this a bit and adds `module.exports` support for commonjs modules.

Still needs a lot of tests to make sure it does its job.

### DIFF
--- a/src/vm.ts
+++ b/src/vm.ts
@@ -28,12 +28,12 @@ export function createPluginConfigVM(
     )
     vm.run(
         `
-        const module = {};
+        const module = { exports: {} };
         const exports = {};
         const __pluginLocalMeta = { global: {} };
         const __pluginMeta = { ...__pluginHostMeta, ...__pluginLocalMeta };
         const __getGlobalWithMeta = (key) => {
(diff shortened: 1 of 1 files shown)


src/vm.ts
```diff
@@ -29,25 +26,28 @@ export function createPluginConfigVM(
...
+        `
+        const module = { exports: {} };
+        const exports = {};
+        const __pluginLocalMeta = { global: {} };
+        const __pluginMeta = { ...__pluginHostMeta, ...__pluginLocalMeta };
+        const __getGlobalWithMeta = (key) => {
+            const method = exports[key] || module.exports[key] || global[key];
+            if (!method) { return null };
+            return (...args) => method(...args, __pluginMeta)
+        }
```
Twixes, line 39: It may be just me being thick, but I don't exactly grasp what all these "metas" do and how this works 🤔 Some comments here could be enlightening

mariusandra: 👍 . I tried making it a bit more clear.

Twixes: Thanks, that'd definitely clearer!

Twixes (approved):

PR #28, opened 2020-11-27
Inline node celery, clean up on signals

### LINKED ISSUE #20: Plugin server doesn't stop on CTRL+C
The process keeps on running in the background.

mariusandra:
Few changes with this PR:

- Inline node-celery (MIT licensed), strip it to only what we need and fix all old typescript types. I removed all `ampq` code, and more importantly, removed the `AsyncResult` code, which you could not opt out. The original implementation would store a "event processed" notification in Redis for 24h after each task was completed. We had a similar situation with the python celery recently... and users complaining of running out of memory. Removing this in python-celery took our own Redis memory usage from 10GB to 10MB.
- Implement graceful shutdown on SIGTERM, SIGINT (ctrl+c), SIGHUP. Previously the app would just leave a process hanging in the background. Closes #20 
- Implement graceful shutdown for Node Celery. This was just not implemented.
- Node-celery was creating too many Redis connections. After the refactor and cleanup I got this down from 4 to 2. This will make heroku users happy.

Still WIP as I want to run a full E2E test on this, yet I have managed to royally bork up my python setup :/.

New tech debt:
- Add tests for celery (the [old ones](https://github.com/actumn/celery.node/tree/master/test) are using mocha & chai)

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -5,6 +5,7 @@ import { setupPlugins } from './plugins'
 import { startWorker } from './worker'
 import schedule from 'node-schedule'
 import Redis from 'ioredis'
+import { startWebServer, stopWebServer } from './web/server'
 
 const defaultConfig: PluginsServerConfig = {
     CELERY_DEFAULT_QUEUE: 'celery',
@@ -13,6 +14,9 @@ const defaultConfig: PluginsServerConfig = {
     REDIS_URL: 'redis://localhost/',
     BASE_DIR: '.',
     PLUGINS_RELOAD_PUBSUB_CHANNEL: 'reload-plugins',
+    DISABLE_WEB: false,
+    WEB_PORT: 3008,
+    WEB_HOSTNAME: '0.0.0.0',
 }
 
 export async function startPluginsServer(config: PluginsServerConfig): Promise<void> {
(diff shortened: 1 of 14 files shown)


src/server.ts
```diff
@@ -47,10 +57,22 @@ export async function startPluginsServer(config: PluginsServerConfig): Promise<v
...
     // every 5 sec set a @posthog-plugin-server/ping redis key
-    schedule.scheduleJob('*/5 * * * * *', function () {
+    const job = schedule.scheduleJob('*/5 * * * * *', function () {
         redis.set('@posthog-plugin-server/ping', new Date().toISOString())
         redis.expire('@posthog-plugin-server/ping', 60)
     })
-
     console.info(`✅ Started posthog-plugin-server v${version}!`)
+
+    for (const signal of ['SIGINT', 'SIGTERM', 'SIGHUP']) {
```
Twixes, line 66: Actually could we put this in its own function like `setUpExitHandling(handler: (() => (void | Promise<void>))): void` ? Because the web server does identical `process.on(signal, handler)` with SIGINT, SIGTERM and SIGHUP and

Twixes: Ah, nvm, that's integrated!

mariusandra: I actually moved `stopWebServer` also here.

Twixes (approved):
Looks good!

PR #29, opened 2020-11-30
Tests and other improvements

mariusandra:
Changes:
- Adds test for `vm.ts`
- Deprecates `lib.js`
- Cleans up code in the vm
- Other misc changes here and there


src/__tests__/vm.test.ts
```diff
@@ -0,0 +1,436 @@
...
+        },
+    }
+    const processEvent = prepareForRun(mockServer, pluginConfig.team_id, pluginConfig, 'processEvent', event)
+    expect(processEvent).toBeDefined()
+    await processEvent!(event)
+
+    expect((fetch as any).mock.calls[0][0]).toContain('http://localhost/e/?ip=1&_=')
+    expect((fetch as any).mock.calls[0][1].body).toContain('data=')
+    expect((fetch as any).mock.calls[0][1].body).toContain('&compression=lz64')
+    expect((fetch as any).mock.calls[0][1]).toMatchObject({
```
Twixes, line 430: I think this would be better as `fetch as unknown`.

Twixes (approved):
Looks good, just a typing comment because linter complains. And nitpicking I kind of dislike the dirname `__tests__`… I'd go for just `tests`, or even put the `.test.ts` files side by side with the `.ts` ones they're testing.

src/__tests__/vm.test.ts
mariusandra: TypeScript is even less happy with that:
image

And that's a TS error not ESLint suggestion.

I added a few lines to `eslintrc` to ignore `any`s under tests.

Also, these variables are technically `known`, I'm just ignoring them for convenience... and thus `any` seems like a better fit.

mariusandra:
Thanks! I got rid of the `any` issue with some extra eslint rules. I don't think we need to type out the exact structure of jest mocks in order to run the tests. I'm for anything that makes it easier to write tests!

Regarding the `__tests__` folder, I'd actually argue for keeping it. We have 4 options, each with some issues:

- `worker.test.ts` file in the same folder -- hard to see what is tested if some files lack test files and has problems if we want some global tests that are not connected to any one file. The folders will get huge as well.
- `src/tests/worker.ts` -- the folder `tests` disappears into the other folders under `src` (`celery`, `extensions`, `tests`, `web`) and hard to separate its importance. 
- `src/__tests__/worker.ts` -- very clearly the `__tests__` folder is separate from all the other folders, towering above the other folders. You'll easily see inside each folder if it has tests or not. Can't be worse than `__init__.py` :D. 
- `tests/worker.ts` -- escaping `src` and putting all tests inside `tests` is also fine for smaller projects that don't have a lot of nested folders. However soon we'll be replicating the /src folder structure inside another folder. 

The last two options are the only ones I find make sense... and of the two I like `__tests__` better because it brings tests closer to the actual code that is being tested.

I'd also argue for the files to be `worker.test.ts` and not `worker.ts` no matter what folder they are in, because that makes it easier when having many tabs open in your IDE.

Dunno :). Let me know if you feel strongly enough towards changing it.

Twixes:
This is okay.

PR #30, opened 2020-12-01
Worker & Util Tests

mariusandra:
I'll rebase and update the eslint file once #29 is in.

Changes:
- Adds `ioredis-mock` and runs the tests through that.
- Adds E2E(ish) check for celery workers though `ioredis-mock`. 
- Adds tests for compression utils


jest.setup.redis-mock.js
```diff
@@ -1,8 +1,19 @@
 jest.mock('ioredis', () => {
-    const mPool = {
-        get: jest.fn(),
-        set: jest.fn(),
-        disconnect: jest.fn(),
+    const Redis = require('ioredis-mock')
+    if (typeof Redis === 'object') {
```
Twixes, line 3: Why does this check type? Won't it always be the same?

src/__tests__/utils.test.ts
```diff
@@ -0,0 +1,54 @@
...
...BxaWC+sHVOZ4CGV0rvB/r5G9W8HPmxU4e8c7xcC/grklmCzHJJwSesLqmoHJenAAbc0b2jyIa1S8DU00BnNxmIorNbREWx/FTa1lVENu7QfDpbYzPODn9CJg1sukujgEd45EYazCKeOw0eY/1T5WUiYfz/Crit1UYqNwHbhrrWKObi3tuu2UfGRil3lf5bkC5ZiAeDpZ99jT+V/p9nayf9Oyz3Z5/9LgN7wdoqXfPsUbLPDzd4utoPtaGouEkNUKpOnR0cLvDbn83rIV0cTLtUFXxx9WC6iMhQs0/tSy+s3BGWKxZXmNHfPc86BwBkRDnw7BVrcWV8X4ivCjFx1MhVY3O0aWbSrBarIMcQWyWtjsq0HW8FdLYk6kFA+EiDT34y44VF3meHRl/l0scFHNCZ5ojRpTn7d4AX9d84EjZAQk0TSMun0zxfWGFX5H3jd/tCrr6JnsPFE/jda7nv9f2t//r8MvIHd5LV0Vi/5oiCAoUB5PMClpIDnL5GYtnMi9ZuTgDueC+C3acklX1vWmzcw1a0E3DACZZmwrEYdzpl5cqJgl8w2ZLj/dFpWbM06eFhjqHk+M89byF5wQyzKp7GEYVuAtvWTlkFcBgOonhMFzbhkios7/TSGRpR+uHroVG/gG4f8ggo/ZEXY34QJ+6n2SLHb4jvUvlYKinc04+YpXJdcld+lD+9N9rXVWkeb43RuQtWVCvuAmIRUx6lr41qrOaavi8fSqJjOqiJVDdQNbmcyT6g8ta6vrzW/VZbHR4v9f1/uUXKR8DlJtmov4miqvdCFUomcOhvCTtVe44vqbcec22vCfVlf73UsVrsOQb7ZLr93Bn3ZsO7/iu+zNIBP3/86u/1fu9Ha1/+XgAf9X1XaNgn+zXZDd0OFLJs5t45/nmjzjIJHm7lCUG5aN0WlabZouOTwo+0JwcUppBw0AWRGQxYzGv1ow1dfAf0FS2JDV4l7o21T9bcUlo0dVrGd+oaYrz+hxtW1hrUpkuM8mTpVdaXm/300YujPCtw8X2xF9ulV9YhJmVO5MbrkK6rPyk/U80ZQEuH6frgzrfIfb5j6Rk+FCD/7Hnsq/xvH7+V/y23v8/8loMoTQViCV58eX62IyUDZqRJGkfkVi9QS0e0SJemKVXefCoXSCf17zhV9cNjbGV6pVKWg0XS/tBvSHvawhz38MeE//YvvUAAoAAA='
```
Twixes, line 6: TBH I'd like to see a comment explaining the contents of these.

src/vm.ts
```diff
@@ -1,5 +1,5 @@
 import { VM } from 'vm2'
-import fetch from 'node-fetch'
+import * as fetch from 'node-fetch'
```
Twixes, line 2: What do the introduced asterisks change? 🤔

jest.setup.redis-mock.js
mariusandra: I added a link describing where I got the code from... And fixed the other link.

src/__tests__/utils.test.ts
mariusandra: Added something

src/vm.ts
mariusandra: These go together with the `esModuleInterop` change in `tsconfig.json`. If we're writing typescript, might as well write real typescript, and this is what that looks like. Also, everybody including React is moving away from default exports, so we might as well adapt.

src/utils.ts
```diff
@@ -1,7 +1,7 @@
 import { Readable } from 'stream'
-import tar from 'tar-stream'
-import gunzip from 'gunzip-maybe'
-import AdmZip from 'adm-zip'
+import * as tar from 'tar-stream'
+import * as AdmZip from 'adm-zip'
+const gunzip = require('gunzip-maybe')
```
Twixes, line 4: Could we maybe use [Node's built-in zlib](https://nodejs.org/api/zlib.html) (`zlib.unzip()`/`zlib.gunzip()`) here?

mariusandra: I think this was a bit out of scope... but done!

mariusandra: Nice catch, btw!

PR #34, opened 2020-12-02
Port Python process_event_ee over to plugin server

Twixes:
This is a port of the the `process_event_ee` pipeline from Python to TypeScript. Designed to work with PostHog/posthog#2605, it does what the original does, just by consuming events from Kafka topic `events_write_ahead_log`, allowing us to run plugin processing right before ingestion into ClickHouse (via Kafka as well).


mariusandra:
Hey! This worker test error is pretty awkward. I'm not sure what's causing it and can't reproduce it myself. Hopefully #58 will make this better.

Some general comments though. 

1) This PR is getting quite big. Could you split it up in two parts: kafka & process_event. 
2) I'd love to get the kafka part ready, reviewed and merged tomorrow, hopefully even with some tests. Then I'd put this live on app, reading from the WAL and just processing random hardcoded plugins (`reverse-strings-plugin` or whatever) to see if the nodejs service holds up. The result of those plugins would just be discarded... or passed on to some other kafka topic without any ingestion.
3) The code shouldn't actually work if I read it correctly? Inside this server we have the main thread and worker threads that process tasks. Currently the main thread does not start `setupPlugins`, so it does not know what plugins are in the DB. This is by design to not waste resources. The worker threads call `setupPlugins` before launching to init all the plugins. The code in this `ingestion-save` branch runs the kafka consumer only in the main thread and calls `runPlugins` in it to process events. There are no plugins to run.
4) Ideally kafka should run in the main thread, passing messages/tasks to the piscina worker threads and waiting for them to respond. An alternative option is to run kafka inside each thread itself, but that feels quite inefficient. I think there should be just one pipe that gets the messages. It's possible with Celery, so why not with Kafka.

src/worker/queue.ts
```diff
@@ -48,3 +51,66 @@ export function startQueue(
...
+function startQueueKafka(
+    server: PluginsServer,
+    processEvent: (event: PluginEvent) => Promise<PluginEvent | null>
+): Queue {
+    const eventsProcessor = new EventsProcessor(server)
+
+    eventsProcessor.kafkaConsumer.on('ready', () => {
+        eventsProcessor.kafkaConsumer.subscribe([KAFKA_EVENTS_HANDOFF])
+        // consume event messages in batches of 100 every 100 ms
+        setInterval(() => {
```
mariusandra, line 64: This limits us to max 1000 messages/sec, which will not be enough for app... and is way below what the plugin server could handle. So we need something more robust here. 

Celery uses [something like this](https://github.com/PostHog/posthog-plugin-server/blob/4037c91f8b64083ef6a99f7ff34bab8cfdb332ab/src/celery/broker.ts#L104), where it just asks for the next message when it can.

The way `piscina` currently works is that we spawn `WORKER_CONCURRENCY` (e.g. 8) threads that can handle `TASKS_PER_WORKER` (defaults to 10) async tasks. When 8 * 10 threads are being processed, the rest get put into piscina's queue.

Currently if celery produces tasks fast enough and piscina remains busy, the queue will grow endlessly. This is a problem since if the server then crashes (e.g. out of memory), we lose all the events in the queue. The correct approach would be to only ask celery for another task if we have a worker that can handle it. Basically in this example if there are more than 80 promises pending with piscina, we shouldn't ask for a new task with celery (instead of the current `process.nextTick` logic).

The same logic should be applied to receiving a batch of messages from Kafka as well.

There's [some piscina code here](https://github.com/piscinajs/piscina#backpressure) regarding maximum queue sizes that could help, but AFAIK that code waits until it has processed all events (queue 0 and all workers idle) before then asking for the next event from that stream, so it's not the full solution. We should always keep our workers busy.

Twixes: 100% agree that 1k events/s is too low. Well, we can just consume messages one at a time as they come ([node-rdkafka "flowing mode"](https://github.com/Blizzard/node-rdkafka#standard-api-1)), but then batching is more work on our side (and possibly slightly less performant). We can also up batch size and consumption frequency to some arbitrary amounts e.g. 10k events every 50 ms (so 20k events/s). I believe batching is something we're not actually using yet (not in the Celery pipeline and also not here), but the aim is to make use of it.

src/utils.ts
```diff
@@ -230,14 +230,13 @@ export class UUIDT extends UUID {
...
+        const series = UUIDT.currentSeriesPerMs.get(unixTimeMs) ?? 0
         if (UUIDT.currentSeriesPerMs.size > 10_000) {
             // Clear class dict periodically
             UUIDT.currentSeriesPerMs.clear()
         }
-        const nextSeries = typeof series === 'number' ? (series + 1) % 65_536 : 0
-        UUIDT.currentSeriesPerMs.set(unixTimeMs, nextSeries)
-        return nextSeries
+        UUIDT.currentSeriesPerMs.set(unixTimeMs, (series + 1) % 65_536)
+        return series
```
Twixes, line 239: This change is to start series from 0 instead of from 1.

src/worker/queue.ts
```diff
@@ -3,23 +3,24 @@ import * as Sentry from '@sentry/node'
...
+export async function startQueue(
     server: PluginsServer,
     processEvent: (event: PluginEvent) => Promise<PluginEvent | null>,
     processEventBatch: (event: PluginEvent[]) => Promise<(PluginEvent | null)[]>
-): Queue {
+): Promise<Queue> {
     const relevantStartQueue = server.KAFKA_ENABLED ? startQueueKafka : startQueueRedis
-    return relevantStartQueue(server, processEvent, processEventBatch)
+    return await relevantStartQueue(server, processEvent, processEventBatch)
 }
```
Twixes, line 17: Awaiting start*Queue to avoid unhandled promise rejections.

src/idl/protos.d.ts
```diff
@@ -0,0 +1,248 @@
+import * as $protobuf from 'protobufjs'
```
Twixes, line 1: This file, `protos.d.ts`, is compiled from `.proto` definitions. You should disregard it.

src/idl/google/protobuf/timestamp.proto
```diff
@@ -0,0 +1,147 @@
+// Protocol Buffers - Google's data interchange format
```
Twixes, line 1: This file, `timestamp.proto`, is copied verbatim from the main repo. You should disregard it.

src/idl/events.proto
```diff
@@ -0,0 +1,15 @@
+syntax = "proto3";
```
Twixes, line 1: This file, `events.proto`, is copied verbatim from the main repo. You should disregard it.

Twixes:
Relevant info in https://github.com/PostHog/posthog/pull/2898#issuecomment-759153687.

src/utils.ts
```diff
@@ -267,6 +266,15 @@ export class UUIDT extends UUID {
     }
 }
 
+export function castTimestampOrNow(timestamp?: DateTime | string | null): string {
```
Twixes, line 269: Compare original: https://github.com/PostHog/posthog/blob/master/ee/clickhouse/models/util.py#L70-L80.

src/ingestion/process-event.ts
```diff
@@ -1,30 +1,466 @@
-import { DateTime } from 'luxon'
-import { PluginsServer, Properties } from 'types'
-import { UUIDT } from '../utils'
 import { PluginEvent } from '@posthog/plugin-scaffold'
```
Twixes, line 1: Compare originals: https://github.com/PostHog/posthog/blob/master/ee/clickhouse/process_event.py and https://github.com/PostHog/posthog/blob/master/ee/clickhouse/models/event.py.
This file is essentially these two Python modules (combined because of the helpers' need to access `EventProcessor.db`).

src/idl/protos.d.ts
mariusandra: Should this be in git then? Normally you tend to avoid putting compiled stuff in to prevent polluting git with huge files. Instead we could put a "yarn compile" step in "yarn start"...?

mariusandra (commented):
I'll test all of this thoroughly later today. One small inline comment for now... and then there's this. We need tests 😅 . This code will process billions of events monthly for paying customers, so "testing in production" is really not an option here. At the least we should have all the same tests that run on the python codebase also run this code. I guess having the same `.py` test-suite running on this is probably inconvenient (if possible, go for it!), so at least the spirit of each python test should be captured in some tests here.

src/idl/protos.d.ts
Twixes: It's easier for development when it's available compiled right away, but now that you pointed this out, I realized this can be compiled in `yarn prepare`, so it should actually be seamless for the developer, without committing the file. 🎉

src/worker/queue.ts
Twixes: Actually not awaiting, instead handling start error inside start*Queue with `.catch`.

Twixes:
100% agree on tests, working on that now. Was just thinking that the rest ot the code can be reviewed before that.

mariusandra:
Absolutely! I'm still stuck behind other stuff (and a few calls coming up), though I'll still do a first full review today. Have fun testing! :D

mariusandra:
Still haven't done a very thorough review, but from the call some points to look at:
- tests 🎉 
- django uses signals to sync postgres persons <-> clickhouse persons... and maybe somewhere else. check this
- check if something changed in the main repo in the process_event_* functions
- move the ingestion into a piscina worker thread

src/config.ts
```diff
@@ -77,5 +89,5 @@ export function overrideWithEnv(
             }
         }
     }
-    return newConfig
+    return newConfig as PluginsServerConfig
```
mariusandra, line 92: nit: is this cast needed?

src/extensions/posthog.ts
```diff
@@ -1,79 +1,81 @@
...
-                $lib: 'posthog-plugin-server',
-                $lib_version: version,
-                ...properties,
-            },
+    } else {
+        // Sending event to our Redis>Postgres pipeline
+        const client = new Client(server.redis, server.PLUGINS_CELERY_QUEUE)
+        sendEvent = async (data) => {
+            client.sendTask(
+                'posthog.tasks.process_event.process_event_with_plugins',
```
mariusandra, line 56: Thinking out loud (and this can be done later), but this string should probably also end up in a `setting`. It's used in 2 places already.

src/extensions/posthog.ts
```diff
@@ -1,79 +1,81 @@
...
 
-    const sendEvent = server.KAFKA_ENABLED ? sendEventKafka : sendEventRedis
-
     return {
         capture(event, properties = {}) {
-            const { timestamp, ...otherProperties } = properties
-            sendEvent(event, otherProperties, timestamp || DateTime.utc().toISO())
+            const { timestamp = DateTime.utc().toISO(), ...otherProperties } = properties
+            const data: InternalData = {
+                distinct_id: distinctId,
```
mariusandra, line 67: Out of scope, but writing down just to write it down: I can see how spoofing "distinct_id" can be useful for plugins. For example when importing stuff from elsewhere or even just making `$set` queries to add data to users. 

It's a small change, but taking `distinct_id` from the data above (or `distinctId`?) like is done with the timestamp can unlock many future uses.

tests/helpers/kafka.ts
```diff
@@ -0,0 +1,82 @@
...
+    public producer: Producer
+    public consumer: Consumer
+
+    private isStarted: boolean
+
+    constructor() {
+        super()
+        this.kafka = new Kafka({
+            clientId: `plugin-server-test-${new UUIDT()}`,
+            brokers: process.env.KAFKA_HOSTS!.split(','),
```
mariusandra, line 19: Can we get this from the `server` somehow? Using `process.env` feels dirty, even if it's just in a test.

tests/postgres/worker.test.ts
```diff
@@ -1,11 +1,11 @@
 import { PluginEvent } from '@posthog/plugin-scaffold/src/types'
```
mariusandra, line 1: I'm slightly confused with the naming of the folders here. Some (or all?) tests inside `postgres` are ones that to the best of my understanding operate on things not explicitly related to postgres... Things like checking if the VMs, workers, etc work well. They should be in a higher ingestion-agnostic folder IMO to emphasize this.

src/ingestion/kafka-queue.ts
```diff
@@ -140,6 +142,7 @@ export class KafkaQueue implements Queue {
         consumer.on(CRASH, ({ payload: { error, groupId } }) => {
             status.error('⚠️', `Kafka consumer group ${groupId} crashed:\n`, error)
             Sentry.captureException(error)
+            // killGracefully()
```
mariusandra, line 145: `//` dead code

src/ingestion/kafka-queue.ts
```diff
@@ -130,7 +132,7 @@ export class KafkaQueue implements Queue {
 
     private static buildConsumer(kafka: Kafka): Consumer {
         const consumer = kafka.consumer({
-            groupId: 'plugin-server',
+            groupId: 'clickhouse-ingestion',
```
mariusandra, line 135: is this something that could/should go in a setting/config as well?

tsconfig.json
```diff
@@ -12,7 +12,7 @@
         "sourceMap": true,
         "baseUrl": "./src",
         "outDir": "./dist",
-        "types": ["node", "jest"],
+        "types": ["node", "jest", "long"],
```
mariusandra, line 15: what's `long`?

tests/postgres/worker.test.ts
mariusandra: Also, there are broken imports in the `plugin`, `queue` and `vm` tests. Why not move them back?

src/db.ts
```diff
@@ -0,0 +1,138 @@
...
+import { ClickHouse } from 'clickhouse'
+import { Producer } from 'kafkajs'
+import { DateTime } from 'luxon'
+import { Pool, PoolConfig } from 'pg'
+import { KAFKA_PERSON, KAFKA_PERSON_UNIQUE_ID } from './ingestion/topics'
+import { unparsePersonPartial } from './ingestion/utils'
+import { Person, PersonDistinctId, RawPerson } from './types'
+import { castTimestampOrNow } from './utils'
+
+export class DB extends Pool {
```
mariusandra, line 11: I think this is putting too many responsibilities together. To follow the same pattern we have with each other service we connect to, I'd still keep a `db` (or `postgres`) pure DB connection object on the `server`.

The code in this file could be split into a new `orm` (or `db`) class and then just take `postgres` as another input.

src/server.ts
```diff
@@ -165,6 +188,8 @@ export async function startPluginsServer(
             await stopFastifyInstance(fastifyInstance!)
         }
         await queue?.stop()
+        await pubSub?.quit()
+        await kafkaProducer?.disconnect()
         pubSub?.disconnect()
```
mariusandra, line 193: any issue with `pubSub.quit()` being before `pubSub.disconnect()`?

mariusandra (commented):
Here's a first look at some of the files. I still have the entire ingestion pipeline to go through (nr 1 todo for today), but these comments could get some ball rolling while I spend time on it.

src/server.ts
Twixes: Ah, yep, I think disconnect should go first. Added `quit` as well because otherwise Redis connection wasn't shut down cleanly (or at all sometimes).

tsconfig.json
Twixes: Used by Protobuf.

src/ingestion/kafka-queue.ts
Twixes: Not really, it'd mess ingestion up to change the group ID after starting it for the first time, so this is an internal detail (not worth it to bloat the settings list IMO).

tests/helpers/kafka.ts
Twixes: Not really, because this is supposed to be entirely independent from `server` (the way ClickHouse is).

src/config.ts
Twixes: Is not

src/db.ts
Twixes: I dunno, I wouldn't want to decouple them actually… More complexity in `createServer` and we want to use these methods every single time, so I'm in favor of keeping this integrated.

tests/postgres/worker.test.ts
Twixes: Problem is, these tests will require a bit of a redesign to work with both pipelines. So for now separate.

src/worker/queue.ts
```diff
@@ -72,15 +72,16 @@ async function startQueueKafka(
...
+        await server.eventsProcessor.processEventEE(
             distinct_id,
             ip,
             site_url,
             event,
             team_id,
             DateTime.fromISO(now),
-            sent_at ? DateTime.fromISO(sent_at) : null
+            sent_at ? DateTime.fromISO(sent_at) : null,
+            uuid!
```
mariusandra, line 84: This `!` feels a bit unsafe, though I guess that's the simplest way to assert that it's always there, which it should be for EE. Not sure if anything to do (`if (!uuid) { console.error && return }`?), just thinking out loud again :).

tests/clickhouse/process-event.ts
```diff
@@ -0,0 +1,26 @@
...
+    await resetTestDatabaseClickhouse()
+    await kafkaObserver.start()
+})
+afterEach(() => {
+    closeServer()
+})
+
+test('event is passed through', async () => {
+    expect(1).toEqual(1)
+})
```
mariusandra, line 26: Could probably use stronger tests :)

src/db.ts
mariusandra: I'd still push for separating `postgres` out of this class. This `DB` is now a meta class that uses all different database engines for its work, so extending postgres's `Pool` doesn't really make sense to me, even if that's the database this class mostly uses.

It's hard to explain why the first one in these is without `.postgres`:

```
this.query
this.clickhouse.query
this.kafkaProducer.send
```

I mean, extending classes is usually a `Object > Mammal > Dog` type of thing. Here it's doing `Object > Mammal > Dog & Reptile` 😅

src/worker/queue.ts
Twixes: That's fair, will add a check

src/db.ts
Twixes: Very well then, separated

mariusandra: Tests were failing as a lot of places still called `server.db.query`. I swapped them to `server.db.postgresQuery`. 

We don't have just one `db` unfortunately, so keeping `server.db.query` like it was now adds cognitive steps. The alternatives were `server.db.postgresQuery` and `server.postgres.query`... and of those I found `server.db.postgresQuery` a better fit, since we have so many other methods now on `db.*`, and it's easy to just pass around one object... Hope that makes sense 😅

Twixes:
Hm, something broke in 2199a79, as Postgres tests just don't ever finish (or start, hard to say).

src/ingestion/process-event.ts
```diff
@@ -1,38 +1,455 @@
...
+            } catch {
+                // Catch race condition where in between getting and creating,
+                // another request already created this person
+                personFound = await this.db.fetchPerson(teamId, distinctId)
+            }
+        }
+        this.db.updatePerson(
+            personFound!,
+            setOnce ? { ...properties, ...personFound!.properties } : { ...personFound!.properties, ...properties }
+        )
```
mariusandra, line 179: This looks like a bug. It should be `updatePerson(personFound, { properties: setOnce ? {...} : {...} })`, right?

Could probably `return` or `await` this as well

src/db.ts
```diff
@@ -0,0 +1,148 @@
...
+            })
+        }
+        return personCreated
+    }
+
+    public async updatePerson(person: Person, update: Partial<Person>): Promise<Person> {
+        const updatedPerson: Person = { ...person, ...update }
+        await this.postgresQuery(
+            `UPDATE posthog_person SET ${Object.keys(update).map(
+                (field, index) => field + ' = $' + (index + 1)
```
mariusandra, line 85: `field` here is passed to SQL without sanitisation. We have to be really really sure nobody starts passing arbitrary SQL in the keys of the `update` argument. 

There was/is already a case with passing bad data into `update` below in `processEvent`. Who knows who will be working on this codebase in the future and what this function will be called with. So we should escape or sanitize these fields somehow.

src/db.ts
```diff
@@ -0,0 +1,148 @@
...
+    public async updateDistinctId(
+        personDistinctId: PersonDistinctId,
+        update: Partial<PersonDistinctId>
+    ): Promise<void> {
+        const updatedPersonDistinctId: PersonDistinctId = { ...personDistinctId, ...update }
+        await this.postgresQuery(
+            `UPDATE posthog_persondistinctid SET ${Object.keys(update).map(
+                (field, index) => field + ' = $' + (index + 1)
+            )} WHERE id = $${Object.values(update).length + 1}`,
+            [...Object.values(update), personDistinctId.id]
```
mariusandra, line 139: Same case about escaping field names here.

mariusandra (changes requested):
So, finally finished a full pass over the code! It's definitely missing a lot of tests, but overall it's very good! I did find a few things of note:

- bug with calling `db.updatePerson` in one place
- missing code from python (`_add_missing_feature_flags`)
- one `TODO` for adding `posthoganalytics.capture`
- a db column serialization issue

Regarding tests, at the very least we should replicate the python tests that would get deleted otherwise. Mostly it's [this `test_process_event.py` file](https://github.com/PostHog/posthog/blob/8d65088c5dfdf66149628d31be702a2d539e6345/posthog/tasks/test/test_process_event.py), running under both Clickhouse and Postgres environments.

Twixes:
Feature flags, yep, regarding them, we'll have to do something tricky: we need to get the exact same results in the plugin server as in the main server. I see some general ways:
- we can recreate the whole backend logic in here - complicated and prone to discrepancies (e.g. a differently used hash, database filtering etc. plus ongoing changes in functionality)
- we can hit the `decide` endpoint to get the truth - rather inefficient
- we can relay the truth somehow else?

mariusandra:
I traced the breakage to these lines:

```js
        pubSub?.disconnect()
        await pubSub?.quit()
```

`disconnect` is a hard disconnect, which quits as well. However after fixing them something is still broken...

mariusandra:
Re feature flags, ugh. Scope creep. Might be worth escalating to slack.

src/ingestion/process-event.ts
```diff
@@ -1,38 +1,455 @@
...
+                href: el['attr__href'] ? el['attr__href'].slice(0, 2048) : null,
+                attr_class: el['attr__class'] ? el['attr__class'].split(' ') : null,
+                attr_id: el['attr__id'],
+                nth_child: el['nth_child'],
+                nth_of_type: el['nth_of_type'],
+                attributes: Object.fromEntries(Object.entries(el).filter(([key]) => key.startsWith('attr__'))),
+            }))
+        }
+
+        const team: Team = (await this.db.postgresQuery('SELECT * FROM posthog_team WHERE id = $1', [teamId])).rows[0]
```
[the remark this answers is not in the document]
Twixes: Hmm, we can't actually do anything about that other than discarding the event, which is an occurrence I'd rather log, so best to leave as is IMO

src/ingestion/process-event.ts
```diff
@@ -1,38 +1,455 @@
...
+
+        if (!team.anonymize_ips && !('$ip' in properties)) {
+            properties['$ip'] = ip
+        }
+
+        this.storeNamesAndProperties(team, event, properties)
```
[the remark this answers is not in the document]
Twixes: Yeah, I saw that but this was the only thing I held off on porting because of the complexity. Per our discussion, now handled earlier in the pipeline, in the capture endpoint

src/db.ts
Twixes: Added `sanitizeSqlIdentifier` to delete quotes just in case

src/db.ts
Twixes: Also `sanitizeSqlIdentifier`

src/ingestion/process-event.ts
```diff
@@ -1,38 +1,455 @@
...
+
+        return await this.createEvent(eventUuid, event, team, distinctId, properties, timestamp, elements_list)
+    }
+
+    private async storeNamesAndProperties(team: Team, event: string, properties: Properties): Promise<void> {
+        // In _capture we only prefetch a couple of fields in Team to avoid fetching too much data
+        let save = false
+        if (!team.ingested_event) {
+            // First event for the team captured
+            // TODO: capture "first team event ingested"
```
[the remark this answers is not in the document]
Twixes: Done. I had to rework `DummyPostHog` to capture using that instead of introducing `posthog-node` – feel free take a look at the implementation

src/ingestion/process-event.ts
```diff
@@ -10,18 +10,23 @@ import { ClickHouse } from 'clickhouse'
...
+    posthog?: DummyPostHog
 
     constructor(pluginsServer: PluginsServer) {
         this.pluginsServer = pluginsServer
         this.db = pluginsServer.db
         this.clickhouse = pluginsServer.clickhouse!
         this.kafkaProducer = pluginsServer.kafkaProducer!
+        if (pluginsServer.POSTHOG_TEAM_ID) {
+            this.posthog = createPosthog(this.pluginsServer, parseInt(pluginsServer.POSTHOG_TEAM_ID))
+        }
```
mariusandra, line 29: In python we're emitting this "first team event ingested" event to app.posthog.com. The code above would do that only if co-located with posthog in the same database (so, cloud only). We should be able to capture these events no matter where this is running.

I'd swap this out with `posthog-js-lite` or even a raw API call. `-js-lite` is a pretty bad library, but it's good enough to abstract capture calls into fetch requests for the toolbar, so it could also work here.

src/ingestion/process-event.ts
```diff
@@ -114,12 +119,8 @@ export class EventsProcessor {
...
                 await this.alias(properties['$anon_distinct_id'], distinctId, teamId)
             }
-            // TODO: roll the two updatePersonProperties calls into one to `UPDATE posthog_person` only once
-            if (properties['$set']) {
-                this.updatePersonProperties(teamId, distinctId, properties['$set'])
-            }
-            if (properties['$set_once']) {
-                this.updatePersonProperties(teamId, distinctId, properties['$set_once'], true)
+            if (properties['$set'] || properties['$set_once']) {
+                this.updatePersonProperties(teamId, distinctId, properties['$set'] || {}, properties['$set_once'] || {})
```
mariusandra, line 123: 😍

tests/utils.test.ts
```diff
@@ -295,3 +296,13 @@ describe('UUIDT', () => {
...
+
+describe('sanitizeSqlIdentifier', () => {
+    it('removes quotes inside and adds quotes around identifier', () => {
+        const rawIdentifier = 'some_field"; DROP TABLE actually_an_injection;'
+
+        const sanitizedIdentifier = sanitizeSqlIdentifier(rawIdentifier)
+
+        expect(sanitizedIdentifier).toStrictEqual('"some_field; DROP TABLE actually_an_injection;"')
+    })
+})
```
mariusandra, line 308: What about `firstKey = "haha; -- \\"` It would escape the closing quote. Combined with `otherKey = "; here we're outside the quotes... -- "` you could have some fun. Since we're the ones providing the keys, let's whitelist characters and remove everything not in some sane set.

mariusandra (changes requested):
Some inline comments and then these points:

- I'm not sure of the `POSTHOG_TEAM_ID` bit. I'd replace it with a http library to send events to `app.`, like we do in django. We're already deploying private clouds, so even if we deploy this ingestion only to cloud, we're going to need it to talk to other servers. We could use `-js-lite` perhaps? It's bad, but a good enough abstraction over raw HTTP calls I think.

- For me the full e2e kafka ingestion tests are not that critical. The thing that's absolutely needed before I could merge this PR are these [`test_process_event.py`](https://github.com/PostHog/posthog/blob/master/posthog/tasks/test/test_process_event.py) tests. The easiest way to get them up is to add postgres ingestion (just two insert statements, right?) and test it all against that. If all those special alias, $ip, set, etc cases work well, this PR will be almost good to go. At least we could already deliver postgres opt-in plugin ingestion in this release. I'd be happy to share the workload somehow.

- That said, some test fails again. Should I deep dive here as well? I might get too deep to be out before the next code freeze...

tests/utils.test.ts
Twixes: Yup, basically just made it remove all chars besides letters, digits and underscores, nuclear option

src/ingestion/process-event.ts
Twixes: Totally forgot about instances other than Cloud! `posthog-js-lite` will do the job just fine I think, rewritten to use it

mariusandra (approved):
LGTM!

PR #35, opened 2020-12-02
More plugin tests

mariusandra:
- Tests loading plugins, plugin config and plugin attachments from postgres
- Tests running these plugins
- Test loading plugins from the local filesystem
- Test various failure scenarios

Missing from here:
- Tests for loading global plugins (this doesn't work now and will be rewritten anyway)
- Tests for loading lib.js files (will be deprecated pretty soon)


mariusandra:
Ready to be nitted ;).

mariusandra:
Added a few more tests for SQL functions. Ready to be looked at.

Twixes (approved):
LGTM

PR #36, opened 2020-12-03
Multithreaded server

mariusandra:
Seems to work, though very WIP. I'll add details later.

Had a lot of trouble getting typescript, start:dev, start:dist and jest to just work with the worker threads...


mariusandra:
There's one issue with a failing test (more below), but the code can be reviewed already.

Way too many changed with this PR and looking back this should have been a bunch of smaller PRs, but here we are. 

Changed:
- Remove `rollup` and just use `tsc`. There's no need to compile all of our ".js" files into one big ".js" file. Plus if I want to start a worker thread, I need it in a separate ".js" folder that can be included by node.
- Add [`piscina`](https://github.com/piscinajs/piscina) to parallelise plugin workers. It's a new project with not *that* many downloads, yet seems to be made with people who have credentials. It works well.
- Refactor the tests to use `helpers`
- Add a `--concurrency` flag to the CLI
- Store all the `plugins`, `pluginConfigs` and other data on the `server`
- Give up and turn back on the `esModuleInterop` tsconfig flag.
- Implement a horrible mocking system (`server.__jestMock`) to get fake data into the worker threads. 
- Create `src/worker` folder and refactor some files. Rename `src/worker` -> `src/worker/queue`
- Create a test to make sure that adding more cores actually increases performance

Todo:
- Looking at the tests as they are in 2 different folders, it might indeed be better to put all of them under a common `tests/` home. This is outside the scope of this project.
- The worker threads can't be tested in Github Actions as they don't get the redis and pg mocks from normal jest. Thus they try to connect to the action pg/redis instances and fail. 

I also created a benchmark. Running it with this [CPU-intense plugin](https://github.com/PostHog/posthog-plugin-server/blob/8f42315419fb346b45a784815a441899fe345cd9/src/worker/__tests__/piscina.test.ts#L52):

```js
function processEvent (event, meta) {
    let j = 0; for(let i = 0; i < 200000; i++) { j = i };
    event.properties = { "somewhere": "over the rainbow" }; 
    return event
}
```

... gives me the following results:

```
// Thread count --> events/sec
{
    "1": 6956,
    "2": 12635,
    "4": 20826,
    "8": 23000,
    "12": 27003
}
```

More tests to follow.

I'll see what can be done for the test. However it shouldn't change much and the rest of the code can already be looked at.

bin/posthog-plugin-server
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env node
 
-require(__dirname + '/../dist/index.js')
+require(__dirname + '/../dist/src/index.js')
```
Twixes, line 3: This `/dist/src/` structure looks a bit odd, can we avoid the `src` here?

package.json
```diff
@@ -2,21 +2,22 @@
...
         "build": "yarn clean && yarn compile",
         "clean": "rimraf dist/*",
-        "compile": "rollup -c",
+        "compile": "tsc -p .",
         "lint": "eslint .",
         "lint:fix": "eslint --fix .",
         "prettier": "prettier --write .",
         "prettier:check": "prettier --check .",
-        "prepublish": "yarn test && yarn build"
+        "prepublishOnly": "yarn test && yarn build",
```
Twixes, line 19: Looks a bit out of place with camelCase, BTW I think we could do away with "only"?

package.json
```diff
@@ -73,12 +74,6 @@
         "lint-staged": ">=10.5.1",
         "prettier": "^2.1.2",
         "rimraf": "^3.0.0",
-        "rollup": "^2.22.1",
-        "rollup-plugin-babel": "^4.4.0",
-        "rollup-plugin-commonjs": "^10.1.0",
-        "rollup-plugin-dts": "^1.4.9",
-        "rollup-plugin-node-resolve": "^5.2.0",
-        "rollup-plugin-typescript2": "^0.29.0",
```
Twixes, line 81: Nice.

src/server.ts
```diff
@@ -1,15 +1,26 @@
...
+
+function overrideWithEnv(config: PluginsServerConfig): PluginsServerConfig {
+    const newConfig: Record<string, any> = { ...config }
+    for (const [key, value] of Object.entries(config)) {
+        if (process.env[key]) {
+            newConfig[key] = process.env[key]
+        }
+    }
+    return config
+}
```
Twixes, line 21: :+1: As as side note, I think we could move to only env var-based server configuration instead of passing in JSON (another PR).

src/worker/queue.ts
```diff
@@ -39,5 +41,5 @@ export function startWorker(server: PluginsServer): Worker {
 
     worker.start()
 
-    return worker
+    return async () => worker.stop()
```
Twixes, line 44: Why not just return the whole worker for more flexibility. Could then call `worker.stop` ourselves somewhere else, but also any other `worker` method.

tasks/db-init.ts
```diff
@@ -0,0 +1,88 @@
...
+        tag           varchar(200),
+        archive       bytea,
+        from_json     boolean not null,
+        from_web      boolean not null,
+        error         jsonb
+    );
+`
+
+const createPluginAttachment = `
+    create table ${ifNotExists} posthog_pluginattachment
```
Twixes, line 34: Totally nitpicking, but conventionally SQL statements are ~~SCREAMING~~ uppercase, which is slightly clearer.

tsconfig.json
```diff
@@ -7,6 +7,8 @@
         "emitDecoratorMetadata": true,
         "experimentalDecorators": true,
         "moduleResolution": "node",
+        "esModuleInterop": true,
+        "allowJs": true,
```
Twixes, line 11: For which dependency do we allow JS?

bin/posthog-plugin-server
mariusandra: I agree amd I tried to... yet wasting enough effort trying, without some manual rm/mv/etc steps, I couldn't get this to work :(.

The problem was mainly with `package.json`, which is in the root folder and thus the typescript wants to move the root of "src" also into dist.

Solving that, the next problem was with `allowJs`. We need that for the `worker/piscina.js` file, which runs the workers as needed. Changing that to ".ts" and having it work with all "ts-node" (start:dev), "tsc" (build) and "jest" (test) seemed like a bigger rabbit hole that I was willing to get into. 

Thus to not waste too much time here, I just moved on :/.

package.json
mariusandra: That's unfortunately `npm`/`yarn` requiring this capitalisation.

- `prepublish` runs the scripts at both "yarn install" and "yarn publish"
- `prepublishOnly` runs the script at only "yarn publish"

I don't want all the tests to run at `yarn` all the time. Otherwise I get too many errors and it just seems weird. Some tests also a working database, which is created with the `tasks:db-init` script. I'm somewhat hesitant to add this to the test script. Also, [posthog-js also uses `prepublishOnly`](https://github.com/PostHog/posthog-js/blob/master/package.json#L15)

package.json
mariusandra: Pure `tsc` is a better choice for purely nodejs apps. I originally wanted rollup so I could compile all the dependencies into one `.js` file that could be quickly installed and executed in the main `posthog` repo. There were some problems with that:
- Some dependencies didn't want to be compiled into one `.js` file due to a bunch of random reasons, so I had to keep them as separate `require` statements.
- Eventually it was easier to just keep all dependencies outside the `index.js` and require them in, only compiling all of `posthog-plugin-server` into one `.js`
- Introducing piscina workers, which require loading a separate `.js` file that gets started in a new thread, broke that system... and it was easier to just remove rollup.

src/server.ts
mariusandra: I tend to agree! We might want a CLI override (e.g. `--database-url=has://priority/over/env`), yet agreed that the current `--config={json}` approach is rather raw.

src/worker/queue.ts
mariusandra: Indeed, why not.
Historically the old `queue` ran both `worker.stop()` and `client.stop()` on shutdown, yet now that celery is inlined and there's no separate connection being opened in `client`, we can indeed directly return the `worker`. Fixed

tasks/db-init.ts
mariusandra: SCREAMING implemented! I just uppercased all keywords and left the field types in place (`varchar`), even though they could scream as well...

tsconfig.json
mariusandra: For the `worker/piscina.js` file

src/server.ts
```diff
@@ -22,7 +22,7 @@ function overrideWithEnv(config: PluginsServerConfig): PluginsServerConfig {
 
 export const defaultConfig: PluginsServerConfig = overrideWithEnv({
     CELERY_DEFAULT_QUEUE: 'celery',
-    DATABASE_URL: 'postgres://localhost:5432/posthog',
+    DATABASE_URL: 'postgres://kkjkj:kkhkhkj@localhost:5432/posthog',
```
Twixes, line 25: What happened here? 🤔

bin/posthog-plugin-server
Twixes: Hm, okay, maybe some other time.

package.json
Twixes: Oh, I had no idea that that was something `prepublish` is doing and `prepublishOnly` isn't. Good to know.

src/server.ts
mariusandra: Whoops! Debug run amok!

Twixes (approved):
Nice, GH check passing. I think this is good to go then.

mariusandra:
Finally! Tests pass!

And 2 workers on github also gives better results than 1 worker:
image

This test might become flaky at some point, but good for now!

Thanks for the review. Merging!

PR #39, opened 2020-12-07
Process events in batches

mariusandra:
~~Still WIP~~

Ready for review (comments below)


mariusandra:
This is ready for a look!

Inside the PR:
- Add support for a `processEventBatch(batch: PluginEvent[], meta: PluginMeta)` function inside plugins that can operate on a batch of events. For example to send 100 events at once to S3 instead of making 100 HTTP requests.
- Either write `processEvent` or `processEventBatch` in your plugin. The other gets stubbed automatically.
- Batches are split by team - all events in one batch are for the same team
- Remove support for `Posthog.capture` inside `processEvent` and `processEventBatch`. Segment also doesn't support this in their destination functions. The problem is that since `$identity` events don't contain a `token` field, we can't initialize the posthog instance like we can for other events. We can add this back later by asking for the API key from postgres. However I'd just add it then to the scheduled/webhook plugins and not the `processEvent` side of thing to avoid infinite loops.
- Update worker benchmark tests to see if batching makes a difference. (Spoiler: it does... will reply in the next comment)

mariusandra:
I added one more fix since batching seemed to slow down sync events. Now in the stubbed `processEventBatch` function we only `await Promise.all([...])` if we need to, greatly speeding things up!

Tested on a M1 Air. I got up to 100k events/sec with batches of 100. 🚀 

```
// Before last fix
      ┌─────────┬────────────────┬───────────┬────────┬───────────┬──────────┬───────────┬───────────┬───────────┐
      │ (index) │    testName    │ coreCount │ events │ batchSize │ 1 thread │ 2 threads │ 4 threads │ 8 threads │
      ├─────────┼────────────────┼───────────┼────────┼───────────┼──────────┼───────────┼───────────┼───────────┤
      │    0    │    'simple'    │     8     │ 10000  │     1     │  39453   │   70847   │   81106   │   76728   │
      │    1    │    'simple'    │     8     │ 10000  │    10     │  24070   │   40898   │   54639   │   59051   │
      │    2    │    'simple'    │     8     │ 10000  │    100    │  30800   │   52167   │   69978   │   73826   │
      │    3    │   'for200k'    │     8     │ 10000  │     1     │  12650   │   22944   │   37368   │   43391   │
      │    4    │   'for200k'    │     8     │ 10000  │    10     │   9495   │   17370   │   27272   │   35226   │
      │    5    │   'for200k'    │     8     │ 10000  │    100    │  10329   │   19656   │   33916   │   37400   │
      │    6    │ 'timeout100ms' │     8     │ 10000  │     1     │   1010   │   1987    │   3352    │   5101    │
      │    7    │ 'timeout100ms' │     8     │ 10000  │    10     │   8839   │   16482   │   27245   │   31118   │
      │    8    │ 'timeout100ms' │     8     │ 10000  │    100    │  16397   │   29250   │   38102   │   38000   │
      └─────────┴────────────────┴───────────┴────────┴───────────┴──────────┴───────────┴───────────┴───────────┘

// After last fix
      ┌─────────┬────────────────┬───────────┬────────┬───────────┬──────────┬───────────┬───────────┬───────────┐
      │ (index) │    testName    │ coreCount │ events │ batchSize │ 1 thread │ 2 threads │ 4 threads │ 8 threads │
      ├─────────┼────────────────┼───────────┼────────┼───────────┼──────────┼───────────┼───────────┼───────────┤
      │    0    │    'simple'    │     8     │ 10000  │     1     │  44184   │   70350   │   79122   │   81577   │
      │    1    │    'simple'    │     8     │ 10000  │    10     │  31499   │   51685   │   75783   │   82023   │
      │    2    │    'simple'    │     8     │ 10000  │    100    │  46066   │   75966   │   99487   │  115690   │
      │    3    │   'for200k'    │     8     │ 10000  │     1     │  12684   │   23340   │   37095   │   42061   │
      │    4    │   'for200k'    │     8     │ 10000  │    10     │  10462   │   18992   │   31391   │   35820   │
      │    5    │   'for200k'    │     8     │ 10000  │    100    │  11667   │   22335   │   39604   │   43295   │
      │    6    │ 'timeout100ms' │     8     │ 10000  │     1     │   1011   │   1981    │   3417    │   5115    │
      │    7    │ 'timeout100ms' │     8     │ 10000  │    10     │   8801   │   16270   │   26732   │   31813   │
      │    8    │ 'timeout100ms' │     8     │ 10000  │    100    │  16475   │   28314   │   33264   │   38892   │
      └─────────┴────────────────┴───────────┴────────┴───────────┴──────────┴───────────┴───────────┴───────────┘
```

tests/worker.test.ts
```diff
@@ -62,10 +72,12 @@ function setupPiscina(workers: number, code: string, tasksPerWorker: number) {
 }
 
 test('piscina worker test', async () => {
-    const coreCount = os.cpus().length
+    // Uncomment this to become a 10x developer and make the test run just as fast!
+    const isDevRun = false
```
Twixes, line 76: What is this for?

tests/worker.test.ts
```diff
@@ -7,48 +7,58 @@ import * as os from 'os'
 import { LogLevel } from '../src/types'
 
 jest.mock('../src/sql')
-jest.setTimeout(300000) // 300 sec timeout
+jest.setTimeout(600000) // 600 sec timeout
```
Twixes, line 10: Pretty long time, could we make this test take up less time somehow?

src/plugins.ts
```diff
@@ -180,26 +180,22 @@ async function loadPlugin(server: PluginsServer, pluginConfig: PluginConfig): Pr
...
+            let errored = false
+            const { processEvent } = pluginConfig.vm.methods
+            const startTime = performance.now()
+            try {
+                returnedEvent = (await processEvent(returnedEvent)) || null
+            } catch (error) {
+                errored = true
+                await processError(server, pluginConfig, error, returnedEvent)
             }
+            const ms = Math.round((performance.now() - startTime) * 1000) / 1000
```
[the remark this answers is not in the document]
mariusandra: I removed it altogether since we always round the throughput up to the nearest integer

src/vm.ts
```diff
@@ -65,9 +65,32 @@ export function createPluginConfigVM(
...
+                    const e = processEvent(event, meta)
+                    if (e && typeof e.then !== 'undefined') {
+                        waitFor = true
+                    }
+                    return e
+                })
+                const response = waitFor ? (await Promise.all(processedEvents)) : processedEvents;
+                return response.filter(r => r)
+            }
+        // we have processEventBatch, but not processEvent
```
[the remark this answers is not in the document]
mariusandra: The user handled it for us in that case. I think it's fine for the user to give both functions (e.g. making custom batching that only works with max 10 events), yet if they want to give just one of the two, the system should still work. Also if they give zero, the system should work... and then just do the plugin differently.

tests/worker.test.ts
mariusandra: The test actually takes closer to 2min than 10min. but when I did some heavy testing locally it went over the 5min mark. I incremented this just to not have to manually change it next time I run heavier tests. I think it's fine to keep as is. Anything longer than that has surely gotten stuck.

src/plugins.ts
```diff
@@ -209,3 +205,47 @@ export async function runPlugins(server: PluginsServer, event: PluginEvent): Pro
...
+    let allReturnedEvents: PluginEvent[] = []
+
+    for (const [teamId, teamEvents] of eventsByTeam.entries()) {
+        const pluginsToRun = getPluginsForTeam(server, teamId)
+
+        let returnedEvents: PluginEvent[] = teamEvents
+
+        for (const pluginConfig of pluginsToRun.reverse()) {
+            if (pluginConfig.vm?.methods?.processEventBatch && returnedEvents.length > 0) {
+                const { processEventBatch } = pluginConfig.vm.methods
```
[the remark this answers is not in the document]
mariusandra: cleaned it up a bit

tests/worker.test.ts
mariusandra: I added an extra line to describe it. Basically when working on tests and running them many times in a row, it made sense to run a "light" version of the test.

```
    // Uncomment this to become a 10x developer and make the test run just as fast!
    // Reduces events by 10x and limits threads to max 8
    const isLightDevRun = false
```

src/plugins.ts
```diff
@@ -235,8 +234,7 @@ export async function runPluginsOnBatch(server: PluginsServer, batch: PluginEven
                     errored = true
                     await processError(server, pluginConfig, error, returnedEvents[0])
                 }
-                const ms = Math.round((performance.now() - startTime) * 1000) / 1000
-                logTime(pluginConfig.plugin?.name || 'noname', ms, errored)
+                logTime(pluginConfig.plugin?.name || 'noname', startTime - performance.now(), errored)
```
Twixes, line 237: Aren't we subtracting a larger (newer) value from a smaller (older) one?

tests/worker.test.ts
```diff
@@ -138,8 +138,7 @@ test('piscina worker test', async () => {
                 for (let i = 0; i < rounds; i++) {
                     await processCountEvents(piscina, events / batchSize, batchSize)
                 }
-                const throughput = Math.round(1000 / ((startTime - performance.now()) / events / rounds))
-                result[`${threads} thread${threads === 1 ? '' : 's'}`] = Math.round(throughput)
+                result[`${threads} thread${threads === 1 ? '' : 's'}`] = Math.round(startTime - performance.now())
```
mariusandra, line 141: This is not correct as we want events per second here.

src/plugins.ts
mariusandra: You're right! I See you're in the middle of fixing things, so feel free to reverse this!

tests/worker.test.ts
Twixes: Yep, was too fast here, fixed.

PR #41, opened 2020-12-07
Benchmarks & various changes

mariusandra:
- Add `tasks per worker` setting
- Fix some test deprecation warnings
- Make the benchmark test print graphs
- Add two more types of benchmarks

Depends on #40


mariusandra:
Also added:
- Better handling of CLI arguments (directly created from the `defaultConfig`)
- Renamed `piscina.test.ts` to `worker.test.ts`
- Exit server if can't connect to redis
- Add `__jestSetTimeout` inside VM when testing with jest to simulate requests
- Add a simple option to set the log level in order to have less noisy tests
- Chunk test events into batches of 100 for better async. (Waiting for 100 promises 10 times is better than waiting for 10k promises 1 time)

Can be reviewed!

src/index.ts
```diff
@@ -11,31 +13,36 @@ type Argv = {
...
+                }
             }
-            startPluginsServer(parsedConfig, makePiscina)
-        },
-    }).argv
+        }
+        setLogLevel(parsedConfig.LOG_LEVEL || defaultConfig.LOG_LEVEL)
+        startPluginsServer(parsedConfig as PluginsServerConfig, makePiscina)
+    },
+}).argv
```
Twixes, line 48: Do we actually need a command ("start")? The server does nothing BUT start, so having to write "start" doesn't add anything.

src/types.ts
```diff
@@ -3,8 +3,11 @@ import { Redis } from 'ioredis'
 import { PluginEvent, PluginAttachment, PluginConfigSchema } from 'posthog-plugins'
 import { VM, VMScript } from 'vm2'
 
+export type LogLevel = 'debug' | 'info' | 'log' | 'warn' | 'error' | 'none'
```
Twixes, line 6: Nitpick: good use case for an enum.

tests/worker.test.ts
```diff
@@ -0,0 +1,133 @@
...
+            await piscina.destroy()
+        }
+        results.push(result)
+        console.log(JSON.stringify({ result }, null, 2))
+    }
+    console.table(results)
+    // expect that adding more cores (up to coreCount) increases throughput
+    // for (let i = 1; i < workers.length; i++) {
+    //     expect(result[`${workers[i - 1]} cores`]).toBeLessThan(result[`${workers[i]} cores`])
+    // }
```
Twixes, line 132: I may be wrong, but this looks like commented out dead cod.

.eslintrc.js
```diff
@@ -9,6 +9,7 @@ module.exports = {
         '@typescript-eslint/no-non-null-assertion': 'off',
         '@typescript-eslint/no-var-requires': 'off',
         '@typescript-eslint/no-explicit-any': 'off',
+        '@typescript-eslint/no-empty-function': 'off',
```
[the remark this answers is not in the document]
mariusandra: fixed!

src/config.ts
```diff
@@ -0,0 +1,48 @@
...
+        DISABLE_WEB: false,
+        WEB_PORT: 3008,
+        WEB_HOSTNAME: '0.0.0.0',
+        WORKER_CONCURRENCY: 0, // use all cores
+        TASKS_PER_WORKER: 1,
+        LOG_LEVEL: 'info',
+    }
+}
+
+function getConfigHelp(): Record<string, string> {
```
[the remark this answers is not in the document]
mariusandra: fixed! Had to do some magic because of a special `__jestMocks` that should be excluded from this

src/index.ts
mariusandra: Agreed, removed

src/types.ts
mariusandra: Enum'd!

tests/worker.test.ts
mariusandra: RIP

Twixes (approved):

PR #42, opened 2020-12-08
Add `incr` and `expire` to `meta.cache`

mariusandra:
- Adds `incr`. With max 100 concurrent events per worker, the `await cache.set('counter', await cache.get('counter', 0) + 1)` pattern no longer works, so this makes a lot of sense.
- Expiration is also a useful thing for a cache, so added it as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
         "pg": "^8.4.2",
         "piscina": "^2.1.0",
         "posthog-js-lite": "^0.0.5",
-        "posthog-plugins": "^0.2.3",
+        "posthog-plugins": "^0.2.4",
         "tar-stream": "^2.1.4",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -34,7 +34,7 @@
         "pg": "^8.4.2",
         "piscina": "^2.1.0",
         "posthog-js-lite": "^0.0.5",
-        "posthog-plugins": "^0.2.3",
+        "posthog-plugins": "^0.2.4",
```
Twixes, line 37: This should probably be pinned to a specific version if this is actually important.

Twixes (approved):
👍

package.json
mariusandra: pinned!

PR #50, opened 2020-12-09
Add Sentry

### LINKED ISSUE #48: Add Sentry
If the plugin server crashes in production, we have no way of knowing.

mariusandra:
- Adds a `init.ts` file for code that should be run on init in the main thread and also all the worker threads
- Adds support for sentry error logging if SENTRY_DSN is set
- Closes #48


mariusandra:
I'll go ahead and merge it, so I could release a new version of plugin server to try on Heroku.

PR #51, opened 2020-12-09
Process shutdown

mariusandra:
- Add some error handling to the queue
- Wait 2 sec before shutting down workers on reload (needs a better solution)
- Refactor `setTimeout` promises do `delay` util


mariusandra:
Solves a few bugs I found running in Heroku. Merging directly.

PR #53, opened 2020-12-09
Cleaner logging

Twixes:
This is some changes from extracted #34, aiming to make logging slightly cleaner.


mariusandra:
Great! :)

PR #54, opened 2020-12-09
Conf in env

mariusandra:
- Add $CONFIG (all other cli args work with envs, so why not this as well)
- Simplify package.json scripts


mariusandra:
I'll merge this and release 0.4.3 to unblock some work in the main posthog repo...

PR #60, opened 2020-12-10
Add force exit

Twixes:
On first exit signal (SIGINT / SIGTERM / SIGHUP) e.g. caused by Ctrl+C the server will try to shut down gracefully. Now, on second exit signal it will notify that the third one will cause force shutdown, and on third exit signal it will force exit the process with code 0.

Exit

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -67,13 +67,17 @@ export async function startPluginsServer(
     let queue: Worker | undefined
     let closeServer: () => Promise<void> | undefined
 
-    let shuttingDown = false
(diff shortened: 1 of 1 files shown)


mariusandra (approved):
Seems legit.

What was the reason for developing this though? Some this didn't close on their own in testing?

Twixes:
`node-rdkafka` seems to be blocking exit completely when connecting. 🤔 Not yet sure, but I was unable to get out and this is handy for such situations.

PR #61, opened 2020-12-10
Add Statsd

### LINKED ISSUE #38: Plugin server metrics
There are many metrics that the [worker-based multithreaded plugin server](https://github.com/PostHog/posthog-plugin-server/pull/36) can export. 

For example the worker/task library provides:
- Average task runtime: https://github.com/piscinajs/piscina#property-runtime-readonly
- Average CPU % spent waiting for IO: https://github.com/piscinajs/piscina#property-waittime-readonly

Plus we can export things like:
- Number of events processed
- Breakdown by category (scheduled task, webhook, process_event, etc)
- Failure rate

Celery in posthog is reporting with statsd. Should I do the same here? I'm new to this monitoring world. @fuziontech what would you suggest?

mariusandra:
Using #34 as inspiration, here's a statsd implementation that logs the time it takes to process events (both total per task and per-plugin) and some other piscina metrics (utilisation, worker count, queue size). 

It seems to work and log enough things:

image

Closes #38


mariusandra:
Merging based on feedback in Slack.

PR #63, opened 2020-12-11
Scheduled tasks

mariusandra:
This implements scheduled tasks inside plugins!

Inside plugins you can now use three extra functions: `runEveryMinute(meta)`, `runEveryHour(meta)`, `runEveryDay(meta)`

```js
async function runEveryMinute (meta) {
    const cursor = await meta.cache.get('last_fetched_cursor')
    const response = await fetch(`http://my.api.com/get_stargazers?since=${cursor}`)
    for (const { username } of response.stargazers) {
        posthog.capture('someone starred you!', { username })
    }
    const last_cursor = response.stargazers.map(s => s.created_at).sort().reverse()[0] || cursor
    await meta.cache.set('last_fetched_cursor', last_cursor)
}
```

The above is still pseudocode (need to re-enable the `posthog` var), but this should basically work.

The functions are currently run per worker, but not per thread (so one time per worker instance per minute/hour/day). There's no locking system to prevent code from running e.g. 2 times if you have 2 plugin servers running. I think this is fine for a lot of cases though and definitely OK for a beta!

### DIFF
--- a/src/error.ts
+++ b/src/error.ts
@@ -1,11 +1,11 @@
-import { PluginConfig, PluginError, PluginsServer } from './types'
+import { PluginConfig, PluginConfigId, PluginError, PluginsServer } from './types'
 import { PluginEvent } from 'posthog-plugins'
 import { setError } from './sql'
 import * as Sentry from '@sentry/node'
 
 export async function processError(
     server: PluginsServer,
-    pluginConfig: PluginConfig,
+    pluginConfig: PluginConfig | PluginConfigId,
     error: Error | string,
     event?: PluginEvent | null
 ): Promise<void> {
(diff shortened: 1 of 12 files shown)


mariusandra:
Another fun fact about `runEveryMinute` and other functions. I added debouncing. This means that if your one `runEveryMinute` task is taking more than a minute, the next minute nothing gets started. Only after the task has finished will another one start on the next minute.

mariusandra:
And there's a github action test error. Everything passes locally of course...

Twixes:
Code looks good after a `runTask` fix, though I'd test this end-to-end before merging (didn't yet). And I'm getting a failure locally as well.

Fail

mariusandra:
We definitely need better E2E tests... and/or deepter tests for `server.ts`.

I ran this E2E and all works. Found one last bug - the `runEveryMinute` task was run every second. Otherwise, all seems good!

mariusandra:
image

PR #66, opened 2020-12-11
Support source plugins

mariusandra:
This is what is needed in the plugin server to support plugins written with the code editor


Twixes:
It's surprisingly little

PR #70, opened 2020-12-14
Consume events from EE Kafka events WAL

Twixes:
This is the base of #34, allowing for consuming events from Kafka.


src/server.ts
```diff
@@ -25,11 +24,17 @@ export async function createServer(
...
-    })
+    redis
+        .on('error', (error) => {
+            Sentry.captureException(error)
+            console.error('🔴 Redis error encountered! Trying to reconnect...')
+            console.error(error)
+        })
+        .on('ready', () => {
+            if (!areWeTestingWithJest()) {
+                console.info(`✅ Connected to Redis at ${serverConfig.REDIS_URL}!`)
```
mariusandra, line 35: I might be overanalysing this, but the `REDIS_URL` env often contains passwords and other sensitive data, which *might* not be nice to output into the log.

src/types.ts
```diff
@@ -125,3 +133,32 @@ export interface PluginConfigVMReponse {
...
+}
+
+export interface EventData extends PluginEvent {
+    properties?: Properties
+    timestamp?: string
+    $set?: Properties
+    offset?: number
+}
+
+export type Properties = Record<string, any>
```
mariusandra, line 164: Could it make sense to move some of this (or all of this) inside the `PluginEvent` type inside the `posthog-plugins` package directly?

Twixes:
Caused some stupid import problems, but it's good now, so sorry for that.
The suggested approach for testing this end-to-end is running `ee/docker-compose.ch.yml` in the main repo and `yarn compile && EE_ENABLED=1 KAFKA_HOSTS=localhost DATABASE_URL="postgres://posthog:posthog@127.0.0.1:5439" ./bin/posthog-plugin-server` in this one. `yarn start` is a bit flaky as the Kafka consumer tends to take a while to wind down and that doesn't work well with autoreload.
Now for some weird stuff, CI here is failing due to a segfault, don't know what's up with that as that's not what I'm seeing locally. 🤔

src/worker/queue.ts
```diff
@@ -48,3 +59,83 @@ export function startQueue(
...
+                            const timer = new Date()
+                            const rawEventMessage = JSON.parse(message.value!.toString()) as RawEventMessage
+                            const parsedEventMessage: ParsedEventMessage = {
+                                ...rawEventMessage,
+                                data: JSON.parse(rawEventMessage.data),
+                                now: DateTime.fromISO(rawEventMessage.now),
+                                sent_at: rawEventMessage.sent_at ? DateTime.fromISO(rawEventMessage.sent_at) : null,
+                            }
+                            console.info(`Processing event ${parsedEventMessage.data.event} from WAL`)
+                            const processedEvent = await processEvent({
```
mariusandra, line 101: This bit should run `processEventBatch` on all the events at once.

src/worker/queue.ts
```diff
@@ -48,3 +59,83 @@ export function startQueue(
...
+
+function startQueueKafka(
+    server: PluginsServer,
+    processEvent: (event: PluginEvent) => Promise<PluginEvent | null>
+): Queue {
+    const eventsProcessor = new EventsProcessor(server)
+
+    eventsProcessor.kafkaConsumer.on('ready', () => {
+        eventsProcessor.kafkaConsumer.subscribe([KAFKA_EVENTS_WAL])
+        // consume event messages in batches of 1000 every 50 ms
```
mariusandra, line 71: This scares me. 

I would still greatly prefer a solution like https://github.com/PostHog/posthog-plugin-server/issues/62 - where we pause the consumption when the piscina queue gets full and resume once we have capacity again. Like it is now, if the server gets any backlog, it'll just consume messages endlessly. They will all be `await`ing... and we will probably run out of memory and explode.

src/ingestion/process-event.ts
```diff
@@ -0,0 +1,66 @@
+import { KafkaConsumer, Producer, ProducerStream } from '@posthog/node-rdkafka'
+import { DateTime } from 'luxon'
+import { PluginsServer, EventData, Properties, Queue } from 'types'
+import { UUIDT } from '../utils'
+import { KAFKA_EVENTS, KAFKA_SESSION_RECORDING_EVENTS } from './topics'
+import { Pool } from 'pg'
+
+export class EventsProcessor implements Queue {
```
mariusandra, line 8: This looks like an empty class. I guess you have plans for it in the future, but then I would separate `process_event_ee` and anything that has to do with kafka into two separate lumps of code.

src/worker/queue.ts
```diff
@@ -48,3 +59,83 @@ export function startQueue(
...
+                                    DateTime.fromISO(now),
+                                    sent_at ? DateTime.fromISO(sent_at) : null
+                                )
+                                console.info(`Ingested event ${parsedEventMessage.data.event}!`)
+                            } else {
+                                console.info(`Discarding event ${parsedEventMessage.data.event}`)
+                            }
+                            server.statsd?.timing(`posthog-plugin-server-ingestion`, timer)
+                        }
+                        eventsProcessor.kafkaConsumer.commit()
```
mariusandra, line 126: I think there's a timing issue here that may result in data loss. My knowledge of Kafka is limited, though I assume committing here moves some sort of "read index" down the list of events to what was just read, so they would not be read again. Basically I'm assuming it doesn't commit reads batch by batch, but just commits the last read index.. For example with batches A, C and D read and processed, batch B is still processing due to some long running code, committing D would also mark B as read.

Assuming a new batch is processed every 50ms, we can run into a situation where we get 100 messages in the first batch and 0 messages 50ms later, process them in an instant and commit "yup, we're here now", even though the first batch is still processing in the background.

mariusandra (changes requested):
Good work so far! Some comments inline.

I've mentioned it at some point before, and I still think we have to implement https://github.com/PostHog/posthog-plugin-server/issues/62 somehow within this PR. Basically we just need a `pause` and `resume` method on the kafka queue.

If you'd like, I can have a look at implementing these two methods into celery myself...

src/worker/queue.ts
Twixes: With an interval we can just `return` immediately if some `server.isQueueFull` (or something like that) is true, so that should be simple enough.

mariusandra: Indeed, that's a simple and clever workaround that will block the explosion. The next step will then still be to use `process.nextTick` and immediately get more events once the queue frees up.

src/ingestion/process-event.ts
Twixes: 👍 Can clean that up.

src/types.ts
Twixes: I'll look into it, but the way the plugin server handles this sort of type doesn't seem entirely consistent with the rest of PostHog (which isn't really consistent with itself at all, given all the legacy stuff and cruft…), so not sure.

src/worker/queue.ts
Twixes: It definitely should, though I don't think we run it anywhere inside the Celery pipeline at all yet? I'll give it a go here though.

mariusandra: Yeah, we don't use it for Celery as celery only returns events one by one. The `processEventBatch` function was basically made with Kafka in mind. It might be wise to implement some batching in front of celery as well, to for example batch all events every 100ms, up to 100 per batch, but that's for another day.

mariusandra:
Forgot one point. Would be great to have some tests as well for the kafka parts :).

src/types.ts
Twixes: It seems to make sense actually. Done.

src/worker/queue.ts
Twixes: There is a potential issue, but it's actually different, as this is an at-least-once scenario – that is most events will be processed once, but occasionally more than once (so no data loss, the opposite). Exactly-once (like, _exactly_ exactly) consumption in Kafka is not a straightforward thing and requires transactions, which `node-rdkafka` doesn't support it seems.

mariusandra:
The code looks good to me, but unfortunately I can't run anything. I'm having issues with docker-compose (some networking issue - the "db" hostname can't be resolved in the web/worker/etc containers.... 🤮)

Some things to still add or change:

- We now support `posthog.capture` in the plugins. It just dispatches a celery task with the captured event. Can we have something similar for Kafka? So that it just adds an event to the back of the queue, which we'll later read again and process.

- Jest segfaults. Segfaults usually happen when C code tries to access memory or instructions that it doesn't have access to. I'm guessing the `librdkafka` c extension still tries to do something after the test finishes, but probably has nobody to report to. Something is probably not getting cleaned up properly. I'm just guessing here. It could be something completely different. In any case our test suite shouldn't segfault :).

- Can you remove the `console.log` in process event? We will be hooking this up to a big pipe soon and that'll just be a lot of noise.

- I'd still love to see some tests for the kafka part... but we can push this to a new issue to move fast here.

mariusandra:
Trying to debug this via github codespaces (where kafka also segfaults) and I have learned that if I disable this line:

```js
// src/ingestion/kafka-queue.ts
import { KafkaConsumer, ... } from '@posthog/node-rdkafka'
```

then all the tests will run. 

Even if the KafkaConsumer is never used, simply importing it makes all tests fail when they call `piscina.runTask`.

To be continued.

Twixes:
That seems somehow related to Node bindings for C(++) compatibility with Workers (which `node-rdkafka` is). Made a fix to package registration (https://github.com/Blizzard/node-rdkafka/pull/857) but maybe there's more to it? Not sure

mariusandra:
This seems very related indeed. Jest could also be messing with us here, since it uses workers to run tasks in parallel (even though we disabled this). We do need a proper fix to it before merging this though :/

Twixes:
Replaced `node-rdkafka` with `kafkajs` and the whole things works A LOT smoother! No more segfaults, a more robust API, no issues with winding down, and it's all more transparent than bindings to C. The development runtime `ts-node-dev` with autoreload works great now!

src/extensions/posthog.ts
```diff
@@ -9,26 +12,67 @@ export interface DummyPostHog {
...
-        client.sendTask(
-            'posthog.tasks.process_event.process_event_with_plugins',
-            [distinctId, null, null, data, pluginConfig.team_id, new Date(), new Date()],
-            {}
-        )
-    }
+    const producer = server.EE_ENABLED ? server.kafka!.producer() : null
+    producer?.connect()
+
+    const sendEvent = server.EE_ENABLED
```
mariusandra, line 18: nit, but with `? :` blocks that huge, it's worth using an if loop, especially due to this `producer` stuff in front

src/extensions/posthog.ts
```diff
@@ -9,26 +12,67 @@ export interface DummyPostHog {
...
+                  timestamp: utcNow,
+                  properties: {
+                      $lib: 'posthog-plugin-server',
+                      $lib_version: version,
+                      ...properties,
+                  },
+              }
+
+              if (server.EE_ENABLED) {
+              } else {
```
mariusandra, line 68: useless if

src/server.ts
```diff
@@ -51,12 +70,12 @@ export async function createServer(
         }
     }
 
-    const server: PluginsServer = {
+    const server: Omit<PluginsServer, 'eventsProcessor'> = {
```
mariusandra, line 73: This feels hacky. Why not make it optional on the type?

mariusandra (approved):
I basically have some nits/comments. Good to merge otherwise!

mariusandra:
Also, I'd still really love some tests for the kafka queue... but let's do that later then.

src/server.ts
Twixes: It's just omitted for initialization here, but we do want it required on the type.

PR #75, opened 2020-12-15
Celery queue backpressure

mariusandra:
- Pauses celery if we are busy processing all the tasks we got so far
- Resumes when the workers are free again.
- Also implements a 50ms wait time for celery to receive the next event if none were received in the last round

I added a `Pausable` interface and wrote some more ugly code in `server.ts` that I'm not immensely proud of. I think the entire `server.ts` file needs to be refactored (into a class perhaps?) and I'd fix all this then.

I see that your kafka PR had a generic `Queue` interface, which I think is even better. Feel free to adjust as needed if/when you merge these things.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -74,10 +74,15 @@ export async function createServer(
     return [server, closeServer]
 }
 
+// TODO: refactor this into a class, removing the need for many different Servers
+type ServerInstance = {
+    stop: () => Promise<void>
+}
+
 export async function startPluginsServer(
-    config: PluginsServerConfig,
+    config: Partial<PluginsServerConfig>,
     makePiscina: (config: PluginsServerConfig) => Piscina
-): Promise<void> {
+): Promise<ServerInstance> {
     console.info(`⚡ posthog-plugin-server v${version}`)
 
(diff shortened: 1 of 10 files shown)


src/server.ts
```diff
@@ -135,13 +143,21 @@ export async function startPluginsServer(
...
+        }
 
         if (!server.DISABLE_WEB) {
             fastifyInstance = await startFastifyInstance(server)
         }
 
         queue = startQueue(server, processEvent)
+        piscina.on('drain', () => {
+            queue?.resume()
+        })
```
Twixes, line 160: This should be in startQueue, since we'll want different `drain` handlers in the Redis and Kafka `startQueue`s.

Twixes (approved):
This makes sense, though definitely the need for a refactor is increasing…

src/server.ts
mariusandra: Hmm... sure about that? The `pause` code is just a few lines above and I think just getting `pause`/`resume` called on it is all the queue needs to know... it shouldn't really know much about what is `piscina` I think.

I'll merge this for now and we can refactor this when merging the Kafka PR.

PR #77, opened 2020-12-16
Vendor libs

mariusandra:
- make `setupPlugin` async
- add `google.cloud.bigquery` to available libraries

### DIFF
--- a/src/extensions/google.ts
+++ b/src/extensions/google.ts
@@ -1,0 +1,17 @@
+import * as BigQuery from '@google-cloud/bigquery'
+
+type DummyCloud = {
+    bigquery: typeof BigQuery
+}
+
+type DummyGoogle = {
+    cloud: DummyCloud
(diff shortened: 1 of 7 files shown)


src/extensions/google.ts
```diff
@@ -0,0 +1,17 @@
...
+
+type DummyCloud = {
+    bigquery: typeof BigQuery
+}
+
+type DummyGoogle = {
+    cloud: DummyCloud
+}
+
+export function createGoogle(): DummyGoogle {
```
Twixes, line 11: Wow, people invested millions of dollars to create Google and here we are, doing the same with just one function!

Twixes (approved):
Seems straightforward.

src/extensions/google.ts
mariusandra: rookies

PR #78, opened 2020-12-16
Change to @posthog/hot-shots

mariusandra:
Just to get rid of the optional dependency errors that pollute up logs and confuse everyone that looks at them.

All changes here: https://github.com/PostHog/hot-shots


Twixes (commented):
Wouldn't just slapping option `--ignore-optional` on `yarn install` solve it?

PR #79, opened 2020-12-17
Run scheduler on only one instance

mariusandra:
- Slight refactor, moving scheduling stuff to `services/schedule.ts`
- Add `node-redlock` and run scheduled tasks only when we hold a lock
- Still missing tests for this


Twixes:
Some merge conflict resolution needed here.

mariusandra:
Finally ready for a review!

In the test for this I had to use the real redis (instead of ioredis-mock) because the latter wouldn't support `evalsha` (eval a Lua script by passing in its SHA hash...), which node-redlock uses. Thus I removed the redis mock from the global scope and only added it to those tests that actually used it (queue.test.ts and worker.test.ts)... and later even removed it from worker.test.ts, as it started to have some issues. I also had to tweak the tests a bit, since using `redis.get` on a list returned the list with `ioredis-mock`, but gave an error with `ioredis`.

src/services/schedule.ts
```diff
@@ -0,0 +1,136 @@
...
+                    return
+                }
+                try {
+                    lock = await lock.extend(lockTTL)
+                    lockTimeout = setTimeout(extendLock, extendDelay)
+                } catch (error) {
+                    console.error('🔴 Redlock can not extend lock!', error)
+                    Sentry.captureException(error)
+                    weHaveTheLock = false
+                    lockTimeout = setTimeout(tryToGetTheLock, 0)
```
Twixes, line 58: Why the 0 ms timeouts? Also I think the realistic minimum is like 4 ms?

mariusandra: I think this could actually be replaced by `process.nextTick`. The idea is to call this function as soon as possible, but outside the current stack. 4ms sounds quite random number to put in code though.

Twixes: This: https://stackoverflow.com/a/7221609 is pretty old but suggests that in Node there is no minimum like in browsers and in fact 0 is ran right away (though not _quite_, so maybe it does work the intended way here? I wouldn't totally bet on it though, `nextTick` sounds better for the job)

mariusandra: Indeed, I want it to run right away, but async, so it's basically the same as [`process.nextTick`](https://nodejs.org/api/process.html#process_process_nexttick_callback_args). I chose `setTimeout` though just to be consistent with this `lockTimeout = ...` system, for the odd case it will get cancelled before it starts. Changed to `nextTick` now.

Twixes: Hm, still got `lockTimeout = setTimeout(tryToGetTheLock, 0)` though, do we want that `nextTick`ified as well or not?

Twixes: Let's go back to `setTimeout` actually https://dev.to/logicmason/settimeout-vs-setimmediate-vs-process-nexttick-3lj2

PR #84, opened 2020-12-18
Enhance logging

### LINKED ISSUE #65: Rework logging to show thread ID
With multithreading it's confusing which `console.*` message comes from where. A nice addition would be prefixing e.g. `[MAIN] 😮 Something happened` for the main thread and e.g. `[0003] 🍆 A thing started` for thread ID 3.
Also, a silly but handy thing would be including the emoji as a separate parameter. Some devs hate them, but I think they really highlight things in logs. 😂
So something like:
```TypeScript
import { threadId } from 'worker_threads'

function info(message: string, emoji?: string): void {
    console.info(`[${threadId?.toString().padStart(4, '0') ?? 'MAIN'}]${emoji ? ' ' + emoji : ''} ${message}`)
}
```

Twixes:
This resolves #65.

### DIFF
--- a/src/celery/worker.ts
+++ b/src/celery/worker.ts
@@ -1,6 +1,7 @@
 import { Queue } from '../types'
 import Base from './base'
 import { Message } from './message'
+import { status } from '../status'
 
 type Handler = (...args: any[]) => Promise<void>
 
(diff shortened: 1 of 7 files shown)


mariusandra (approved):
Good for me! Some conflict to fix though.

PR #87, opened 2021-01-05
Plugin storage

### LINKED ISSUE #74: Persistent storage
The [github star sync plugin](https://github.com/PostHog/posthog/issues/1896#issuecomment-744424515) needs to store state (`last read page`, `last ingested time`) somewhere.

The only thing available is `cache` (redis), which should **not** be used as persistent storage. 

A solution is needed.

I wouldn't want to introduce new services and databases (mongodb, anyone?), so the easiest thing is probably a postgres-backed `localStorage` API that you can call from within plugins.

mariusandra:
- Depends on https://github.com/PostHog/posthog/pull/2852
- Adds `storage.get(key, defaultValue)` and `storage.set(key, value)` into meta, which act as buckets to store random JSON data
- Useful for scheduled and other plugins to store state
- I removed the postgres mock and made all tests use a real DB (defaults to "posthog_test"). The piscina worker threads anyway can't be mocked and need a real postgres and a real redis to be present, so why not for the main app as well. This makes the test env closer to production and makes testing easier. Side-effect, must run `createdb postgrest_test && yarn task:db-init` locally before running test.
- Closes #74


tests/vm.test.ts
```diff
@@ -626,17 +661,18 @@ test('attachments', async () => {
 
 test('runEvery', async () => {
     const indexJs = `
-        function runEveryMinute(meta) {
+        function runEveryMinute (meta) {
```
Twixes, line 664: Why did the space appear here? 🤔

tests/helpers/worker.ts
```diff
@@ -2,14 +2,12 @@ import Piscina from 'piscina'
 import { makePiscina } from '../../src/worker/piscina'
 import { defaultConfig } from '../../src/config'
 import { LogLevel } from '../../src/types'
-import { mockJestWithIndex } from './plugins'
 
-export function setupPiscina(workers: number, code: string, tasksPerWorker: number): Piscina {
+export async function setupPiscina(workers: number, tasksPerWorker: number): Promise<Piscina> {
```
Twixes, line 6: Why `async` if it doesn't `await`?

package.json
```diff
@@ -18,7 +18,7 @@
         "prettier": "prettier --write .",
         "prettier:check": "prettier --check .",
         "prepublishOnly": "yarn test && yarn build",
-        "task:db-init": "ts-node ./tasks/db-init.ts"
+        "task:db-init": "NODE_ENV=test ts-node ./tasks/db-init.ts"
```
Twixes, line 21: Why mandatory NODE_ENV=test?

src/config.ts
```diff
@@ -4,9 +4,11 @@ export const defaultConfig = overrideWithEnv(getDefaultConfig())
 export const configHelp = getConfigHelp()
 
 export function getDefaultConfig(): PluginsServerConfig {
+    const isTestEnv = process.env.NODE_ENV === 'test'
```
Twixes, line 7: Ah, we probably should actually use `NODE_ENV` instead of `areWeTestingWithJest()`'s `JEST_WORKER_ID` 🤔

Twixes (commented):
Just a couple of Qs.

tests/vm.test.ts
mariusandra: To be consistent with other function declarations and prettier (which doesn't format inside strings).
Basically I was CMD+F'ing and couldn't find every place this `runEveryMinute` was created.

tests/helpers/worker.ts
mariusandra: it did await at some point, but I removed it since and forgot to update the function. thanks for catching!

src/config.ts
mariusandra: yeah, fixed!

package.json
mariusandra: Hmm... I basically put it so I wouldn't wipe my localhost `posthog` database, but would instead create tables in the `posthog_test` database. This just made running the script a bit easier on the command line.

The only time you would ever run the `create table` command within `posthog-plugin-server` is in a test environment. Otherwise all of this is handled by django. So as an added security I think it's fine... 🤷

Twixes: Ah, makes sense, the script could maybe be named to indicate that it's only for testing.

mariusandra: sure, renamed to `init-test-db.ts` and fixed the merge conflicts

Twixes (changes requested):
Tests red 🆘

mariusandra:
Fixed with an appropriate commit message.

Twixes (approved):
Tests green ❇️

PR #90, opened 2021-01-07
Read config from package.json

### LINKED ISSUE #13: Move from `plugin.json` to `package.json`

Twixes:
This resolves #13. Adds support for `package.json`-based configs (_preferably_ with ~~`posthog-plugin-config`~~ `configSchema` instead of `config` and with conventional `homepage` instead of `url`), while making no breaking changes.

### DIFF
--- a/src/plugins.ts
+++ b/src/plugins.ts
@@ -105,21 +105,39 @@ async function loadPlugin(server: PluginsServer, pluginConfig: PluginConfig): Pr
     try {
         if (plugin.url?.startsWith('file:')) {
             const pluginPath = path.resolve(server.BASE_DIR, plugin.url.substring(5))
-            const configPath = path.resolve(pluginPath, 'plugin.json')
 
             let config: PluginJsonConfig = {}
-            if (fs.existsSync(configPath)) {
+            let wasConfigLoaded = false
+
+            for (const jsonFileName of ['package.json', 'plugin.json']) {
+                const configPath = path.resolve(pluginPath, jsonFileName)
                 try {
-                    const jsonBuffer = fs.readFileSync(configPath)
-                    config = JSON.parse(jsonBuffer.toString())
-                } catch (e) {
-                    await processError(
-                        server,
-                        pluginConfig,
-                        `Could not load posthog config at "${configPath}" for plugin "${plugin.name}"`
-                    )
-                    return false
+                    const rawConfigUpdate = fs.readFileSync(configPath).toString()
+                    let configUpdate: any
+                    try {
+                        configUpdate = JSON.parse(rawConfigUpdate)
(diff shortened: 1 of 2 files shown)


mariusandra:
Few thoughts:
1. Some test is breaking.
2. There's more work to be done, as the posthog app itself also reads the github/npm packages and [fetches information from the plugin.json it finds from there](https://github.com/PostHog/posthog/blob/259e801de601775f8f2d94517bd4d39f8c180b4e/posthog/api/plugin.py#L97)
3. I've somehow been stalling with these tasks, as it might make sense to take one step back and re-evaluate how we want to distribute plugins in the first place. We have 3 methods now: 1) source copypasta, 2) github, 3) npm... all with slightly different ways to configure them. We should clarify, document or consolidate all these ways before adding more permutations. I'll make a new issue regarding this shortly.

Twixes:
Alright, so:

1. Fixed.
2. That work is done in PostHog/posthog#2928.
3. I think we decided not to make fundamental changes to installation methods, though correct me if I'm wrong as I'm actually unsure what conclusion we had from the meeting.

src/plugins.ts
```diff
@@ -105,23 +105,47 @@ async function loadPlugin(server: PluginsServer, pluginConfig: PluginConfig): Pr
...
+                    try {
+                        configUpdate = JSON.parse(rawConfigUpdate)
+                    } catch {
+                        await processError(server, pluginConfig, `File ${jsonFileName} is not valid JSON`)
+                        return false
+                    }
+                    if (configUpdate['posthog-plugin-config']) {
+                        // plugin.json has `config`, but in package.json we prefer `posthog-plugin-config` for clarity
+                        configUpdate.config = configUpdate['posthog-plugin-config']
+                        delete configUpdate['posthog-plugin-config']
```
mariusandra, line 126: Same stuff here, we should load `configSchema` instead of `posthog-plugin-config`.

Also, I think instead of this update system, it's probably fine to just 1) read plugin.json, 2) if not found, read package.json.

Twixes:
We won't be going forward with this as per PostHog/plugin-server#13.

PR #101, opened 2021-01-15
Containerize

### LINKED ISSUE #95: Create Dockerfile
This should be dockerized for simplicity of spinning up with docker-compose setups.

Twixes:
## Changes

This containerizes the plugin server, which will allow us to pull it from DockerHub in `docker-compose.ch.yml` or from Amazon ECR in PostHog Cloud. MUCH cleaner solution than running the plugin server inside the main repo.
Closes #95.

**Don't merge before #34.**

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -3,6 +3,5 @@
 !yarn.lock
 !tsconfig.json
 !src
-!bin
 !.eslintrc.js
 !.prettierrc
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,8 @@
+*
+!package.json
+!yarn.lock
+!tsconfig.json
+!src
+!bin
+!.eslintrc.js
+!.prettierrc
--- a/.github/workflows/containerization.yml
+++ b/.github/workflows/containerization.yml
@@ -21,15 +21,18 @@ jobs:
               run: |
                   PACKAGE_VERSION=$(node -p "require('./package.json').version")
                   PACKAGE_VERSION_TAG="v$PACKAGE_VERSION"
+                  DOCKER_TAGS="posthog/plugin-server:latest,ghcr.io/posthog/plugin-server:latest,public.ecr.aws/p1o5l3m0/posthog-plugin-server:latest"
                   if git rev-parse "$PACKAGE_VERSION_TAG"; then
                     IS_NEW_VERSION=false
                   else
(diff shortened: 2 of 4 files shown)


mariusandra:
I'm totally OK with packaging this repo into a docker image, but I have some worries about using this image in our docker-compose files, helm charts, cloudformation templates and other deployments.

I really liked (and optimised for) the KISS approach of everything just being in a `posthog/posthog` mono-container (a [majestic monolith](https://m.signalvnoise.com/the-majestic-monolith/) if you will) and then running whatever script is needed to start the service that's requested.

The big issue we will run into is versioning. We either always need to remember (e.g. when updating the helm chart) to update the version of both the `posthog/posthog` image and the `posthog/plugin-server` image (and double checking to make sure it's the right plugin server version that's in plugins/package.json).... or if we have deployments tagged to `latest` or `latest-release` on either both the main and/or plugin container, we need to always make sure they are compatible with each other.

For example, if we release a `latest` plugin server that requires the `posthog_pluginstorage` table to exist, yet the `latest` posthog container does not contain that migration, things will break in unexpected ways. The problem will be worse if we use `latest-release` tags, like we do with `posthog/posthog` in the cloudformation template. The posthog image will update every 2 weeks, but the plugin-server will possibly update more frequently and its latest release can be incompatible with the latest release of posthog. Thus people's deployments will start failing randomly and we can't allow that.

So I'd say fine for packaging this as a separate docker image, but the main app needs to be tagged to a specific plugin-server version. The easiest way to do this is to just update the plugins/package.json file inside the main repo and use that directly, in all deployments.

Twixes:
100% agree we should always pin to version in users' production setups. Should be fine this way.

mariusandra:
Where would you use this plugin-server docker image? You mentioned ee/docker-compose.ch.yml, but that's just something we use locally for testing... and I don't think it's wise to update a number in this file every time we release a new version...

Twixes:
Any containerized production setup. Starting with PostHog Cloud.

mariusandra:
Then I still don't understand how do you intend to keep the versions in sync? I'm my understanding this adds a lot of complexity and manual steps with dubious benefits.

Twixes:
Well well this is actually done now, for better iteration when preparing the plugin server for full-fledged deployment on Cloud.

PR #106, opened 2021-01-19
Send custom timestamps

mariusandra:
## Changes

Turns out this wasn't working after all. Now it's doing the same as the `posthog-js` library --> removing the `timestamp` property and sending it as the event's timestamp.

## Checklist

-   [x] Jest tests


Twixes (approved):
Ah, nice. I _was_ a bit suspicious about the `const utcNow = DateTime.utc().toISO()` bits earlier when you raised this use case, but didn't really check it. With a test that's a slam dunk though.

PR #107, opened 2021-01-20
Release with npm instead of yarn

Twixes:
## Changes

Should resolve https://github.com/PostHog/plugin-server/commit/c5e05be5331af29a10af2642af6a1ce4cb3b0c26#r46144662.


mariusandra:
This eslint thing is weird and broke out of the blue it seems. When I run prettier on these files it adds the semicolons back leading into an endless cycle?

mariusandra (approved):

PR #109, opened 2021-01-20
Better StatsD metrics

Twixes:
## Changes

This is some better metrics, will think about some more too. Suggestions in this regard welcome! @mariusandra


mariusandra (approved):
Luuks greit.

PR #110, opened 2021-01-21
Make some thing nicer

Twixes:
## Changes

Just a bunch of things cleaned up.

### DIFF
--- a/src/ingestion/kafka-queue.ts
+++ b/src/ingestion/kafka-queue.ts
@@ -91,18 +91,18 @@ export class KafkaQueue implements Queue {
         if (!this.wasConsumerRan || this.isPaused()) {
             return
         }
-        console.error('⏳ Pausing Kafka consumer...')
+        status.info('⏳ Pausing Kafka consumer...')
         await this.consumer.pause([{ topic: KAFKA_EVENTS_WAL }])
-        console.error('⏸ Kafka consumer paused!')
+        status.info('⏸ Kafka consumer paused!')
     }
 
     async resume(): Promise<void> {
         if (!this.wasConsumerRan || !this.isPaused()) {
             return
         }
-        console.error('⏳ Resuming Kafka consumer...')
+        status.info('⏳ Resuming Kafka consumer...')
         await this.consumer.resume([{ topic: KAFKA_EVENTS_WAL }])
-        console.error('▶️ Kafka consumer resumed!')
+        status.info('▶️ Kafka consumer resumed!')
     }
 
     isPaused(): boolean {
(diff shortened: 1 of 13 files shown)


src/ingestion/kafka-queue.ts
```diff
@@ -80,24 +84,25 @@ export class KafkaQueue implements Queue {
             },
         })
         this.wasConsumerRan = true
+        return promise
```
mariusandra, line 87: Why not await the promise here as well? I mean, in the end it works the same way (this `queue.start` is always directly awaited if I see correctly?), but just `await`ing here would show intent a bit better. Maybe I'm missing something :)

src/status.ts
```diff
@@ -1,23 +1,21 @@
...
 export const status: Status = {
-    info(emojiOrMessage: string, message?: string) {
-        console.info(transform(emojiOrMessage, message))
+    info(icon: string, ...message: any[]) {
+        console.info(getPrefix(), icon, ...message.filter(Boolean))
     },
-    error(emojiOrMessage: string, message?: string) {
-        console.error(transform(emojiOrMessage, message))
+    error(icon: string, ...message: any[]) {
+        console.error(getPrefix(), icon, ...message.filter(Boolean))
```
mariusandra, line 19: As it is now and looking at the usage, why not just remove the `icon` argument?

mariusandra (changes requested):
Two comments, but not really blockers. Also tests fail (blocker).

src/status.ts
Twixes: It's just for verbosity really.

mariusandra: It's a bit confusing if not used as intended. The entire message is said to be an icon...
image

src/ingestion/kafka-queue.ts
Twixes: Hm, alright. Also put in all of start() into the promise while at it.

Twixes: Actually also put `eachBatch` in its own `KafkaQueue` method, much cleaner this way.

mariusandra: Hm... is this 100% correct though? Could be, but looking at it:
- Is there a chance the callbacks run and the promise resolves before the `wasConsumerRan = true` line, causing it not to run? Maybe Node will still run that code, but I'm not sure.
- Those `on` callbacks. They're not cleared in `stop()`, which is used when reloading the server. Could they linger on and cause issues?

src/status.ts
Twixes: That's a few oversights, fixed. :)

src/ingestion/kafka-queue.ts
Twixes: * Overall the method should be just fine, just moved that `wasConsumerRan = true` before `consumer.run()`
* Not really an issue, they don't do anything besides resolving/rejecting a promise, which is idempotent. Possibly a tiny tiny memory leak in case of an unexpectedly large number of reconnects? But unlikely to be a problem.

mariusandra: the plugin server might have to reload often, so this should be fixed somehow, probably...

PR #114, opened 2021-01-26
Revamp CI to use actual Django test DB

Twixes:
## Changes

Major change in the way tests are set up, extracted from #34 for clarity. We need this for the plugin server to work reliably in tandem with the Django server when the latter's DB schema inevitably undergoes changes, particularly for comprehensive event processing.

## Checklist

-   [x] Jest tests


mariusandra (approved):
Some missing `'organization_id' is declared here` causing the benchmark to break, otherwise looks good!

PR #116, opened 2021-01-28
Postgres ingestion + process event tests

mariusandra:
## Changes

- Adds these tests from python:
image

## Checklist

-   [ ] Jest tests


Twixes:
`Client` is now called twice because of initializing in `EventsProcessor` for the webhooks Celery task.

Screen Shot 2021-01-28 at 11 39 43

src/server.ts
```diff
@@ -19,6 +19,19 @@ import { status } from './status'
...
+// postgres will only return strings as dates
+types.setTypeParser(1083 /* types.TypeId.TIME */, (timeStr) =>
+    timeStr ? DateTime.fromSQL(timeStr, { zone: 'utc' }).toISO() : null
+)
+types.setTypeParser(1114 /* types.TypeId.TIMESTAMP */, (timeStr) =>
+    timeStr ? DateTime.fromSQL(timeStr, { zone: 'utc' }).toISO() : null
+)
+types.setTypeParser(1184 /* types.TypeId.TIMESTAMPTZ */, (timeStr) =>
+    timeStr ? DateTime.fromSQL(timeStr, { zone: 'utc' }).toISO() : null
+)
```
Twixes, line 34: This looks interesting, but not sure what it does 🤔

src/utils.ts
```diff
@@ -286,7 +286,7 @@ export function castTimestampOrNow(timestamp?: DateTime | string | null): string
     } else if (typeof timestamp === 'string') {
         timestamp = DateTime.fromISO(timestamp)
     }
-    return timestamp.toUTC().toFormat('yyyy-MM-dd HH:mm:ss.u')
+    return timestamp.toUTC().toISO()
```
Twixes, line 289: Hmm, isn't the old format required by ClickHouse? At least that's what original ingestion uses

mariusandra: Hm... would be great to have E2E clickhouse tests as well to verify this works :). 
I swapped this our since `.toISO()` contains a `Z`, indicating the UTC timezone. The `.toFormat` version contains no timezone information and something at some point re-parses it with my local timezone (`CET`), resulting in some timezone process event test failing.

src/server.ts
mariusandra: Added some clarification and moved it around

src/utils.ts
Twixes: So after this change this will only work with CH if CH has `date_time_input_format` set to `best_effort` instead of the default `basic`. https://clickhouse.tech/docs/en/operations/settings/settings/#settings-date_time_input_format
Originally this function was only used in EE ingestion in Python, but I think w can repurpose it by adding parameter `iso = false` (or something like that) and using it where applicable

mariusandra: How about this? https://github.com/PostHog/plugin-server/pull/116/commits/137dcde9e674b1f760543715e360177d6093f187

mariusandra:
I think this can be merged into the `ingestion-save` branch. I'm going to stop working on it for today (and this sprint). There are too many things to do before postgres ingestion is ready and all the tests pass, which would be the prerequisite for merging this to master. 

Some new things to keep in mind:
- Make sure everyone who uses nodejs ingestion has ASYNC_EVENT_ACTION_MAPPING set to true. Otherwise no actions will be matched
- Create Elements & Session Recording Events in nodejs
- Other random bugfixes

mariusandra:
@Twixes this is ready for a look and to be merged back inside `ingestion-save`. All the postgres tests are ported over and pass:

image

I was quite flexible with the helper methods for the tests. Feel free to move them around to wherever they should go.

mariusandra:
Also, these tests were quite essential to getting ingestion closer to being merged. Writing them helped catch a few hard-to-track bugs, which are now squashed. Now we just need them to work for clickhouse as well :).

Twixes (approved):
Tests well ported

PR #117, opened 2021-01-28
Reorder plugins per team based on `order`

mariusandra:
## Changes

- Takes the `order` of plugins into account per team
- Still keeps support for the `defaultConfigs`, though we might just remove that in the future.

## Checklist

-   [ ] Jest tests

### DIFF
--- a/src/plugins.ts
+++ b/src/plugins.ts
@@ -83,14 +83,16 @@ export async function setupPlugins(server: PluginsServer): Promise<void> {
         }
     }
 
-    if (server.defaultConfigs.length > 0) {
-        server.defaultConfigs.sort((a, b) => a.order - b.order)
-        for (const teamId of Object.keys(server.pluginConfigsPerTeam).map((key: string) => parseInt(key))) {
-            server.pluginConfigsPerTeam.set(teamId, [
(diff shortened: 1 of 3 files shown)


src/plugins.ts
```diff
@@ -83,14 +83,16 @@ export async function setupPlugins(server: PluginsServer): Promise<void> {
         }
     }
 
-    if (server.defaultConfigs.length > 0) {
-        server.defaultConfigs.sort((a, b) => a.order - b.order)
-        for (const teamId of Object.keys(server.pluginConfigsPerTeam).map((key: string) => parseInt(key))) {
-            server.pluginConfigsPerTeam.set(teamId, [
+    const sortFunction = (a: PluginConfig, b: PluginConfig) => (a.order || 0) - (b.order || 0)
```
Twixes, line 86: Shouldn't `PluginConfig.order` always be a number?

mariusandra: It doesn't have `not null` in the database, so 🤷 .

Twixes: So, it will with https://github.com/PostHog/posthog/pull/3122

mariusandra: ok, fixed then here as well

PR #120, opened 2021-02-01
Ingestion postgres e2e test

mariusandra:
## Changes

- Adds (almost) E2E test for postgres ingestion
- Introduces a new variable, `PLUGIN_SERVER_INGESTION` to enable/disable postgres ingestion when no kafka. 

This variable is very close to the `PLUGIN_SERVER_INGESTION_HANDOFF` env [from here](https://github.com/PostHog/posthog/pull/3107/files), but I decided to remove the `_HANDOFF` bit, since envs are shared between all services and I from the point of the plugin server, there's no `HANDOFF` anymore. This env will "make me think" if I'm inside the plugin server, as to who's handing off what to whom.... So I propose to change it also in [the PR](https://github.com/PostHog/posthog/pull/3107/files).

## Checklist

-   [ ] Jest tests

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -20,6 +20,7 @@ export function getDefaultConfig(): PluginsServerConfig {
         KAFKA_CLIENT_CERT_B64: null,
         KAFKA_CLIENT_CERT_KEY_B64: null,
         KAFKA_TRUSTED_CERT_B64: null,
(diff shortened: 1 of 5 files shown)


Twixes:
Very well, can rename to `PLUGIN_SERVER_INGESTION` everywhere

PR #122, opened 2021-02-02
Ingestion and e2e benchmarks

mariusandra:
## Changes

- **WIP!**
- Simple benchmarks to see how fast ingestion can be. 
- `ingestion.benchmark.ts` tests the speed of the `processEvent` function directly
- `e2e.benchmark.ts` does a full E2E test by 1) starting the full server with postgres + celery, 2) pause the queue, 3) load `N` events into celery, 4) resume the queue 5) count how long it takes until `N` events show up in the database. It also runs one simple plugin that changes a property.

Here are the results so far (on my 2019 16" MBP):

```
    Ingested 3000 events in 6.8s (2.27ms per event)
    Ingested 10000 events in 26.08s (2.61ms per event)
```

The next step is to add parallelism and see if these numbers improve.

## Checklist

-   [ ] Jest tests

### DIFF
--- a/.github/workflows/cd.yaml
+++ b/.github/workflows/cd.yaml
@@ -26,7 +26,7 @@ jobs:
                     IS_NEW_VERSION=false
                   else
                     IS_NEW_VERSION=true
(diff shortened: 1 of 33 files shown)


mariusandra:
Superseded by #132

PR #123, opened 2021-02-03
Optimize Kafka queue offset handling

Twixes:
## Changes

A way to optimize offsets and to avoid weird situations.

## Checklist

-   [ ] Jest tests

### DIFF
--- a/src/ingestion/kafka-queue.ts
+++ b/src/ingestion/kafka-queue.ts
@@ -20,10 +20,10 @@ export type BatchCallback = (messages: Message[]) => Promise<void>
  * ```
  * Now some plugin discards the last two! Returned map for use in resolveOffset() will in result look like this:
  * ```JS
- * { 'a': 1, 's': 4, 'd': 4, 'f': 4}
+ * { 'a': 1, 's': 4, 'd': 4, 'f': 4 }
  * ```
- * Because 'd' and 'f' were discarded by a plugin, when we save 's' to the database, we'll know that we also in fact
- * covered 'd' and 'f' and should resolve the last offset - belonging to 'f' – not an intermediary one.
+ * Because 'd' and 'f' were discarded by a plugin, when we save 's' to the database, we'll know that at the same time
+ * we in fact also covered 'd' and 'f' and should resolve the last offset - belonging to 'f' – not an intermediary one.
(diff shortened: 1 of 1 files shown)


src/ingestion/kafka-queue.ts
```diff
@@ -1,13 +1,48 @@
 import * as Sentry from '@sentry/node'
 import { Kafka, Consumer, Message, EachBatchPayload } from 'kafkajs'
 import { PluginsServer, Queue, RawEventMessage } from 'types'
-import { KAFKA_EVENTS_INGESTION_HANDOFF } from './topics'
+import { KAFKA_EVENTS_WAL } from './topics'
```
mariusandra, line 4: Why the change from `KAFKA_EVENTS_INGESTION_HANDOFF`?

mariusandra (commented):
I feel like this code is way too complex for what it should do. It also leaves the kafka_offset inside event.

I tried another way to get the same result with https://github.com/PostHog/plugin-server/pull/124 ... though I haven't run the code yet.

Twixes:
Oh, sry, actually I was doing some debugging due to tests failing, so this is not in a reviewable state ATM (was reviewable at 6feda22).

If this is too complex, what would your alternative suggestion be? This isn't super simple, but would cover all cases pretty well (and there are various ones we have to consider, e.g. some plugin: added events there never were in the Kafka queue, or removed a series of events that were in the Kafka queue, or removed all of them, or removed every other event, or removed all of them).

mariusandra:
Which cases would #124 not cover?

mariusandra:
just saw your reply :), will give it a thought after dinner

src/ingestion/kafka-queue.ts
```diff
@@ -71,7 +102,7 @@ export class KafkaQueue implements Queue {
             }
             const singleIngestionTimer = new Date()
             await this.saveEvent(event)
-            resolveOffset(event.kafka_offset!)
+            resolveOffset(eventUuidToKafkaOffset.get(event.uuid!)!)
```
Twixes, line 105: TODO: don't run this if the event was not present in batch from Kafka (meaning it was inserted by a plugin's `processEventBatch`)

PR #124, opened 2021-02-03
Resolve all kafka offsets

mariusandra:
## Changes

- Here's another attempt at #123, but with 4 times fewer changed lines.

## Checklist

-   [ ] Jest tests

### DIFF
--- a/src/ingestion/kafka-queue.ts
+++ b/src/ingestion/kafka-queue.ts
@@ -6,14 +6,12 @@ import { PluginEvent } from '@posthog/plugin-scaffold'
 import { status } from '../status'
 import { killGracefully } from '../utils'
 
-export type BatchCallback = (messages: Message[]) => Promise<void>
-
 export class KafkaQueue implements Queue {
     private pluginsServer: PluginsServer
     private kafka: Kafka
     private consumer: Consumer
     private wasConsumerRan: boolean
-    private processEventBatch: (batch: PluginEvent[]) => Promise<any>
+    private processEventBatch: (batch: PluginEvent[]) => Promise<PluginEvent[]>
     private saveEvent: (event: PluginEvent) => Promise<void>
 
     constructor(
@@ -39,28 +37,18 @@ export class KafkaQueue implements Queue {
         isStale,
     }: EachBatchPayload): Promise<void> {
         const batchProcessingTimer = new Date()
-        const rawEvents: RawEventMessage[] = batch.messages.map((message) => ({
-            ...JSON.parse(message.value!.toString()),
-            kafka_offset: message.offset,
-        }))
-        const parsedEvents = rawEvents.map((rawEvent) => ({
-            ...rawEvent,
-            data: JSON.parse(rawEvent.data),
-        }))
-        const offsetMap = new Map<string, string>()
-        const pluginEvents: PluginEvent[] = rawEvents.map((rawEvent) => {
-            const { data: dataStr, kafka_offset: kafkaOffset, ...restOfRawEvent } = rawEvent
-            const event = { ...restOfRawEvent, ...JSON.parse(dataStr) }
-            offsetMap.set(event.uuid, kafkaOffset)
+        const pluginEvents = batch.messages.map((message) => {
+            const { data: dataStr, ...rawEvent } = JSON.parse(message.value!.toString())
+            const event = { ...rawEvent, ...JSON.parse(dataStr) }
             return {
                 ...event,
                 site_url: event.site_url || null,
                 ip: event.ip || null,
             }
         })
-        const processedEvents: PluginEvent[] = (
-            await this.processEventBatch(pluginEvents)
-        ).filter((event: PluginEvent[] | false | null | undefined) => Boolean(event))
+
+        const processedEvents = await this.processEventBatch(pluginEvents)
+
         for (const event of processedEvents) {
             if (!isRunning()) {
                 status.info('😮', 'Consumer not running anymore, canceling batch processing!')
@@ -75,16 +63,10 @@ export class KafkaQueue implements Queue {
             const offset = event?.uuid ? offsetMap.get(event.uuid) : null
             if (offset) {
                 resolveOffset(offset)
-                offsetMap.delete(event.uuid!)
             }
             await heartbeat()
             await commitOffsetsIfNecessary()
             this.pluginsServer.statsd?.timing('kafka_queue.single_ingestion', singleIngestionTimer)
-        }
-        for (const removedOffset of offsetMap.values()) {
-            resolveOffset(removedOffset)
-            await heartbeat()
-            await commitOffsetsIfNecessary()
         }
         this.pluginsServer.statsd?.timing('kafka_queue.each_batch', batchProcessingTimer)
     }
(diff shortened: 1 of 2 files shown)


src/ingestion/kafka-queue.ts
```diff
@@ -71,11 +72,20 @@ export class KafkaQueue implements Queue {
...
+            }
             await heartbeat()
             await commitOffsetsIfNecessary()
             this.pluginsServer.statsd?.timing('kafka_queue.single_ingestion', singleIngestionTimer)
         }
+        for (const removedOffset of offsetMap.values()) {
+            resolveOffset(removedOffset)
+            await heartbeat()
+            await commitOffsetsIfNecessary()
+        }
```
Twixes, line 88: This changes almost nothing from the current situation, as it only resolves offsets of events that were discarded after the ones that were not discarded. It doesn't make much sense to resolve offsets completely out of order like this, only the last `resolveOffset` in this will even actually matter. We might as well remove all the `resolveOffset` calls anywhere in this method and just do it at the end with the last offset.
The reason for doing `resolveOffset` is to keep track of processed events INSIDE the `for (const event of processedEvents)` loop, so that if we have, say, 5 events, and for some reason the plugin server exits while processing event no. 3 and only manages to save 1 and 2 to CH, it will know to restart from event no. 3. The problem is, if in the current case if event no. 2 is discarded, its offset won't be resolved, and after restarting the server will restart from event no. 2, in effect processing it twice.
This change _doesn't_ improve that. It only fixes the problem in between batches if the last n events of the batch are discarded, which #123 handles too, except more deliberately.

Twixes (changes requested):

src/ingestion/kafka-queue.ts
mariusandra: So actually we should just enable `eachBatchAutoResolve: true`? From the docs:

> `eachBatchAutoResolve` configures auto-resolve of batch processing. If set to true, KafkaJS will automatically commit the last offset of the batch if `eachBatch` doesn't throw an error. Default: true.

Am I missing something? I see you had a long comment there about some dragons ahead. 

The way understand it, If we crash, it'll commit whatever was marked with `resolveOffset`. Otherwise it'll always commit the last offset of the batch, even if the last event was discarded. Sounds like what we need? In the worst case, when crashing at event "6" in the sequence "1", "2", "skip", "skip", "skip", "6", it'll retry events 3-5... but it's highly likely they will be skipped again... and I'm OK with more straightforward code here vs more accurate skipping. In practice, most batches contain just an event or two anyway.

Both #123 and this implemented one thing wrong though. The events might be returned from `await this.processEventBatch` in any order (and indeed with random events added to the mix). They could just be reversed. I now added some logic to ingest the events in the original order.

mariusandra: Oh and holy macaroni, all tests are green??

src/ingestion/kafka-queue.ts
```diff
@@ -39,27 +37,29 @@ export class KafkaQueue implements Queue {
...
-            await this.processEventBatch(pluginEvents)
-        ).filter((event: PluginEvent[] | false | null | undefined) => Boolean(event))
+
+        const processedEvents = await this.processEventBatch(pluginEvents)
+
+        // Sort in the original order that the events came in, putting any randomly added events to the end.
+        // This is so we would resolve the correct kafka offsets in order.
+        processedEvents.sort(
+            (a, b) => (uuidOrder.get(a.uuid!) || pluginEvents.length) - (uuidOrder.get(b.uuid!) || pluginEvents.length)
+        )
```
Twixes, line 61: Nice that the tacked on events are last!
Though as for the non-null assertion, as far as I understand, `processEventBatch` can return an array of practically anything, including non-objects falsy or truthy. 🤔

src/ingestion/kafka-queue.ts
```diff
@@ -88,9 +91,7 @@ export class KafkaQueue implements Queue {
             await this.consumer.subscribe({ topic: KAFKA_EVENTS_INGESTION_HANDOFF })
             // KafkaJS batching: https://kafka.js.org/docs/consuming#a-name-each-batch-a-eachbatch
             await this.consumer.run({
-                // TODO: eachBatchAutoResolve: false, // don't autoresolve whole batch in case we exit it early
-                // The issue is right now we'd miss some messages and not resolve them as processEventBatch COMPETELY
-                // discards some events, leaving us with no kafka_offset to resolve when in fact it should be resolved.
```
Twixes, line 93: This comment was a bit more cryptic than it should have been, but the concern was with potentially resolving offsets for unsaved events in two cases:

* if we return early because of the `isRunning` check - however now my understanding is that if the consumer is not running, then it won't be possible to commit offsets, not completely certain that though
* if we return early because of the `isStale` check - this may happen if `consumer.seek()` or something like that is used to move partition offset somewhere, this is a bit more concerning, though also I'm unsure about the specifics of such a situation

In the end though, just to control this behavior without any doubt (we should always prefer at-least-once over at-most-once), I'd rather uncomment this (seting `eachBatchAutoResolve` to `false`) and instead just resolve the offset deliberately.
Here's the relevant comment from KafkaJS source:

kafkajs/src/consumer/runner.js:240

src/ingestion/kafka-queue.ts
```diff
@@ -39,27 +37,29 @@ export class KafkaQueue implements Queue {
...
-            const { data: dataStr, ...restOfRawEvent } = rawEvent
-            const event = { ...restOfRawEvent, ...JSON.parse(dataStr) }
+
+        const uuidOrder = new Map<string, number>()
+        const uuidOffset = new Map<string, string>()
+        const pluginEvents = batch.messages.map((message, index) => {
+            const { data: dataStr, ...rawEvent } = JSON.parse(message.value!.toString())
+            const event = { ...rawEvent, ...JSON.parse(dataStr) }
+            uuidOrder.set(event.uuid, index)
+            uuidOffset.set(event.uuid, message.offset)
```
Twixes, line 47: OK, so this still doesn't address cases like: we've got events `a`, `b`, `c`; `b` and `c` gets discarded; we never resolve `b`'s and `c`'s offsets; in case of restart `b` and `c`, get processed again. But then also it's true that plugins can return whatever the hell they want in any order, so no method can be _completely_ trusted here (though in most cases there won't be shenanigans). Even sorting by UUIDs (which, the ones generated by us, are time-sortable with UUIDT), the `uuid` property may be randomized, so… Alright, whatever. In any case it's better to process an event at-least-once than at-most-once.

src/ingestion/kafka-queue.ts
mariusandra: It should always return this or crash: https://github.com/PostHog/plugin-server/pull/124/files#diff-5d37292cf21755714262793e7553ee317a12b157d0601e27a0ca26e096a60c00L267

Though, yeah, there's no real sanitization. The plugins could return just `{ "n00b": "lololololol" }` as an event and it would make its way here...

src/ingestion/kafka-queue.ts
mariusandra: Yep, this can go deep...

src/ingestion/kafka-queue.ts
Twixes: I suppose the `get` is going to return `undefined` in any such case, so this should be fine even if the assertion is too optimistic.
Side note, I think we make this assumption in some more dangerous places too, so probably will have to take a look at some event structure validation in another PR.

src/ingestion/kafka-queue.ts
Twixes: Done

Twixes (approved):

PR #127, opened 2021-02-04
Remove "_handoff" from Kafka topic

mariusandra:
## Changes

Since we're doing more than just handing off events with this topic (e.g. self-ingesting via posthog.capture), the name should be updated before all of this goes live. Changes to posthog itself will follow.

## Checklist

-   [ ] Jest tests


mariusandra:
Removed from posthog with: https://github.com/PostHog/posthog/pull/3107/commits/3609b83a9de2be74cdd082f744f8c5402e461fe6

mariusandra:
Merging because: 
image

PR #128, opened 2021-02-04
Switch ClickHouse driver

Twixes:
## Changes

This switches the ClickHouse driver for our own fork of @apla/clickhouse (which doesn't have TS support originally unfortunately). This library gives us more options, for instance providing our own TLS CA.

## Checklist

-   [x] Jest tests


src/db.ts
```diff
@@ -61,7 +63,7 @@ export class DB {
     public async fetchPersons(database: Database.ClickHouse): Promise<ClickHousePerson[]>
     public async fetchPersons(database: Database = Database.Postgres): Promise<Person[] | ClickHousePerson[]> {
         if (database === Database.ClickHouse) {
-            return (await this.clickhouseQuery('SELECT * FROM person')) as ClickHousePerson[]
+            return (await this.clickhouseQuery('SELECT * FROM person')).data as ClickHousePerson[]
```
mariusandra, line 66: Why not add the `.data` as part of `clickhouseQuery`? 

(same could be said for postgres...)

Twixes: Well, basically because we do practically the same thing for Postgres… Also, just in case we want to use something like `.statistics.elapse` some time

PR #133, opened 2021-02-05
Sort imports

Twixes:
## Changes

They were a bit of a mess, one ESLint plugin and things are much more orderly.


mariusandra:
Does this also sort them automatically on commit?

My imports normally look like this and I add/remove things automatically with the IDE:

image

It would be annoying to need to manually drag the things around :)

Twixes:
Yep, this is just ESLint, we run it on every changed `*.{js,ts}` file.

mariusandra:
This does mess up something in the tests:

image

PR #138, opened 2021-02-08
PgBouncer Test

mariusandra:
## Changes

- Totally WIP
- Adds a test to make sure everything works through pgbouncer

## Checklist

-   [ ] Jest tests


mariusandra:
Not sure if this will be valuable in the end and worth the effort. I wanted to test that the transactions work via pgbouncer, but there should be no problem, as [we are](https://github.com/PostHog/deployment/blob/2f85e68e11a9a4622f5bc695adf03ecb84fa20c0/aws/infra/services/pgbouncer/pgbouncer.tf#L132) in [transaction pooling mode](https://www.pgbouncer.org/features.html).

PR #132, opened 2021-02-05
Benchmark ingestion

mariusandra:
## Changes

- New version of #122 (to avoid a bad merge)
- Adds a simple "ingestion only" benchmark (measures `await processEvent`)
- Adds Kafka & Celery E2E ingestion benchmarks (from "posthog.capture" to event in postgres or clickhouse)
- The E2E benchmarks run one simple sync plugin and then add the result to the database. Benchmarks for plugins that e.g. wait for 20sec before doing anything will look wildly different and should be tested next
- Also adds some scripts to package.json to wipe the local kafka/clickhouse/zookeeper docker images

Results on my 16" 2.4GHz i9 MBP for the E2E benchmarks with 4 workers:

```
    [Celery & Postgres] Ingested 3000 events in 6.94s (432.18 events/sec, 2.31ms per event)
    [Celery & Postgres] Ingested 3000 events in 7.09s (423.41 events/sec, 2.36ms per event)
    [Celery & Postgres] Ingested 10000 events in 26.09s (383.24 events/sec, 2.61ms per event)
    [Celery & Postgres] Ingested 10000 events in 26.04s (384.08 events/sec, 2.6ms per event)

    [Kafka & ClickHouse] Ingested 3000 events in 25.62s (117.09 events/sec, 8.54ms per event)
    [Kafka & ClickHouse] Ingested 3000 events in 20.68s (145.06 events/sec, 6.89ms per event)
    [Kafka & ClickHouse] Ingested 3000 events in 19.85s (151.13 events/sec, 6.62ms per event)
    [Kafka & ClickHouse] Ingested 3000 events in 21.00s (142.83 events/sec, 7.00ms per event)
    [Kafka & ClickHouse] Ingested 10000 events in 143.94s (69.47 events/sec, 14.39ms per event)
    [Kafka & ClickHouse] Ingested 10000 events in 136.75s (73.12 events/sec, 13.68ms per event)
```

There's definitely room for improvement! 

## Checklist

-   [ ] Updated Settings section in README.md, if settings are affected
-   [ ] Jest tests

PR #144, opened 2021-02-11
Parallel Ingestion

mariusandra:
## Changes

- Adds parallel ingestion via a simple piscina task (but doesn't hook that up anywhere yet)
- Adds another noisy benchmark to show some numbers:

Calling `EventsProcessor.processEvent()` directly in the main thread, with `await` or `Promise.all`:

```
    [Postgres] Await Ingested: 3000 events, 4.25 sec, 706.15 events/sec, 1.42ms per event
    [Postgres] Promise.all Ingested: 3000 events, 2.18 sec, 1379.07 events/sec, 0.73ms per event
```

Running in multiple threads in piscina (numbers are events/sec): 

```json
    {
      "result": {
        "coreCount": 16,
        "events": 10000,
        "1 thread": 2061,
        "2 threads": 3035,
        "4 threads": 3654,
        "8 threads": 2306,
        "12 threads": 1921,
        "16 threads": 1694
      }
    }
```

Each piscina thread runs up to 10 ingestions in parallel.

Bumping the concurrency to 4 threads on an 8 core machine (that also runs postgres/pycharm/etc), it's possible to get 3654 events/sec instead of the 706.15 events/sec we get with sequential awaiting like now.

That gives us more leeway, but is not yet the full E2E picture.

Next step: making kafka more parallel. [This line](https://github.com/PostHog/plugin-server/blob/9838c5a8bd460d0560104541e7c2960be3d2a217/src/ingestion/kafka-queue.ts#L72) should somehow run in parallel...

## Checklist

-   [ ] Jest tests


mariusandra:
I added support for parallel ingestion of kafka messages. Basically just `Promise.all(e => ingestEvent(e))` and only committing the offset in the end. 

Without parallel ingestion
```
    [Kafka & ClickHouse] Ingested 3000 events (in 2 batches) in 22.23s (134.95 events/sec, 7.41ms per event)
```

With parallel ingestion
```
    [Kafka & ClickHouse] Ingested 3000 events (in 2 batches) in 7.29s (411.26 events/sec, 2.43ms per event)
```

If we discard the offsets and only commit batch by batch (like we do anyway for `await this.processEventBatch`, which might also have side effects, e.g. uploading an event twice to bigquery), then this is a `3x` improvement in ingestion speed. I think it's worth it.

mariusandra:
In other news, moving event ingestion into the worker thread increased the processing speed for postgres/celery from ([old stats](https://github.com/PostHog/plugin-server/pull/132#issue-568238138)):

```
    [Celery & Postgres] Ingested 3000 events in 6.87s (436.48 events/sec, 2.29ms per event)
```

to:

```
    [Celery & Postgres] Ingested 3000 events in 2.67s (1125.62 events/sec, 0.89ms per event)
```

That's another 2.5x win! I don't know if we could get much better than that with kafka/clickhouse, as this is with a case where all events were processed async and then ingested async. It's definitely something we could reach though.

mariusandra:
After chunking the batch of ~1500 events into chunks of ~40, Kafka E2E ingestion can now perform at this speed:

```
    [Kafka & ClickHouse] Ingested 3000 events in 5.72s (524.92 events/sec, 1.91ms per event)
```

... admittedly with a `do not much plugin`. This feels comfortable enough to enable for everyone.

Twixes (approved):

PR #146, opened 2021-02-12
Silence logs

mariusandra:
## Changes

- Silence runtime logs from within the plugin VMs (those are sent to postgres anyway)
- Remove sentry debug log
- Fix an issue with posthog-js-lite and some other small bugs
- Bump the version as well to `0.8.1` so that once merged, a new release is directly cut

## Checklist

-   [ ] Jest tests

### DIFF
--- a/src/ingestion/utils.ts
+++ b/src/ingestion/utils.ts
@@ -2,7 +2,6 @@ import * as Sentry from '@sentry/node'
 import crypto from 'crypto'
 
 import { BasePerson, Element, Person, RawPerson } from '../types'
-import Timeout = NodeJS.Timeout
 
 export function unparsePersonPartial(person: Partial<Person>): Partial<RawPerson> {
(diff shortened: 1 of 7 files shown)


mariusandra:
Added another [bit](https://github.com/PostHog/plugin-server/pull/146/commits/0b2b534f449b661f770bcd2ece177d0f5b63b0af) that should let us know if something gets stuck in Kafka.

src/ingestion/utils.ts
```diff
@@ -154,3 +155,10 @@ export function chainToElements(chain: string): Element[] {
 
     return elements
 }
+
+export function timeoutGuard(message: string): NodeJS.Timeout {
+    return setTimeout(() => {
+        console.log(`⌛⌛⌛ ${message}`)
+        Sentry.captureMessage(message)
+    }, 30000)
+}
```
Twixes, line 164: This is only a notification, not an exit from the function timing out, is that correct?

mariusandra: yeah, for now just so that we see anything

Issue #57, opened 2020-12-09
Defend against malicious plugins


mariusandra:
We should figure out what are the limits of the plugin system and find ways to protect against malicious plugins, deliberate or not. For example plugins that:
- Await for 300sec and then sends on the event
- Block the thread indefinitely with a long running for loop
- Take up 100GB of RAM
- Mint crypto
- etc

Ideally none of these cases should bring down the server and long running tasks should get killed in some way.

Twixes:
Digging for ideas: what can we do in this regard?

mariusandra:
We could set [resource limits](https://github.com/piscinajs/piscina/blob/current/examples/resourceLimits/index.js) to something sane to avoid workers using up all memory.

As for long running tasks, we can just capture and measure how long each task takes in the main thread. Piscina natively supports [aborting tasks](https://github.com/piscinajs/piscina/blob/current/examples/abort/index.js), but this won't help us here, as it only works when `concurrentTasksPerWorker` is `1`, but we set it higher in order to not block all tasks when a plugin is `async` waiting for an API to return a value.

We can keep track of how long any task takes and if we find something taking too long, we can terminate the entire worker (probably after some PRs to piscina to get this low level access). This would also terminate any other tasks that are blocked inside the worker thread (either also async'ing or just blocked because of a long running sync task that makes them queue up). I'm not sure there's much we can do about that, other than trying our best to retry.

We could/should also have some form of punishment for plugins that exceed resource limits (either time or memory) and prevent them from starting again after restarting the worker.

And all of this should come with tests obviously :).

There's probably more we can do, but these seem to be the lowest hanging fruit.

mariusandra:
The currently relevant issue #154 is directly connected to this issue. When `processEvent` gets stuck for long, it takes down the entire build pipeline. In the case of #154 it's probably due to an uncaught rejected promise, but it could be something else. Thus we need to implement some timeouts.

Here's an interesting and relevant article: https://medium.com/@bvjebin/js-infinite-loops-killing-em-e1c2f5f2db7f

mariusandra:
A lot of the common CPU-based attacks got handled with PR #155. Out of memory errors are still possible though.

mariusandra:
I made a new issue #159 against out of memory problems. Unless we discover something new, this issue is rather resolved.

PR #155, opened 2021-02-15
Babel Loop Timeouts

mariusandra:
## Changes

- All plugin JS is now processed via `@babel/standalone`. This unlocks cool stuff like native TS support in the future. We might also look into caching the processed code.
- I've added a babel plugin that injects code into all `for`, `while` and `do while` loops, as described in [this article](https://medium.com/@bvjebin/js-infinite-loops-killing-em-e1c2f5f2db7f) using a babel plugin modified from one published by [jsbin](https://github.com/jsbin/loop-protect). Basically we'll check at the beginning of the loop and throw an error if it has been running too long.

This greatly decreases our risk against some of the attacks described in #57 . This does not yet protect us against ~~long running promises nor~~ "out of memory" scenarios. These could be built in to this PR or done later with another PR, we'll see how it goes.

This might also have some implications and might need to be dialled down for scheduled plugins like `runEveryHour`, where you might actually want a long running for loop.

## Checklist

-   [ ] Jest tests


mariusandra:
Hey @Twixes this is ready for review! 

Basically the code above limits each 1) for/while/do-while loop, 2) promise.then, 3) await... inside the plugin VMs to no more than 30 seconds. It's far from the perfect solution to deter the most determined attackers, but should at least keep us reasonably safe from accidents. In addition to this, we wrap the `setupPlugin`, `processEvent` and `processEventBatch` functions with similar 30 second time limits.

The "30 second" timeout is not applied to the `runEveryMinute`, `..Hour`, `..Day` functions as a whole, yet any promise or for loop inside those functions still gets the 30 second limit. I'm not sure if this will cause problems or if this is fine for now.

Also, this does not protect against memory exploits (e.g. `new Buffer(Math.pow(2, 64))`), but those are harder to accidentally do.

mariusandra:
Had some trouble with tests in the end, but good to go. I added some extra metrics in the end as well.

Twixes (approved):
Postgres suite stalls after merging with master… but besides that this is 👌

mariusandra:
There's one flaky test where it [expects](https://github.com/PostHog/plugin-server/blob/79fdc0ec18c210b2c6ce63387c4c30295d5c2784/tests/postgres/worker.test.ts#L181) one number to be 32, but sometimes it's 33 or 34. So for now I'll just rerun the tests and hope they pass :).

PR #163, opened 2021-02-18
Kafka ingestion logging time fix

mariusandra:
## Changes

- Move the "status.info" higher up so that the "awaits" don't influence the time

## Checklist

-   [ ] Jest tests


mariusandra:
This would hopefully make the `ingestion` time here more consistent: 
image

PR #164, opened 2021-02-18
Prefilter Celery hooks task sending

### LINKED ISSUE #162: Get Celery out of the hot path
We still have Celery [posting a webhook task](https://github.com/PostHog/plugin-server/blob/3f1cb69cecce4fe356ceb41011eb5152ab30e56c/src/ingestion/process-event.ts#L478) every time an event is ingested.

The right solution is to move [webhooks into JS](https://github.com/PostHog/plugin-server/issues/145), but before that, we should just 1) disable the line that sends the celery task, 2a) either move the "is feature available" code to the plugin server that limits these tasks... or 2b) do this webhook processing in `capture.py` directly

In any case, this is a blocker for re-enabling ingestion for everyone.

Twixes:
## Changes

Similar to https://github.com/PostHog/posthog/pull/3368, this introduces better prefilter on sending the hook Celery task, to minimize Celery load.
Resolves #162.

### DIFF
--- a/src/ingestion/process-event.ts
+++ b/src/ingestion/process-event.ts
@@ -475,17 +475,34 @@ export class EventsProcessor {
             )
         }
 
-        this.celery.sendTask('ee.tasks.webhooks_ee.post_event_to_webhook_ee', [
-            {
-                event,
-                properties,
-                distinct_id: distinctId,
-                timestamp,
-                elements_list: elements,
-            },
-            team.id,
-            siteUrl,
-        ])
+        let sendHookTask = !!team.slack_incoming_webhook
+        if (!sendHookTask) {
+            try {
+                const hookQueryResult = await this.db.postgresQuery(
+                    `SELECT COUNT(*) FROM ee_hooks WHERE team_id = $1 AND event = 'action_performed' LIMIT 1`,
+                    [team.id]
+                )
+                sendHookTask = !!hookQueryResult.rows[0].count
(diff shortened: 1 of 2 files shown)


src/ingestion/process-event.ts
```diff
@@ -475,17 +475,34 @@ export class EventsProcessor {
...
+        let sendHookTask = !!team.slack_incoming_webhook
+        if (!sendHookTask) {
+            try {
+                const hookQueryResult = await this.db.postgresQuery(
+                    `SELECT COUNT(*) FROM ee_hooks WHERE team_id = $1 AND event = 'action_performed' LIMIT 1`,
+                    [team.id]
+                )
+                sendHookTask = !!hookQueryResult.rows[0].count
+            } catch (error) {
+                // In FOSS PostHog ee_hook does not exist. If the error is other than that, rethrow it
```
mariusandra, line 487: Could it make sense to just check for `KAFKA_ENABLED` here and if so, disable the entire thing? These failing postgres queries might cause too much log noise for some users perhaps?

Also, on EE, this query might be run hundreds of times each second and perhaps could be cached somehow? That said, this is likely good enough for now and the caching can move into another PR if/when needed.

Twixes: Good point about `KAFKA_ENABLED`.
As for caching, I refactored this to use Redis for 120 s caching per team.

PR #171, opened 2021-02-18
Remove statement_timeout and sequential ingestions

mariusandra:
## Changes

- Remove old sequential ingestion in kafka queue
- Remove `statement_timeout`

## Checklist

-   [ ] Jest tests


mariusandra:
This also removes `heartbeat` from the middle of a processed batch. It tends to `throw` if we're rebalancing resulting us to abort the batch, even though we had already run all the plugins on it. So instead we'll run it in the end and increase the session timeout to 60sec, which should be enough to process and ingest all the events.

PR #193, opened 2021-02-22
Don't "ALTER TABLE person UPDATE"

Twixes:
## Changes

`ALTER TABLE person UPDATE` is ran on every person-updating event and, due to the way ClickHouse works, that causes a drastic increase of load in terms of `person` parts. This contributed to the ClickHouse crash on 100% plugin server ingestion. Before rolling out plugin server ingestion, we need to solve person updates differently.
This PR simply eliminates the update query – not great for data integrity… but also the way Python ingestion is anyway…
Long term, we need a better solution.

### DIFF
--- a/src/db.ts
+++ b/src/db.ts
@@ -111,7 +111,15 @@ export class DB {
     public async fetchPersons(database: Database.ClickHouse): Promise<ClickHousePerson[]>
     public async fetchPersons(database: Database = Database.Postgres): Promise<Person[] | ClickHousePerson[]> {
         if (database === Database.ClickHouse) {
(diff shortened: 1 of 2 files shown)


mariusandra:
So this obviously breaks postgres parity... and is it really true no props are updated for persons in CH? 🤔

Twixes:
Please do crosscheck how it's done in the Python version, but that's how it appears to be from my inspection!

mariusandra:
Now we just `insert` when changing any properties, just like the python ingestion does. The `ReplacingMergeTree` will take care of eliminating the old rows eventually... and in queries we'll just select `max(_timestamp)`...

Twixes (commented):
This should work

PR #196, opened 2021-02-22
ESLint Promise Rules

mariusandra:
## Changes

- I added a few eslint rules to track down unhandled or otherwise misused promises.
- This possibly fixed a few flaky tests, as we were not really awaiting `closeServer` and a few other things in some places
- A few places got a `.finally(() => true)` to explicitly say we don't care about the returned promise

## Checklist

-   [ ] Jest tests

### DIFF
--- a/src/celery/client.ts
+++ b/src/celery/client.ts
@@ -108,10 +108,7 @@ export default class Client extends Base {
     public sendTask(taskName: string, args?: Array<any>, kwargs?: Record<string, any>, taskId?: string): void {
         taskId = taskId || v4()
         const message = this.createTaskMessage(taskId, taskName, args, kwargs)
-        this.sendTaskMessage(taskName, message)
-            .then(() => true)
-            .catch((error) => {
-                throw error
-            })
+        // run in the background
+        void this.sendTaskMessage(taskName, message)
     }
 }
--- a/src/celery/client.ts
+++ b/src/celery/client.ts
@@ -22,14 +22,14 @@ export default class Client extends Base {
         return this.taskProtocols[this.conf.TASK_PROTOCOL]
     }
 
-    public sendTaskMessage(taskName: string, message: TaskMessage): void {
+    public async sendTaskMessage(taskName: string, message: TaskMessage): Promise<void> {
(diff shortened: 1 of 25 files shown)


src/celery/client.ts
```diff
@@ -109,5 +109,9 @@ export default class Client extends Base {
         taskId = taskId || v4()
         const message = this.createTaskMessage(taskId, taskName, args, kwargs)
         this.sendTaskMessage(taskName, message)
+            .then(() => true)
+            .catch((error) => {
+                throw error
+            })
```
Twixes, line 115: This is a bit ugly, but I guess making `sendTask` async would require a lot of `await`s "up" the line?

src/ingestion/kafka-queue.ts
```diff
@@ -167,21 +167,21 @@ export class KafkaQueue implements Queue {
         return await startPromise
     }
 
-    async pause(): Promise<void> {
+    pause(): void {
```
Twixes, line 170: Why not keep this `async`, just not `await`ing in practice?

src/web/server.ts
```diff
@@ -17,6 +17,7 @@ export async function startFastifyInstance(pluginsServer: PluginsServer): Promis
     status.info('👾', 'Starting web server…')
     const fastifyInstance = buildFastifyInstance()
     try {
+        // eslint-disable-next-line
```
Twixes, line 20: Could this only disable the offending rule (I'm not sure which one it is)?

src/celery/client.ts
mariusandra: This should still run in the background, otherwise with redis being as flaky as it is, everything will come to a halt here if we await for it.

I found an easier way to ignore the returned promises though: `void`. Check the latest commits.

src/web/server.ts
mariusandra: Fixed it in another way

src/ingestion/kafka-queue.ts
mariusandra: sure, reverted pausing to being async

PR #198, opened 2021-02-22
Try to get redis key twice

mariusandra:
## Changes

- A very naïve approach at fixing this [Sentry Error](https://sentry.io/organizations/posthog/issues/2231006971/?project=5592816&query=is%3Aunresolved)

## Checklist

-   [ ] Jest tests


mariusandra:
Let's try this. It'll at least let us know if the connection can be salvaged or if the second redis query will also get stuck.

PR #199, opened 2021-02-23
Make plugin server ingestion the default

Twixes:
## Changes

Let's make PLUGIN_SERVER_INGESTION _the_ way.

The complimentary PR in main repo will need to also be adjusted for the new default.

**Warning: this has "bump major" which will result in 1.0.0 release.**

## Checklist


mariusandra:
This makes sense, yet it's a bit early :)

- I'd reserve the stable `1.0.0` version once we have deemed the entire server stable. That means running it on cloud for a few days with no issues.
- Those redis issues need to be fixed...
- The flag should be set in the main `posthog` repo at the same time.

Twixes:
Alright, agree with stability! Hopefully we can get it to a good state soon.
The flag would obvs be updated too, in the autoPR updating plugin server version.

PR #203, opened 2021-02-23
Proposal: Run scheduled tasks immediately at setup

yakkomajuri:
## PRs over issues

The only reason this is a PR is for context, not because this is a proper solution that should get merged. I haven't looked too deeply at how anything works.

## Proposal
The proposal here is essentially to run scheduled tasks right after setup. 

## Motivation
Here https://github.com/PostHog/github-release-tracking-plugin/issues/2 @mariusandra suggests that in order to give my users immediate feedback of my `runEveryDay` plugin working, I should use the cache to determine when was the last time the values were fetched and run the task more often.

This is certainly fine and quite easy to do, but it's not something I want to do as a plugin developer (I'm thinking from a community perspective here). 

For example, if I have a task called `runEveryDay`, why should I rather `runEveryHour` and use the cache to check if it's been a day? 

Plus, the same issue applies to `runEveryHour`. The only way to give immediate feedback to a user now is to use `runEveryMinute`. 

## Proposed Solution

Run the scheduled tasks once at setup. The PR is here for context but it doesn't mean that this is how things would necessarily work. Essentially I want immediate feedback of my plugin running and working with no extra work from my end. 

I, the plugin developer, am lazy. :open_mouth: 

## Pros

- Immediate feedback
- Faster debugging
- Simpler for plugin developers
- Suitable for most plugin types 
    - Especially those where the schedule is more of an arbitrary selection that a surgically precise decision e.g. Release Tracker: I want to check for releases frequently enough but not too frequently, so it's ok for me that it checks immediately, then the next time is let's say within 13h, and then it kicks into the 24h schedule. 
    
## Cons

- Will throw off the schedule for the first cycle
    - Can be harmful to plugins that need precision? Which types would these be?
- Potential overhead?
- ...?

## Further thoughts

Could we have this be a plugin setting? Developer decides if tasks should run at setup, default would probably be `true`.

### DIFF
--- a/src/services/schedule.ts
+++ b/src/services/schedule.ts
@@ -82,6 +82,12 @@ export async function startSchedule(
 
     server.pluginSchedule = await piscina.runTask({ task: 'getPluginSchedule' })
 
+    if (!stopped) {
+        runTasksDebounced(server!, piscina!, 'runEveryMinute')
+        runTasksDebounced(server!, piscina!, 'runEveryHour')
+        runTasksDebounced(server!, piscina!, 'runEveryDay')
+    }
+
     const runEveryMinuteJob = schedule.scheduleJob('* * * * *', () => {
         !stopped && weHaveTheLock && runTasksDebounced(server!, piscina!, 'runEveryMinute')
     })


Twixes (commented):
I think for production usage of plugins, this wouldn't really be acceptable. However to me, it sounds like the use case could be solved with a `runOnLoad` function, ran when the plugin is (re)loaded. WDYT?

mariusandra:
Well, basically that's what `setupPlugin` is for. Nobody prohibits you from writing code like:

```js
async function setupPlugin(meta) {
    await runEveryMinute(meta)
    // other init code
}

async function runEveryHoyr(meta) {
    // update annotations
}
```

or even having a common `updateAnnotations()` function that both places call. Could this abstraction work?

Twixes:
Oh right, totally forgot about `setupPlugin`. That exactly solves it in my understanding.

yakkomajuri:
Yeah that's true, `setupPlugin` does fill this purpose. I think I'll write this up at some point as a plugins best practice. Because even though instinctively I knew I could go about it this way, I guess it being called `setupPlugin` created a division of concerns in my mind that I didn't want to break. But really, it's actually quite fit for this.

PR #204, opened 2021-02-23
Redis connection pool

mariusandra:
## Changes

- With this change each Redis query needs to finish before a new one is allowed (per thread ofc). This might slow down plugins by a bit, but should greatly increase stability of the pipeline. 
- If this works fine, we could add a few more workers to the pool to increase throughput. This will increase the used redis connections, getting us in trouble with Heroku again...

## Checklist

-   [ ] Jest tests


src/db.ts
```diff
@@ -36,17 +38,28 @@ export class DB {
...
 
     constructor(postgres: Pool, redis: Redis.Redis, kafkaProducer?: Producer, clickhouse?: ClickHouse) {
         this.postgres = postgres
-        this.redis = redis
         this.kafkaProducer = kafkaProducer
         this.clickhouse = clickhouse
+        this.redisPool = createPool<Redis.Redis>(
+            {
+                create: () => Promise.resolve(redis),
+                destroy: () => Promise.resolve(undefined),
```
Twixes, line 55: Should this not stop/close/disconnect the Redis client?

src/db.ts
```diff
@@ -36,17 +38,28 @@ export class DB {
...
+            },
+            {
+                min: 1,
+                max: 1,
```
Twixes, line 59: With max 1 this doesn't really test pooling, how about 2?

src/db.ts
mariusandra: Not if using just one client and the pool is effectively a glorified lock.

src/db.ts
mariusandra: Not testing pooling, testing locking ;)

mariusandra:
Let's actually go for a real pool...

mariusandra:
Something's broken with celery now :/. Will investigate/fix later today.

mariusandra:
Ready for a look! I had to change the tests around quite a bit, such as removing `ioredis-mock`, etc...

Twixes (approved):
Looks good!

PR #205, opened 2021-02-23
Allow $set/$set_once on any event

yakkomajuri:
## Changes

For parity with https://github.com/PostHog/posthog/pull/3444, full context can be found there.

## Checklist

-   [x] Jest tests


yakkomajuri:
Something like this seems to be happening quite often with tests:

```
      {"level":"ERROR","timestamp":"2021-02-23T16:34:26.334Z","logger":"kafkajs","message":"[Connection] Response CreateTopics(key: 19, version: 3)","broker":"kafka:9092","clientId":"plugin-server-test-0177cfbd-eb82-0000-f7e8-944fe07a51c4","error":"Topic with this name already exists","correlationId":2,"size":486}
```

yakkomajuri:
Unfortunately tagging you again @mariusandra ...

mariusandra:
Looking good and I like this feature! 

In preparation for the maxmind plugin using this, I added a deep equal check into `updatePersonProperties` to not update the person if the updated properties are the same as the old ones. Otherwise we'll have one new person per event.

I also added another test to check that this also works if there's no user already in the system.

In case these tests pass and nobody has any objections towards this feature, I'd be happy to merge!

yakkomajuri:
We should probably merge https://github.com/PostHog/posthog/pull/3444 together with this - guess we have to add the deep equality check there

mariusandra:
I had to re-run tests 3 times to get them to work (some flakiness with clichkouse person updates and merges), but they're good now!

mariusandra (approved):

PR #211, opened 2021-02-24
Fix webhooks/zapier

mariusandra:
## Changes

Turns out we were missing a `{}`

## Checklist

-   [ ] Jest tests

### DIFF
--- a/src/ingestion/process-event.ts
+++ b/src/ingestion/process-event.ts
@@ -531,17 +531,21 @@ export class EventsProcessor {
 
         if (await this.shouldSendHooksTask(team)) {
             this.pluginsServer.statsd?.increment(`hooks.send_task`)
-            this.celery.sendTask('ee.tasks.webhooks_ee.post_event_to_webhook_ee', [
-                {
-                    event,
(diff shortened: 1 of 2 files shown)


src/ingestion/process-event.ts
```diff
@@ -531,17 +531,21 @@ export class EventsProcessor {
...
+                        event,
+                        properties,
+                        distinct_id: distinctId,
+                        timestamp,
+                        elements_chain: elementsChain,
+                    },
+                    team.id,
+                    siteUrl,
+                ],
+                {}
```
macobo, line 547: Q: kwargs is currently marked as optional - is it not?

mariusandra: hmm, indeed, the `sendTask` function marks them as optional.

However if they're `None`, python/celery fails here

image

I added a `|| {}` to `sendTask` to fix this contract.

mariusandra:
I'm going to merge this to get zapier back. If there's anything to fix, that can be added after :)

macobo (approved):

PR #212, opened 2021-02-24
IP could be explicitly ignored and set to null

mariusandra:
## Changes

- Companion to https://github.com/PostHog/posthog/pull/3464 and requires https://github.com/PostHog/plugin-scaffold/pull/10 (update needed here after merging that) 

## Checklist

-   [ ] Jest tests


Twixes (approved):
Makes sense to me as well

PR #213, opened 2021-02-25
Remove PLUGIN_SERVER_INGESTION

mariusandra:
## Changes

- Remove `PLUGIN_SERVER_INGESTION` env bool
- For OSS, if plugins are enabled, they will now also ingest and no longer send a task to celery
- For EE, only listen to the ingestion topic, no more the WAL and always ingest (all EE installations are ingesting via the plugin server, so nothing changes for anybody)
- Kind of closes #199

## Checklist

-   [ ] Jest tests

### DIFF
--- a/src/worker/queue.ts
+++ b/src/worker/queue.ts
@@ -8,6 +8,7 @@ import { IngestEventResponse } from '../ingestion/ingest-event'
 import { KafkaQueue } from '../ingestion/kafka-queue'
 import { status } from '../status'
 import { PluginsServer, Queue } from '../types'
+import { UUIDT } from '../utils'
 
 export type WorkerMethods = {
     processEvent: (event: PluginEvent) => Promise<PluginEvent | null>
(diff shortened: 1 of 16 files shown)


src/worker/queue.ts
```diff
@@ -66,26 +67,22 @@ function startQueueRedis(server: PluginsServer, piscina: Piscina | undefined, wo
...
         ) => {
-            const event = { distinct_id, ip, site_url, team_id, now, sent_at, ...data } as PluginEvent
+            const event = {
+                distinct_id,
+                ip,
+                site_url,
+                team_id,
+                now,
+                sent_at,
+                uuid: new UUIDT().toString(),
```
Twixes, line 77: Why `uuid` added here?

mariusandra: Because of this: https://github.com/PostHog/plugin-server/blob/22c49ba019cfb5b093e907ccb6883ef281cd0e3d/src/ingestion/process-event.ts#L59-L61

Even though we don't use it when there's no kafka/clickhouse, without it plugin server ingestion was broken on OSS :). I could add a `&& kafka` to the check, but that would require `Optional`-ing a lot of types and adding many more `if no uuid` checks into inner methods...

Twixes: Alright

Twixes (approved):

PR #215, opened 2021-02-25
Support distinct_id in posthog.capture

yakkomajuri:
## Changes

I'll need to write up some docs for this.

Supports passing a `distinct_id` in props to `posthog.capture` so that you can capture events for specific users on scheduled plugins.

## Checklist

-   [x] Jest tests

### DIFF
--- a/src/vm/extensions/posthog.ts
+++ b/src/vm/extensions/posthog.ts
@@ -63,10 +63,10 @@ export function createPosthog(server: PluginsServer, pluginConfig: PluginConfig)
     }
 
     return {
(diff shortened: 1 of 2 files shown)


yakkomajuri:
Flaky tests...

Also only added a test for passing the distinct_id explicitly because a previous test checks for plugin name being used as the distinct_id.

mariusandra:
Should we still keep `distinct_id` in the properties after this? Genuinely asking, not implying anything...

yakkomajuri:
@mariusandra adding here just for transparent history in GH. 

As per Slack, we already have distinct_id as an event prop anyway.

PR #222, opened 2021-02-26
TypeScript plugins natively

### LINKED ISSUE #221: Native TypeScript support
We [run all plugin JS via babel](https://github.com/PostHog/plugin-server/blob/b235de30637f916d25ea243ca3f71ec2f95ca926/src/vm/transforms/index.ts#L7) to add timeouts for loops and promises. In addition to our own babel plugins, we also have the `env` preset loaded.

We could easily add `@babel/preset-typescript` and thus make it possible to just write `index.ts` files instead of `index.js` files. This would mostly eliminate the need for `npm` plugins, as you no longer need to compile your plugins beforehand.

Twixes:
## Changes

Direct support for plugins written in TypeScript. No compilation by user needed. Resolves #221.

## Checklist

-   [x] Jest tests

### DIFF
--- a/tests/transforms.test.ts
+++ b/tests/transforms.test.ts
@@ -1,7 +1,7 @@
 import { createServer } from '../src/server'
 import { PluginsServer } from '../src/types'
 import { code } from '../src/utils'
-import { secureCode } from '../src/vm/transforms'
+import { transformCode } from '../src/vm/transforms'
 import { resetTestDatabase } from './helpers/sql'
 
 let server: PluginsServer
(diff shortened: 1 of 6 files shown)


tests/transforms.test.ts
```diff
@@ -114,4 +114,39 @@ describe('secureCode', () => {
             }
         `)
     })
+
+    it('secures block for loops with timeouts avoiding _LP collision', () => {
```
mariusandra, line 118: `it('supports typescript')`

mariusandra:
Nice! There's a failing test as well..

mariusandra:
In short, this is very cool! Can you double check that with this enabled, all of the plugins in the repository also install and launch fine without compilation errors? If so, then LGTM and can be merged!

PR #228, opened 2021-03-03
Add $device_type to initial properties

mariusandra:
## Changes

- Companion PR to https://github.com/PostHog/posthog-js/pull/198

## Checklist

-   [ ] Jest tests

### DIFF
--- a/src/ingestion/utils.ts
+++ b/src/ingestion/utils.ts
@@ -166,7 +166,15 @@ export function timeoutGuard(message: string, timeout = defaultConfig.TASK_TIMEO
 }
 
 const campaignParams = new Set(['utm_source', 'utm_medium', 'utm_campaign', 'utm_content', 'utm_term'])
-const initialParams = new Set(['$browser', '$browser_version', '$current_url', '$os', '$referring_domain', '$referrer'])
+const initialParams = new Set([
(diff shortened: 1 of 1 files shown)


yakkomajuri (approved):

src/ingestion/utils.ts
```diff
@@ -166,7 +166,15 @@ export function timeoutGuard(message: string, timeout = defaultConfig.TASK_TIMEO
 }
 
 const campaignParams = new Set(['utm_source', 'utm_medium', 'utm_campaign', 'utm_content', 'utm_term'])
-const initialParams = new Set(['$browser', '$browser_version', '$current_url', '$os', '$referring_domain', '$referrer'])
+const initialParams = new Set([
+    '$browser',
+    '$browser_version',
+    '$device_type',
```
Twixes, line 172: IMO we should add `$device` in here too then.

mariusandra: The problem is that `$device` is just not set when you're not on a mobile device. Thus we the var would actually mean `$first_used_mobile_device`... and cause issues with accurate statistics.

Twixes: Hm, right, quirky prop. OK then

PR #229, opened 2021-03-03
Stop flipping plugin order

mariusandra:
## Changes

- We were reversing the order of plugins on every run. Doh!

image

## Checklist

-   [ ] Jest tests


Twixes (commented):
🤔 
There's also a `for (const pluginConfig of pluginsToRun.reverse())` in `runPluginsOnBatch`, I guess that's the same situation?

mariusandra:
Yep, nice catch!

PR #233, opened 2021-03-09
Add support for FOSS webhook task

Twixes:
## Changes

Resolves the issue where webhooks don't for FOSS PostHog users who have opted into plugins.

## Checklist

-   [ ] Jest tests


mariusandra:
This doesn't seem to work. The `event.actions_set.all()` queryset is empty. Probably because of the async action mapping?

image

Issue #59, opened 2020-12-10
Lazy VMs / Improve reloads


mariusandra:
Currently when reloading, we:
- shut down celery
- wait 2 seconds
- shut down all workers
- restart all workers (they will reload all plugins)
- restart celery

Downtime of min 2sec, but more like 3-4sec. Even more if you have hundreds of plugins.

Ideally we should:
- only restart plugins that changed
- restart them live (keep the old one running while the new one restarts) and abort reload if there's an init error
- rolling restart between threads, not all at once

mariusandra:
Also: add test for reloads. When running in the `ingestion-save` branch, reloads killed the server for me. Not so in `master`.

mariusandra:
Speccing this out a bit.

### Why?

When the plugin server reloads, event ingestion stops for about half a minute while we basically restart the entire server. These reloads happen whenever anyone changes anything for any plugin in any team (installing, enabling, disabling, etc).

In Grafana it looks like this:

![image](https://user-images.githubusercontent.com/53387/109303242-d0b6a880-783a-11eb-8920-c1a155784511.png)

This will become a problem if we allow anyone on cloud to use plugins, as any change in any team will bring down the server for everyone for up to a minute.

### What?

When one team updates their plugins, event ingestion for all other teams should not be affected. Moreover, only the plugins that changed (added/deleted/updated) should be reloaded, instead of destroying and then restarting the entire worker pool.

### How?

We need to make the following changes:

1. Required: System to "broadcast" tasks to *all* piscina worker threads. Currently if you `piscina.runTask`, it'll choose one worker thread to run the task on. We need to tell all worker threads that they should reload the relevant VMs. This probably requires submitting a PR to the [piscina repository](https://github.com/piscinajs/piscina). Alternatively we could have an additional redis pubsub subscription (+1 connection per thread) inside each worker that listens to the reload event.

2. Required: System of "Lazy VMs". Right now we wait for all VMs to finish their setup before we start ingesting plugins. With this "Lazy VM" system, we would just do something like `vm.methods.processEvent = (event) => vm.launch().then(launchedVm => launchedVm.methods.processEvent)`. If the VM has been launched, it'll just run `processEvent` as normal. If it hasn't been launched, processing the event will just take a few moments longer. Work on this has already started in #220

[...]

PR #234, opened 2021-03-09
Speed up boot times

macobo:
Resolves https://github.com/PostHog/plugin-server/issues/59
Depends on https://github.com/PostHog/posthog/pull/3607

It:
1. Uses marius' piscina fork to broadcast reloading to all workers. Note I have not yet updated package.json accordingly.
2. VMs are now lazy, meaning you can ask for its methods/tasks before it has finished loading. If the VM fails to load, nulls/empty values are returned accordingly.
3. Diffing plugins works by looking at plugin+config updated_at times.
4. Schedules are loaded lazily by workers. The main thread polls workers for the schedule.

## Checklist

-   [ ] Jest tests


src/plugins/loadPlugin.ts
```diff
@@ -0,0 +1,110 @@
...
+                    `No "main" config key or "index.js" file found for plugin "${plugin.name}"`
+                )
+                return false
+            }
+
+            const jsPath = path.resolve(pluginPath, config['main'] || 'index.js')
+            const indexJs = fs.readFileSync(jsPath).toString()
+
+            const libPath = path.resolve(pluginPath, config['lib'] || 'lib.js')
+            const libJs = fs.existsSync(libPath) ? fs.readFileSync(libPath).toString() : ''
```
mariusandra, line 49: One of the things you could do (though it's conceptually separate and might be better for a different PR) is to remove support for `lib.js` while at it. Removing deprecated codepaths should probably fall under "getting out of beta".

src/plugins/setup.ts
```diff
@@ -0,0 +1,121 @@
...
+            prevConfig &&
+            pluginConfig.updated_at === prevConfig.updated_at &&
+            plugin?.updated_at == prevPlugin?.updated_at
+        ) {
+            pluginConfig.vm = prevConfig.vm
+        }
+
+        if (!pluginConfig.vm) {
+            await loadPlugin(server, pluginConfig)
+        }
```
mariusandra, line 27: Possible optimisation: push all the promises into an array and `await Promise.all(loadPluginPromises)`. This should speed up loading even more. 

One future danger: we might one day load thousands of plugins in parallel this way. For that there's a function `groupIntoBatches` in `utils.ts` that could split the array into chunks of 100, each of which can be awaited sequentially...

mariusandra (commented):
Overall looking very good! Two random flyby comments below

src/plugins/setup.ts
macobo: Good note. I _think_ no chunking is needed here since node.js only does things in event loop and we're not doing _too_ much async in there - all the file reads are sync and we're only hitting the db minimally.

macobo:
So after on-and-off fighting with tests/everything this is now green.

Things that slowed this down:
- ts-node and ts-node-dev issues
- worker.test.ts instability (got rewritten)
- running out of stack on the workers (increased limits)
- For some reason, promises started by the worker during init (but not completed) were _sometimes_ dropped completely, meaning that VMs could get into a state where they never finished initializing. This was worked around by adding a new message that forces VMs to wait on the promises, which did the trick.

src/server.ts
```diff
@@ -251,14 +250,9 @@ export async function startPluginsServer(
...
-                await stopSchedule?.()
-                if (piscina) {
-                    await stopPiscina(piscina)
-                }
-                piscina = makePiscina(serverConfig!)
-                queue = await startQueue(server!, piscina)
-                stopSchedule = await startSchedule(server!, piscina)
+
+                await piscina?.broadcastTask({ task: 'reloadPlugins' })
+                await scheduleControl?.reloadSchedule()
```
mariusandra, line 255: I wonder what would happen if this gets called in the middle or beginning of a Kafka batch? `pluginConfig.vm` will be undefined for a short while on some plugins, possibly resulting in some crashes? Some logic to reload only after the batch has finished... and/or to make sure there's always some object on `vm` might make sense?

src/services/schedule.ts
```diff
@@ -108,13 +108,43 @@ export async function startSchedule(
...
 
-    return stopSchedule
+    const reloadSchedule = async () => {
+        lockTimeout && clearTimeout(lockTimeout)
+        weHaveTheLock = false
+        await lock?.unlock().catch(Sentry.captureException)
+
+        server.pluginSchedule = await loadPluginSchedule(piscina)
+
+        lockTimeout = setTimeout(tryToGetTheLock, 0)
```
mariusandra, line 118: Is this redlock unlocking and locking intentional? This will pause all scheduled tasks and potentially cause some to be missed. Very likely it will also be a different plugin server instance that picks up the lock and resumes the work.

src/vm/lazy.ts
```diff
@@ -0,0 +1,31 @@
...
+    logInfo = ''
+): LazyPluginVM {
+    const promise = createPluginConfigVM(server, pluginConfig, indexJs)
+        .then((vm) => {
+            status.info('🔌', `Loaded ${logInfo}`)
+            void clearError(server, pluginConfig)
+            return vm
+        })
+        .catch((error) => {
+            console.warn(`⚠️ Failed to load ${logInfo}`)
```
mariusandra, line 19: Log via `status`?

mariusandra (commented):
Looks good! I have some comments/concerns inline. I just read through the code, didn't run it yet and will start poking at it now. Submitting what I have until now.

mariusandra:
I casually ran the code and tried reloading/editing/playing around it seems to work fine. Some last observations:

1. The schedule logs a bit more than I'd expect:
image

2. Should we do something against parallel reloads? E.g. if 3 reload events come in 10ms after one another.

src/services/schedule.ts
macobo: So there's a problem in general where having the plugin server be down during HOUR:00 will make it so no queued tasks get run.

I think there needs to be a long-term solution for that but until then I think we can change this to just await until the plugins have (re)loaded

src/vm/lazy.ts
macobo: `status` does not have a warn method, copied the approach from previous code. Suggestions?

macobo:
@Twixes This should now be reviewable.

I haven't changed the logging logic - curious on your thoughts on what proper logging would look like here?

Twixes (changes requested):

package.json
```diff
@@ -64,7 +64,7 @@
         "node-fetch": "^2.6.1",
         "node-schedule": "^1.3.2",
         "pg": "^8.4.2",
-        "piscina": "^2.1.0",
+        "piscina": "https://github.com/PostHog/piscina.git#broadcast-release",
```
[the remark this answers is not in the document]
macobo: Sure. Mind if I handle this separately from this PR?

src/vm/lazy.ts
```diff
@@ -0,0 +1,37 @@
...
+                resolve(null)
+            }
+        }
+        returnValue.failInitialization = () => {
+            resolve(null)
+        }
+    })
+
+    return returnValue as LazyPluginVM
+}
```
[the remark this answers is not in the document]
macobo: Yes, but I don't think it would be any better code-wise. The problematic bit is that the only way to get access to `resolve`/`reject` is by using the Promise constructor makes this a bit messy, but it should still be relatively straight-forward.

src/server.ts
macobo: You mean marius' comment? No.

This line? Absolutely needed. :)

src/vm/transforms/index.ts
```diff
@@ -4,7 +4,13 @@ import { PluginsServer } from '../../types'
 import { loopTimeout } from './loop-timeout'
 import { promiseTimeout } from './promise-timeout'
 
+const memoize: Record<string, string> = {}
+
 export function transformCode(rawCode: string, server: PluginsServer): string {
+    if (process.env.NODE_ENV === 'test' && memoize[rawCode]) {
+        return memoize[rawCode]
+    }
```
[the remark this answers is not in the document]
macobo: I was having difficulty getting the tests run in a stable and fast way. This is test-only code that speeds up repeated compilations of the code (e.g. when running under jest.watch).

As for why not in prod - it introduces its own failure conditions (e.g. OOM, memory leaking etc) which is why I'd keep this test-only.

src/plugins/setup.ts
```diff
@@ -0,0 +1,125 @@
...
+            teamConfigs = []
+            pluginConfigsPerTeam.set(row.team_id, teamConfigs)
+        }
+        teamConfigs.push(pluginConfig)
+    }
+
+    return { plugins, pluginConfigs, pluginConfigsPerTeam }
+}
+
+export async function loadSchedule(server: PluginsServer): Promise<void> {
```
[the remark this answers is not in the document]
macobo: Not really since all callsites (e.g. worker) would then need to do `server.pluginSchedule = foobar` - this is a key piece in how the reloading logic works and I'd rather not leak this outside of this file.

src/worker/worker.ts
```diff
@@ -47,6 +48,12 @@ export async function createWorker(config: PluginsServerConfig, threadId: number
             const { pluginConfigId } = args
             response = cloneObject(await runPluginTask(server, task, pluginConfigId))
         }
+        if (task === 'reloadPlugins') {
+            await setupPlugins(server)
+        }
+        if (task === 'reloadSchedule') {
+            await loadSchedule(server)
+        }
```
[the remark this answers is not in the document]
macobo: I have a WIP branch which adds one more - will change it there!

src/server.ts
Twixes: Yep, meant the comment!

src/vm/lazy.ts
Twixes: Added `status.warn` method :)

src/vm/transforms/index.ts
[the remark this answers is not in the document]
Twixes: Makes sense, added a comment to the code

Twixes (approved):

PR #244, opened 2021-03-12
Improve cache JSON opts

Twixes:
## Changes

Exposes `jsonStringify`/`jsonParse` options in the cache extension.


src/vm/extensions/cache.ts
```diff
@@ -5,11 +5,11 @@ import { PluginsServer } from '../../types'
...
     const getKey = (key: string) => `@plugin/${pluginId}/${typeof teamId === 'undefined' ? '@all' : teamId}/${key}`
     return {
-        set: async function (key: string, value: unknown, ttlSeconds?: number): Promise<void> {
-            return await server.db.redisSet(getKey(key), value, ttlSeconds)
+        set: async function (key: string, value: unknown, ttlSeconds?: number, jsonStringify = true): Promise<void> {
+            return await server.db.redisSet(getKey(key), value, ttlSeconds, jsonStringify)
         },
-        get: async function (key: string, defaultValue: unknown): Promise<unknown> {
-            return await server.db.redisGet(getKey(key), defaultValue)
+        get: async function (key: string, defaultValue: unknown, jsonParse = true): Promise<unknown> {
```
macobo, line 11: Perhaps this should be an options object? This API needs to be quite stable and Nth argument booleans get messy.

Twixes: Fair point, I'll make the change everywhere, as a lone boolean param is indeed rarely a clear or extensible solution.

Twixes: Done

macobo (approved):

PR #245, opened 2021-03-12
Set explicit limits to postgres connection pool:

macobo:
https://node-postgres.com/api/pool#poolquery

- `idleTimeoutMillis` is set to 500ms to make sure we return
connections/don't block other apps.
- `max` is set to 10 (per worker) to make sure we don't leak connections
indefinetly.

## Checklist

-   [ ] Jest tests


Twixes (approved):
10 max seems a bit low to me considering the throughput, but YOLO

macobo:
This is 10 per worker/thread. Given that we process 10 tasks in parallel per worker, this seems appropriate.

Note there's pgbouncer-side things that could come into play here as well: https://posthog.slack.com/archives/C01F6QWTVB5/p1615541141009400?thread_ts=1615538460.001100&cid=C01F6QWTVB5

PR #249, opened 2021-03-16
Batch kafka producer messages

macobo:
Updated kafka producer to send messages in batches. Updated worker.ts tests as well since was having trouble with some of the instability.

This should speed up plugin server significantly. Future optimizations would include making sure we only send kafka messages from main thread to get more out of batching, though I am not sure I can follow the code well here to make sure we don't already.

@Twixes did not update worker.ts to use switch - some if conditions are not using plain equals there but more complicated conditions.

## Checklist

-   [x] Jest tests

### DIFF
--- a/src/db.ts
+++ b/src/db.ts
@@ -19,6 +19,7 @@ import {
     Event,
     Person,
     PersonDistinctId,
+    PluginsServerConfig,
     PostgresSessionRecordingEvent,
     RawOrganization,
     RawPerson,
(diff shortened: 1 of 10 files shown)


src/db.ts
```diff
@@ -121,17 +136,44 @@ export class DB {
     }
 
     // Kafka
+    public async queueKafkaMessage(kafkaMessage: ProducerRecord): Promise<void> {
+        this.kafkaMessageQueue.push(kafkaMessage)
+
+        const timeSinceLastFlush = new Date().getTime() - this.lastFlushTime.getTime()
+        if (timeSinceLastFlush > this.flushFrequencyMs || this.kafkaMessageQueue.length >= this.maxQueueSize) {
+            await this.flushKafkaMessages()
+        }
```
Twixes, line 145: Why both this and the setInterval?

macobo: Good question.

I added this to be the main wait condition - this way it's nicely in the path of ingestion and we do block if the queue gets too long.

the interval is the backup condition for making sure we still flush messages even if no messages come in for a longer time.

Twixes (approved):

PR #250, opened 2021-03-17
 Use DB#sendKafkaMessage in posthog extension

Twixes:
## Changes

Was reading https://kafka.js.org/docs/producing and recalled `acks`. This setting's default means that all brokers must receive the message before it's accepted, but we may not need that. Instead let's try only requiring ack from the leader broker. This should help with produce latency.


Twixes:
Actually let's not sacrifice durability, dropping the `acks` change.

PR #251, opened 2021-03-17
Send an (estimated) latency metric to statsd

macobo:
This should allow us to graph min(metric) over mean(instances).

Couple of things make this a bit tricky/hard to be 100% confident in:
1. Different consumers will have different latencies
2. sentAt is not always set
3. ts can be in the future
4. Clock drift schenanigans

However, since we don't need a _accurate_ estimate (to the second) but
rather e.g. whether we're OK or 2 hours behind this should work.

## Checklist

-   [ ] Jest tests

### DIFF
--- a/src/ingestion/process-event.ts
+++ b/src/ingestion/process-event.ts
@@ -75,6 +75,12 @@ export class EventsProcessor {
         const personUuid = new UUIDT().toString()
 
         const ts = this.handleTimestamp(data, now, sentAt)
+        // Send latency estimate to statsd, but ignore events set in the future.
+        const latencyEstimate = (sentAt || ts).toMillis() - singleSaveTimer.getMilliseconds()
+        if (latencyEstimate > 0) {
+            this.pluginsServer.statsd?.timing('process_event.latency', latencyEstimate)
+        }
+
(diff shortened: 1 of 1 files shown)


src/ingestion/process-event.ts
```diff
@@ -75,6 +75,12 @@ export class EventsProcessor {
         const personUuid = new UUIDT().toString()
 
         const ts = this.handleTimestamp(data, now, sentAt)
+        // Send latency estimate to statsd, but ignore events set in the future.
+        const latencyEstimate = (sentAt || ts).toMillis() - singleSaveTimer.getMilliseconds()
```
Twixes, line 79: AFAIK (I may be wrong) sent_at is provided by client libraries, so I'm not sure if this is v. useful?

fuziontech: I agree, this could get pretty noisy if a user is backfilling historical data. It would make us look way behind when in fact it's just history being submitted after the fact.

Might be worth setting a TS when we start processing a batch or events as they come in and check as we are producing what the diff is between now and the batch start TS

macobo:
Closing for now - it doesn't solve the right problem.

PR #253, opened 2021-03-17
Access control adjustments

Twixes:
## Changes

Adjustments for https://github.com/PostHog/posthog/pull/3486.

## Checklist

-   [x] Jest tests


src/sql.ts
```diff
@@ -1,35 +1,36 @@
 import { Plugin, PluginAttachmentDB, PluginConfig, PluginConfigId, PluginError, PluginsServer } from './types'
 
+function pluginConfigsInForceQuery(specificField?: string): string {
```
macobo, line 3: Suggestion: replace `string` with `'id' | 'plugin_id'`. This makes the following string substitution a lot less scary from an injection standpoint.

src/utils.ts
```diff
@@ -388,6 +388,14 @@ export async function createRedis(serverConfig: PluginsServerConfig): Promise<Re
     return redis
 }
 
+export function pluginDigest(plugin: Plugin): string {
```
macobo, line 391: This seems dead

src/sql.ts
Twixes: Good idea, I used `keyof PluginConfig` for a bit of flexibility

src/utils.ts
Twixes: Merge conflict resolution gone wrong, fixed

PR #255, opened 2021-03-17
Telegraf setting experiment

macobo:
This should allow tagging metrics

## Changes

_Please describe._

## Checklist

-   [ ] Jest tests


fuziontech (approved):
sounds like a good change since we are using telegraf!

PR #258, opened 2021-03-18
Decrease max batch sizes

macobo:
- Remove unneeded code
    
    Seems like kafkajs already does this under the hood: https://github.com/tulios/kafkajs/blob/d13acd58f693bef5f2460ade036757d6314e4710/src/producer/messageProducer.js#L83-L93

- Decrease max batch size(s)
    
    We're seeing errors like `The request included a message larger than the
    max message size the server will accept` due to adding batching.
    
    This is an attempt to see whether reducing max batch sizes will fix
    things.

## Checklist

-   [ ] Jest tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ There's a multitude of settings you can use to control the plugin server. Use th
 | KAFKA_CLIENT_CERT_B64         | Kafka certificate in Base64                                | `null`                                |
 | KAFKA_CLIENT_CERT_KEY_B64     | Kafka certificate key in Base64                            | `null`                                |
 | KAFKA_TRUSTED_CERT_B64        | Kafka trusted CA in Base64                                 | `null`                                |
-| KAFKA_PRODUCER_MAX_QUEUE_SIZE | Kafka producer queue max size before flushing              | `1000`                                |
+| KAFKA_PRODUCER_MAX_QUEUE_SIZE | Kafka producer queue max size before flushing              | `20`                                  |
 | KAFKA_FLUSH_FREQUENCY_MS      | Kafka producer queue max duration before flushing          | `500`                                 |
 | DISABLE_WEB                   | whether to disable web server                              | `true`                                |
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -50,7 +50,7 @@ There's a multitude of settings you can use to control the plugin server. Use th
 | KAFKA_CLIENT_CERT_B64         | Kafka certificate in Base64                                | `null`                                |
 | KAFKA_CLIENT_CERT_KEY_B64     | Kafka certificate key in Base64                            | `null`                                |
 | KAFKA_TRUSTED_CERT_B64        | Kafka trusted CA in Base64                                 | `null`                                |
-| KAFKA_PRODUCER_MAX_QUEUE_SIZE | Kafka producer queue max size before flushing              | `1000`                                |
+| KAFKA_PRODUCER_MAX_QUEUE_SIZE | Kafka producer queue max size before flushing              | `20`                                  |
```
Twixes, line 53: 20 seems like a pretty small number. 🤔 Certainly relative to 1000. But I don't know what the actual sizes tend to be in production. Could we get some metric reporting into DB#flushKafkaMessages?

macobo: Good idea. 

My plan is to experiment a bit here - hard to judge w/o knowing more about the behavior in production. We might even need a different tactic all-together, e.g. measuring the size of each message individually or something like that.

20 is the lowest we can go - if the error goes away after that we'll have more information.

As such, I'll add the queue size metrics once I have some measurements from this. :)

Twixes: I believe we can't increase max request size, as we're managed by Heroku.
Why not sprinkle in some StatsD now?

macobo: > I believe we can't increase max request size, as we're managed by Heroku.

Yes, but not for ever. :)

> Why not sprinkle in some StatsD now?

Keep PR lean, don't want to wait too many hours to get to see the result of this.

PR #259, opened 2021-03-18
Use structured logging over string interpolation

macobo:
Motivation: Our sentry setup is a mess due to errors not getting
collapsed. They're not getting collapsed because our error strings are
unique every time.

The industry best practice for logging and error messages is to _avoid_
interpolating in strings, instead log fully as JSON or adding JSON at
the end. This allows:
- Sentry to collapse errors correctly
- Both sentry and logging tools to allow json search on variables within
the strings.

Read more here:
- https://stackify.com/what-is-structured-logging-and-why-developers-need-it/

## Checklist

-   [ ] Jest tests

### DIFF
--- a/src/db.ts
+++ b/src/db.ts
@@ -86,7 +86,7 @@ export class DB {
         tag?: string
     ): Promise<QueryResult<R>> {
         return this.instrumentQuery('query.postgres', tag, async () => {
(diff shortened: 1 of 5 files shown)


Twixes (approved):
Very well

Issue #261, opened 2021-03-19
Proposal: Refactor the codebase to reflect internal service boundaries


macobo:
As I've worked on the plugin-server codebase the hardest thing for me has been keeping a clear view of how the program functions for two reasons:
1. It's hard to tell if a piece of code is called within the main thread, workers or both 
2. It's hard to map out the message flow due to same or similar-meaning verbs being used all over the place: For example, what's the difference between ingesting and processing events, saveEvent/ingestEvent naming getting changed, processEventBatch being both a worker task name as well as a plugin method name etc.

Our codebase is currently split up up by _functionality_ - e.g. one folder for `ingestion`-related logic, another for `worker` setup, another for `plugin` setup and so on. This makes sense but makes it hard to follow what code is worker-related, what is not.

Proposal for improving (1): 
1. Let's create two subdirectories under `src/` - `src/main` and `src/workers` and `src/shared` - along the internal service boundary lines. So code in `src/workers` contains everything needed to run the worker threads, `src/main` for the main thread and `src/shared` for any code shared between the two (e.g. implementation of createServer, DB etc). 
2. Both `src/worker` and `src/main` import from `src/shared` but from no other directory.
3. Within each subdir, we can further divide the code up as we already do - e.g. plugins, ingestion, etc.
4. Figure out a way to use composition rather than putting most not all queries into one `db.ts` file.

PR #262, opened 2021-03-19
Proposal: Reorganize the codebase

### LINKED ISSUE #47: Refactor `src/*.ts`
There's a lot of stuff in the root of the `src` folder. This should be cleaned up. Kind of pending behind the currently open PRs (#39, #34, #25) and better to do after.

macobo:
## Overview

This PR moves code along the lines proposed in https://github.com/PostHog/plugin-server/issues/261. I find it makes it much easier to follow the code flow - what code is being executed in workers, what in main threads and what everywhere.

No real behavioral changes here, just moving files. Some minor changes:
- Split up server.ts into `src/main/pluginsServer` and `src/shared/server`
- Renamed celery/worker to `src/main/ingestion/celery-queue-worker`. This avoids confusion with `worker` threads.

The eventual goal would be to make `src/shared` as tiny as possible and move everything else under specific subfolders.

Resolves #47.

## Commits

- Move schedule.ts under src/main
- Move queue.ts under src/main
- Move kafka-queue logic under src/main
- Move ingest-event file
- Move web/server under main.ts
- Move plugins code under src/worker
- Move vm code under src/worker
- Move celery worker under src/main
- move process-event under src/worker
- Move src/celery under src/shared
- Move remaining src/ingestion under src/shared
- move utils under src/shared
- Move some top-level code under shared
- Move main pluginsserver code under src/main
- Fix typeerror in tests
- Move createServer under src/shared

## Checklist

-   [ ] Jest tests


Twixes (commented):
I broadly agree. My only two notes are:

1. I think it's a bit awkward that `src/celery/worker.ts` got pulled out of the `celery` directory into `src/main/ingestion/celery-queue-worker.ts`, especially considering that the `celery` directory used to be a whole separate package. The analogy to the Kafka queue is good though.
2. Most of `src/server.ts` ended up in `src/main/pluginsServer.ts`, but not all of it. I think we should keep that file intact instead, as basically it's all `PluginsServer`.

However in any case I'd wait for @mariusandra's input on this too before merging.

macobo:
Thanks for the feedback. Both of these were motivated by trying to keep `shared/` as small as possible.

1. For server.ts I feel strongly. Basically half of the code there only executed/was available in the main thread and is the defacto start point for the whole control flow. The other half is under shared because that god object is used everywhere. I think the issue here is mostly around naming - is `Server` the right name for the shared state?

1. For worker.ts I can see it going both ways, I originally thought it was a pure alternative for `kafka-queue.ts`, but these two behave differently. It just feels a bit awkward having it under `shared/` given that it's not used by worker. If it does become a library it will still be easy to extract the code as needed.

mariusandra:
Regarding these discussed points:

1. I guess it's fine to split up the celery code info folder. I moved the lib into our codebase precisely so we could be flexible with this. (plus for other optimisations...)

2. Totally fine with splitting the "server" and "god" objects from "server.ts". The god object could even be called something like "services.ts" or something, as it basically just hosts connections to all the things we connect to.

mariusandra (approved):
Approved. Some merge conflicts though.

PR #266, opened 2021-03-22
Release a patch version

macobo:
Forgot to bump patch with last PR, doing it now

## Checklist

-   [ ] Jest tests


Twixes:
Probably simpler to just edit package.json in main and increment the version there, as that's all the label does. :)

macobo:
main branch is locked though last time I checked. /shrug

Twixes:
Ahh, right, sorry, speaking from position of admin privilege, in this case label it is

Issue #257, opened 2021-03-17
Long posthog_team table rows slow down ingestion


macobo:
The slowest query in plugin server by far is `SELECT * FROM posthog_team WHERE id = $1`. We spend more time on this query than on all other postgres queries combined (!) in plugin server.

![image](https://user-images.githubusercontent.com/148820/111531769-a424eb00-876d-11eb-963b-1c4d57ebc9ca.png)

It's not slow due to the query, but rather deserialization. Some columns are just that wide, for example we have a team whose `event_names_with_usage` column is 500k characters long.

We should fix this.

macobo:
Did some more measurements:

```
d25vf5hht9th9v> SELECT length(event_names_with_usage::text), length(event_properties::text), length(even
 t_properties_numerical::text), length(event_names::text), length(event_properties_with_usage::text) FRO
 M posthog_team order by length(event_names_with_usage::text) DESC LIMIT 10;
+----------+----------+----------+----------+----------+
| length   | length   | length   | length   | length   |
|----------+----------+----------+----------+----------|
| 574206   | 560      | 105      | 338915   | 2237     |
| 75717    | 58       | 2        | 23186    | 311      |
| 43815    | 439      | 2        | 26553    | 1639     |
| 22731    | 1099     | 325      | 16035    | 4236     |
| 17022    | 1372     | 110      | 6174     | 6422     |
| 16833    | 3862     | 1723     | 7919     | 14233    |
| 15590    | 5738     | 287      | 6412     | 11302    |
| 13981    | 31       | 2        | 7126     | 160      |
| 11744    | 1366     | 510      | 4656     | 5307     |
| 10773    | 577      | 145      | 5135     | 2176     |
+----------+----------+----------+----------+----------+
SELECT 10
Time: 0.227s
```

The problematic (large) columns are `event_names_with_usage` and `event_names`, weighing ~1MB in total for the largest user. This is downloaded for every event, which results in significant load.

These columns are used in the FE to display information/stats about events.

Plugin server has the logic to populate these columns, [making them currently needed to be read](https://github.com/PostHog/plugin-server/blob/a0179ba/src/ingestion/process-event.ts#L415-L454). The corresponding logic in main app can be found [here](https://github.com/PostHog/posthog/blob/84d853748d45f9b4cc8b593f17c9fe228a58dee0/posthog/tasks/process_event.py#L78-L101).

[...]

PR #271, opened 2021-03-22
Add caching for Team fetching

macobo:
After this we cache Team objects in memory. Solves https://github.com/PostHog/plugin-server/issues/257 and speeds up ingestion by reducing the amount of time spent on the top 2 queries.

Future work: 
- Make the schema sane. We're using an array-inside-jsonb which cannot be updated idempotently unless we add pl/pgsql functions to help us. Note that this updating is also (already) racy as well.
- Maybe move this out to main thread or use bloom filters if memory becomes a problem.

## Checklist

-   [ ] Jest tests

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -238,7 +238,6 @@ export interface Team {
     session_recording_opt_in: boolean
     plugins_opt_in: boolean
     ingested_event: boolean
-    __fetch_event_uuid?: string
 }
 
 /** Usable Element model. */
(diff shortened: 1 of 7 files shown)


mariusandra (approved):
Looks good and I think the `__fetch_event_uuid` cache busting system makes sense for now. We should definitely switch to a saner schema here though.

macobo:
Agreed. I'd like to drop the old ingestion code first though. :)

Issue #263, opened 2021-03-19
Kafka messages causing `message too large error`


macobo:
After implementing (kafka producer message batching)[https://github.com/PostHog/plugin-server/pull/249] which was needed to solve https://github.com/PostHog/plugin-server/issues/248, we started seeing the following issue pop up in sentry:

`The request included a message larger than the max message size the server will accept` [Sentry error](https://sentry.io/organizations/posthog/issues/2279241522/?project=5592816&query=is%3Aunassigned+is%3Aunresolved)

The max message size is configured in kafka configuration, probably [here](https://docs.confluent.io/platform/current/installation/configuration/broker-configs.html#brokerconfigs_message.max.bytes).

What we know:

- We can't (yet) change these settings because our kafka instance is hosted on Heroku. cc @fuziontech who will do a migration soon.
- Changing the max batch size from 1000 to 20 messages didn't affect the error.
- [librdkafka](https://github.com/edenhill/librdkafka/blob/master/INTRODUCTION.md#performance) already implements this batching in a way that would avoid running into this error, however we're using a different library.
- Some session recording events might be close to max size, causing us to run into this error

PR #275, opened 2021-03-23
Kafka message batching - set batch size limits

macobo:
- Refactor instrumenting queries
- Check message sizes for flushing

This implements max batch size logic, solving https://github.com/PostHog/plugin-server/issues/263.

The implementation is sort of lazy - we could read the max size from the broker but computing a _very_ accurate message size limit felt outside the scope a bit.

## Checklist

-   [x] Jest tests

### DIFF
--- a/src/shared/kafka-producer-wrapper.ts
+++ b/src/shared/kafka-producer-wrapper.ts
@@ -1,0 +1,90 @@
+import { StatsD } from 'hot-shots'
+import { Producer, ProducerRecord } from 'kafkajs'
+
+import { PluginsServerConfig } from '../types'
+import { timeoutGuard } from './ingestion/utils'
+import { instrumentQuery } from './metrics'
+
+export class KafkaProducerWrapper {
+    /** Kafka producer used for syncing Postgres and ClickHouse person data. */
+    producer: Producer
+    /** StatsD instance used to do instrumentation */
+    statsd: StatsD | undefined
+
+    lastFlushTime: number
+    currentBatch: Array<ProducerRecord>
+    currentBatchSize: number
+
+    flushFrequencyMs: number
+    maxQueueSize: number
+    maxBatchSize: number
+
+    flushInterval: NodeJS.Timeout
+
+    constructor(producer: Producer, statsd: StatsD | undefined, serverConfig: PluginsServerConfig) {
+        this.producer = producer
+        this.statsd = statsd
+
+        this.lastFlushTime = Date.now()
+        this.currentBatch = []
+        this.currentBatchSize = 0
+
+        this.flushFrequencyMs = serverConfig.KAFKA_FLUSH_FREQUENCY_MS
+        this.maxQueueSize = serverConfig.KAFKA_PRODUCER_MAX_QUEUE_SIZE
+        this.maxBatchSize = serverConfig.KAFKA_MAX_MESSAGE_BATCH_SIZE
+
+        this.flushInterval = setInterval(() => this.flush(), this.flushFrequencyMs)
+    }
+
+    async queueMessage(kafkaMessage: ProducerRecord): Promise<void> {
+        const messageSize = this.getMessageSize(kafkaMessage)
+
+        if (this.currentBatchSize + messageSize > this.maxBatchSize) {
+            await this.flush(kafkaMessage)
+        } else {
+            this.currentBatch.push(kafkaMessage)
+            this.currentBatchSize += messageSize
+
+            const timeSinceLastFlush = Date.now() - this.lastFlushTime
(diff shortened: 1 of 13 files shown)


src/shared/kafka-producer-wrapper.ts
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    private getMessageSize(kafkaMessage: ProducerRecord): number {
+        return kafkaMessage.messages
+            .map((message) => {
+                if (message.value === null) {
+                    return 4
+                }
+                if (!Buffer.isBuffer(message.value)) {
+                    message.value = Buffer.from(message.value)
```
mariusandra, line 84: This modifies the `value` in the message, converting it to a buffer if it's not. This feels unexpected for a `getMessageSize` function...

src/shared/kafka-producer-wrapper.ts
```diff
@@ -0,0 +1,90 @@
...
+
+        this.lastFlushTime = Date.now()
+        this.currentBatch = []
+        this.currentBatchSize = 0
+
+        this.flushFrequencyMs = serverConfig.KAFKA_FLUSH_FREQUENCY_MS
+        this.maxQueueSize = serverConfig.KAFKA_PRODUCER_MAX_QUEUE_SIZE
+        this.maxBatchSize = serverConfig.KAFKA_MAX_MESSAGE_BATCH_SIZE
+
+        this.flushInterval = setInterval(() => this.flush(), this.flushFrequencyMs)
```
mariusandra, line 36: I guess this interval is good enough, though I think a "fully correct" solution would use a timeout here. It's too easy to get to a situation where the "flush happens 50ms after queueing simply due to the interval boundaries. Though I guess this is a nit and changing it to a different debouncing strategy won't really affect real world performance.

src/shared/kafka-producer-wrapper.ts
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    async queueMessage(kafkaMessage: ProducerRecord): Promise<void> {
+        const messageSize = this.getMessageSize(kafkaMessage)
+
+        if (this.currentBatchSize + messageSize > this.maxBatchSize) {
+            await this.flush(kafkaMessage)
+        } else {
```
mariusandra, line 44: I was a bit confused by these lines at first. It seems that the `kafkaMessage` is appended to the flushed queue... whereas in reality it's appended to a new and empty queue.

I think the code would read better if the flush takes no arguments and the `else` part of this will run always?

mariusandra (commented):
Overall looks good, though I didn't run it yet (school run ahead... will do right after). Found some things to nit about though :).

src/shared/kafka-producer-wrapper.ts
macobo: It's hard to implement it that way without introducing any races though. 

There were three cases we want to flush:
1. Message size + current batch exceeds max size
2. Too much time passed
3. Too many messages queued.

Cases 2-3 we handle flushing by adding the message to queue, resetting the queue before sending it on the way.

1 is slightly tricky - we want to flush messages that were queued _before_ this message and set the queue to be this single message. A naive solution would do something like 

```ts
await this.flush()
this.queue.push(kafkaMessage)
```

But this introduces a new race condition that could result in too large messages getting sent.

So the parameter was added to combat it. It's slightly smelly but has minimal leakage and is tested so  ¯\_(ツ)_/¯

src/shared/kafka-producer-wrapper.ts
macobo: The interval was added here (in a previous PR) as an additional protection against data loss. Performance-wise this does not matter and I'd keep our own custom kafka producer impl as lean as possible since we're literally reinventing a wheel in a way I hope to delete some day.

src/shared/kafka-producer-wrapper.ts
macobo: Agreed. Suggestions re naming? Note that kafka-js does the exact same thing under the hood IIRC.

src/shared/kafka-producer-wrapper.ts
mariusandra: Ok. Probably would be good to add some deodrant comment to explain this

macobo: Done

src/shared/kafka-producer-wrapper.ts
mariusandra: Can't we just do it in two passes? Check and convert when adding to the queue and even `+=` the length onto something?

Though I'm willing to consider this a nit and include this technical debt in the codebase :P.

mariusandra (approved):
One pending code smell, but willing to overlook. Otherwise LGTM.

PR #276, opened 2021-03-23
Cache distinct_id lookups

macobo:
This distinct_id lookup query was the last cumulatively slow query we
have on plugin-server. This can be quite easily cached as well, so went
ahead and did a small implementation.

With 10 workers, 65 bytes per distinct_id, this is should take ~6.5MB +
LRU overhead of memory per plugin server instance.

## Checklist

-   [ ] Jest tests


mariusandra (approved):
LGTM

PR #283, opened 2021-03-24
Update settings information

Twixes:
## Changes

We've added some some settings recently and some were not described, this updates to `config.ts` and `README.md` to fix that.

## Checklist

-   [x] Updated Settings section in README.md, if settings are affected\


mariusandra (approved):

macobo:
Code-wise we could keep this in sync easier if we didn't `extend Record<string, any>`. Would not help with the readme though.

Twixes:
WDYM by that @macobo?

Twixes:
That's true, how about moving config _under_ `PluginsServer` as property `config` instead of it being merged with the server obj? A matter for another PR ofc

PR #284, opened 2021-03-25
Allow whitelisted imports in babel

mariusandra:
## Changes

- PRs over issues, so here's a babel plugin to replace `import anything from 'anywhere'` with `const anything = __pluginHostImports['anywhere']['anything']`

## Checklist

-   [ ] Jest tests


macobo:
This approach has its own set of versioning issues - basically it's guaranteed someone will be caught out by us using a different library version from what they have specified in `package.json`.

---

Looking at the implementation: So if I did `import * from 'random-library'` this would silently succeed until used?

Might a better approach be to throw during initialization as this _should_ make `loadPlugins` fail.

Also we should have tests for running this style of plugin.

mariusandra:
Addressing the feedback:
- Just like segment, we need to publish the list of versions that are installed. We'd have to do this anyway if the package is in a global or in meta.
- This [will throw](https://github.com/PostHog/plugin-server/pull/284/files#diff-2c308a1b1d52d9d400882a689405d7ec89ef19b3af0b0a0a39e501f506a7179bR15) unless you import from a real library
- This does throw when initializing, there's even [a test](https://github.com/PostHog/plugin-server/pull/284/files#diff-6e38f762b5bf4bb69a0b2523e635853ee3cd07d5a706c2b2cafdbdb6e609a27dR175) to catch this.
- For running tests with this type of plugin, I just added to `vm.test.ts` one test that imports and then uses `fetch`

src/worker/vm/vm.ts
```diff
@@ -15,7 +16,11 @@ export async function createPluginConfigVM(
     pluginConfig: PluginConfig, // NB! might have team_id = 0
     indexJs: string
 ): Promise<PluginConfigVMReponse> {
-    const transformedCode = transformCode(indexJs, server)
+    const imports = {
+        'node-fetch': { default: fetch },
```
Twixes, line 20: Possibly dumb question: by providing `fetch` outright, will nobody be able to access some properties in a way that affects other VMs (e.g. redefine `fetch.isRedirect`)?

mariusandra: Definitely not a dumb question. VM2 should wrap everything in proxies. Worth testing it out though. This PR won't really change anything security-wise though, so might be worth tackling after.

macobo (approved):
Implementation lgtm! :)

mariusandra:
Thanks! Just some flaky tests being flaky here. I'll run them one more time (have to -> merged with master) and if it's just the `pisinca.destroy` line that fails, I'll merge this in.

PR #287, opened 2021-03-26
Fix inserting people

timgl:
## Changes

This is a stub, i wasn't actually able to get the tests running locally (I think because I need to run Kafka?) but here's what I found while digging into a really annoying issue, where users would still show up even though they were explicitly filtered

```
 Person id 0178479a-4999-0000-0ada-4197a37d65ac
   timestamp 2021-03-18 23:09:56
  has email set
 distinct id 1786ba71965d91-0b95953216708b-1e340521-249f00-1786ba71966de1

BUT there is also a person 01786ba7-1df5-0000-48fb-e78c57a2de8e
   timestamp 2021-03-25 23:10:17
   no properties
   with the same distinct id 1786ba71965d91-0b95953216708b-1e340521-249f00-1786ba71966de1
   ^ this person does not exist on postgres
```

My hunch is that this second person was created even though that insert query error'd in Postgres (because the distinct id already exists). @Twixes @macobo do we think I'm on the right track here?

## Checklist

-   [ ] Jest tests

### DIFF
--- a/tests/clickhouse/postgres-parity.test.ts
+++ b/tests/clickhouse/postgres-parity.test.ts
@@ -97,6 +97,29 @@ describe('postgres parity', () => {
         expect(postgresDistinctIds).toEqual(['distinct1', 'distinct2'])
 
         expect(person).toEqual(postgresPersons[0])
+
+        // insert same person, no properties. Should not actually insert
+        const second_person = await server.db.createPerson(DateTime.utc(), {}, team.id, null, true, uuid, [
+            'distinct1',
+            'distinct2',
+        ])
+        await delayUntilEventIngested(() => server.db.fetchPersons(Database.ClickHouse))
+        await delayUntilEventIngested(() => server.db.fetchDistinctIdValues(person, Database.ClickHouse), 2)
+
+        const secondClickHousePersons = await server.db.fetchPersons(Database.ClickHouse)
+        expect(clickHousePersons).toEqual([
+            {
+                id: uuid,
+                created_at: expect.any(String), // '2021-02-04 00:18:26.472',
+                team_id: team.id,
+                properties: '{"userProp":"propValue"}',
+                is_identified: 1,
+                _timestamp: expect.any(String),
+                _offset: expect.any(Number),
+            },
+        ])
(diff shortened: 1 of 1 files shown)


mariusandra:
Browsing through the events, this is what I've learned:

- The original user (with 400+ distinct ids) was created 2021-03-18, the duplicate user was made 2021-03-25 23:10:17 (a week later)
- The duplicate user is from a session where the events started about a second prior, and this session started with someone filling in a login form and then opening a /admin/... url
- Within a second from each other there are actually two $identify calls, with different anonymous ids (`1786ba71965d91-0b95953216708b-1e340521-249f00-1786ba71966de1` with event `01786ba7-24b7-0000-12ad-47306406545b ` and `1786ba71d91306-0ef293ed3d3f41-1e340521-249f00-1786ba71d92c52` with event `01786ba7-1f67-0000-22d1-e8c55331d1a2` that came a second earlier)
- Tracing the code, the user should be [merged](https://github.com/PostHog/plugin-server/blob/87e8c1ecc2b79556455e9e6a6df5a9c64cb9f09a/src/worker/ingestion/process-event.ts#L298) fine ... and then we should [delete the old user](https://github.com/PostHog/plugin-server/blob/87e8c1ecc2b79556455e9e6a6df5a9c64cb9f09a/src/shared/db.ts#L382), with an `ALTER` statement.

Something in the last step went wrong and the person wasn't deleted...

Twixes:
This has been a draft for a while, do we want to get this merged?

timgl:
@mariusandra didn't we fix this some other way?

mariusandra:
It's not really fixed, see [this internal thread](https://github.com/PostHog/internal/issues/236#issuecomment-822229862). We had a "quickfix" where we synced the persons table from postgres to clickhouse (dropping and overriding whatever was there). However the real fix is to change the merge engine behind persons.

That's a task that's big and delicate enough that it'll require some coordination between all the teams...

mariusandra:
Also, I'd close this. We currently do expect `createPerson` to fail if there's another person already in the database. The logic for handling duplicates is in the `updatePersonProperties` / `alias` / `capture` level, not in the raw database `create[Tablename]` level.

PR #288, opened 2021-03-26
Fix shouldSendWebhooks conditions

Twixes:
## Changes

Since #271 got merged, we've got reports from Cloud users that their webhooks stopped firing. Looks like the conditions in the reworked cached `shouldSendWebhooks` were incorrect, resulting in the method wrongly returning `false` for projects with non-empty `slack_incoming_webhook`, and only returning `true` for projects with Zapier hooks (which meant webhooks kept working for our own project).

## Checklist

-   [ ] Jest tests

### DIFF
--- a/src/worker/ingestion/team-manager.ts
+++ b/src/worker/ingestion/team-manager.ts
@@ -5,12 +5,15 @@ import { DB } from '../../shared/db'
 import { timeoutGuard } from '../../shared/ingestion/utils'
 import { Team, TeamId } from '../../types'
 
-type TeamWithEventUuid = Team & { __fetch_event_uuid?: string }
+interface TeamWithEventUuid extends Team {
+    __fetch_event_uuid?: string
+}
+type TeamCache<T> = Map<TeamId, [T, number]>
 
 export class TeamManager {
     db: DB
-    teamCache: Map<TeamId, [TeamWithEventUuid | null, number]>
-    shouldSendWebhooksCache: Map<TeamId, [boolean, number]>
+    teamCache: TeamCache<TeamWithEventUuid | null>
+    shouldSendWebhooksCache: TeamCache<boolean>
 
     constructor(db: DB) {
         this.db = db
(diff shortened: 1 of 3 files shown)


macobo:
Can this have a test added that was previously failing? Tests are all under src/worker/ingestion/team-manager.ts

macobo:
Also can we define an alert for when `hooks.send_task` goes to zero.

src/worker/ingestion/team-manager.ts
```diff
@@ -44,15 +44,20 @@ export class TeamManager {
     }
 
     public async shouldSendWebhooks(teamId: number): Promise<boolean> {
-        const cachedValue = this.getByAge(this.shouldSendWebhooksCache, teamId)
+        const cachedValue = this.getByAge(this.shouldSendWebhooksCache, teamId, 10_000)
```
macobo, line 47: There shouldn't be a reason to change the time limit here.

Twixes: Hm, the reason is that 30 seconds will be confusingly long for users who enable the webhook URL in project settings and then don't see it actually firing when they enable it for an action right away. And the performance cost of 30 s vs 10 s is neglibible.

macobo: I'd keep this as-is, this was chosen on purpose.

I chose 30s because we were fetching 180MB/s on the worker nodes, almost all of it team fetches for a single team. Decreasing this will bump our network bill and increase load on postgres again - not ideal.

Twixes: I increased to 15 s but 30 would be _really_ not a good experience for users.

macobo: This should not be part of this PR. :-1: Reverting and creating an issue.

Basically this partially undoes performance optimizations in a PR that has otherwise little to do with that. It also creates some confusion since the cache is always effectively 15s not 30 given that the team is fetched for every event.

If we want to provide immediate feedback then 15s vs 30s is immaterial, but the difference means fetching our largest teams data 2x more in every worker thread - a considerable amount of data, which was just a week ago slowing down ingestion as a whole.

The proper fix would be separate - using redis pubsub to notify workers.

PR #293, opened 2021-04-01
Fix for non-string distinct IDs 

mariusandra:
## Changes

- These should have been sanitized in the API before reaching the plugin server over Kafka, but let's clear it up anyway.
- Resolves issue for a customer who was using a RudderStack connector to get events into PostHog and had IDs as numbers.

## Checklist

-   [ ] Jest tests


timgl (approved):
LGTM

PR #294, opened 2021-04-05
Remove yargs

Twixes:
## Changes

This is a proposal to make plugin server startup simpler by removing yargs. It's additional complexity to support CLI features with both CLI and env var settings, while we only use this in backend service capacity, using env vars for config and README for reference. For me yargs only complicates code without adding anything actually useful - though maybe there's a different perspective?


mariusandra (approved):
Hey, mixed feelings somehow :). I won't miss them, though it is nice to do `posthog-plugin-server --help` and see all the envs you can use to configure the server without looking at the code... and their defaults. Plus the `posthog-plugin-server --version` flag can be quite useful. Perhaps some users also use them, I'm not sure. Agreed though that configuring these settings via ENVs is a better approach.

image

That said, we're not a real CLI app (yet?) and while the existence of the arguments doesn't bother me, I won't miss them, so green light from my side to do what you want.

mariusandra:
Though... there should probably be *some* alternative help screen then when you load the app with a bunch of CLI arguments (and/or --version)... And the one provided by yargs seems solid :)

Twixes:
Agree with a help screen, I added this simple thing, which I think is better than the yargs one because it's specifically geared for env var reference:
--help

mariusandra (approved):

mariusandra:
Works for me. Tests are failing though.

PR #303, opened 2021-04-09
Kafka consumer pool

mariusandra:
## Changes

- WIP
- If `KAFKA_CONSUMER_CONCURRENY` is set to `> 1`, we will run that many kafka consumers concurrently, each of whom will process events sequentially.
- If `KAFKA_CONSUMER_CONCURRENY` is set to `1`, we will run one consumer and process events in batches like now.

## Checklist

-   [ ] Jest tests


mariusandra:
not going down this route

PR #305, opened 2021-04-09
Remove two lines of unused code :)

yakkomajuri:
## Changes

Since we have eslint configured to not check for unused vars, I guess some of these slip (I understand why we do it though).

Stumbled upon this during my plugin server :diving_mask: session

## Checklist

-   [ ] Jest tests

### DIFF
--- a/src/main/queue.ts
+++ b/src/main/queue.ts
@@ -2,7 +2,6 @@ import Piscina from '@posthog/piscina'
 import { PluginEvent } from '@posthog/plugin-scaffold'
 import * as Sentry from '@sentry/node'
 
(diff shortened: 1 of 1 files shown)


mariusandra (approved):
TIL: getting to know a new codebase == snorkeling

yakkomajuri:
the thing says `diving_mask` :D

mariusandra:
Yeah, you're right, snorkeling is more like opening up each code file, but not scrolling below the surface...

yakkomajuri:
Indeed, you have to dive deep into the pools to understand the plugin server

PR #306, opened 2021-04-12
Add @posthog/plugin-contrib and AWS SDK

mariusandra:
## Changes

Adds AWS SDK (needed for S3) and plugin contrib (needed for batching) for the plugin server.

With these in, a nice batched upload every 10MB would look like this:

```js
import { createBuffer } from '@posthog/plugin-contrib'
import { S3 } from 'aws-sdk'

export function setupPlugin({ global }) {
    global.s3 = new S3(...)
    global.buffer = createBuffer({ limit: 10 * 1024 * 1024, timeoutSeconds: 10 * 60, onFlush: (batch) => {
        console.log(`Flushing ${batch.length} events!`)
        global.s3.upload(...)
    }})
}

export function teardownPlugin({ global }) {
    global.buffer.flush()
}

export function processEvent(event, { config, global }) {
    global.buffer.add(event, JSON.stringify(event).length) // add(object, points)
    return event
}
```

## Checklist

-   [ ] Jest tests


yakkomajuri (approved):
Took a look at the buffer implementation - seems fine to me

Happy to cherry pick the version bump into the release

PR #307, opened 2021-04-12
Attempt making Kafka batching more efficient

Twixes:
## Changes

This will result in batches of Kafka messages being processed once the batch is at least 16 kB or 500 ms since the last once - whichever earlier. Should increase batch sizes on Cloud from 1-2 messages per batch, making processing more efficient.

### DIFF
--- a/src/main/ingestion/kafka-queue.ts
+++ b/src/main/ingestion/kafka-queue.ts
@@ -206,6 +206,8 @@ export class KafkaQueue implements Queue {
             groupId: 'clickhouse-ingestion',
             sessionTimeout: 60000,
             readUncommitted: false,
+            minBytes: 16_384,
+            maxWaitTimeInMs: 500,
         })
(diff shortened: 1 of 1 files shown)


yakkomajuri:
I guess we decided on Slack not to do this?

Twixes:
I'd still kind of like to see how this performs on Cloud.

mariusandra:
Perhaps these should be envs so we can set them up/down without releasing a new version of the plugin server? Defaulting to what we have now...

mariusandra:
That said, I still don't think this will actually change anything. If the plugin server is idle, it's fine to accept batches of size 1. It's even safer (what if the plugin crashes? there will be no other in-flight events then!). If the plugin server is too busy, the batch size will grow to up to ~300 organically.

mariusandra:
I'm going to close this. I think we're not really solving anything if we stall kafka when it's idle to have batches bigger than one.

PR #308, opened 2021-04-12
REDIS_URL is always set to a default value

mariusandra:
## Changes

This makes it usable with the helm charts

## Checklist

-   [ ] Jest tests


Twixes (approved):
:+1:

PR #310, opened 2021-04-12
MMDB refactor

Twixes:
## Changes

This separates MMDB features used by the main thread and worker threads.

## Checklist

-   [x] Jest tests


mariusandra (approved):
why not

PR #314, opened 2021-04-13
Teardown on reload

mariusandra:
## Changes

Runs `teardownPlugin` also if reloading/replacing a plugin. Not only on shutdown.

## Checklist

-   [ ] Jest tests


src/worker/plugins/teardown.ts
```diff
@@ -1,11 +1,16 @@
...
 
-export async function teardownPlugins(server: PluginsServer): Promise<void> {
-    const { pluginConfigs } = server
+export async function teardownPlugins(
+    server: PluginsServer,
+    pluginConfigs: PluginConfig[] | undefined = undefined
+): Promise<void> {
+    if (!pluginConfigs) {
+        pluginConfigs = Array.from(server.pluginConfigs.values())
+    }
```
Twixes, line 10: Why not deconstruct in loop like before?

src/worker/plugins/setup.ts
```diff
@@ -24,6 +25,10 @@ export async function setupPlugins(server: PluginsServer): Promise<void> {
         } else {
             pluginConfig.vm = new LazyPluginVM()
             pluginVMLoadPromises.push(loadPlugin(server, pluginConfig))
+
+            if (prevConfig && (await prevConfig?.vm?.getTeardownPlugin())) {
```
Twixes, line 29: I think it'd be alright to avoid calling `getTeardownPlugin()` since `teardownPlugins` already does it and skips as needed?

tests/postgres/worker.test.ts
```diff
@@ -22,7 +22,7 @@ jest.mock('../../src/worker/ingestion/ingest-event')
 jest.mock('../../src/worker/plugins/run')
 jest.mock('../../src/worker/plugins/setup')
 jest.mock('../../src/worker/plugins/teardown')
-jest.setTimeout(600000) // 600 sec timeout
+jest.setTimeout(1000000) // 600 sec timeout
```
mariusandra, line 25: I'll mind a bit if you don't change it back :)

src/worker/plugins/teardown.ts
mariusandra: I cleaned it up a bit by changing the function's argument from an array to a single `PluginConfig`. The `Array.from()` code from before was to make TS happy. With the new approach it's not needed.

src/worker/plugins/setup.ts
mariusandra: Makes sense. Removed the check from here, as it was also somehow leaking abstractions (details from within `teardownPlugins` made their way out into `setupPlugins`...)

tests/postgres/worker.test.ts
mariusandra: I reverted it...

Twixes (approved):

PR #318, opened 2021-04-15
Rework console extension to send data to Kafka>ClickHouse or Postgres

Twixes:
## Changes

Plugin server changes for https://github.com/PostHog/posthog/pull/3482.

## Checklist

-   [x] Jest tests


src/worker/vm/extensions/console.ts
```diff
@@ -1,11 +1,45 @@
...
+
+export function createConsole(server: PluginsServer, pluginConfig: PluginConfig): ConsoleExtension {
+    function consoleRegister(type: PluginLogEntryType, ...args: unknown[]): void {
+        const entry: PluginLogEntry = {
+            id: new UUIDT().toString(),
+            team_id: pluginConfig.team_id,
+            plugin_id: pluginConfig.plugin_id,
+            timestamp: new Date().toISOString().replace('T', ' ').replace('Z', ''),
+            type,
+            message: args.map(String).join(' '),
```
yakkomajuri, line 16: If an object passes through this, won't it come out as `[object Object]`? 

How about `JSON.stringify`? That should handle numbers, booleans, etc too - although it has its own limitations:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#description

yakkomajuri (commented):
Sincerely naive question, but how come not just use Postgres for this given we still store data there for both EE and FOSS? Is it a matter of future-proofing and keeping as much of the new additions specific to one db?

src/worker/vm/extensions/console.ts
Twixes: You're correct about objects, thanks, fixed that.

Twixes:
Given the potentially high data volume nature of this feature, ClickHouse is definitely more future-proof and only marginally more complicated. @yakkomajuri

yakkomajuri:
Tests are still expecting stuff like `console.log` to be called. Even though we're piping logs somewhere else, we could still call  the native console methods, no? 

Either way, if the answer is no, I guess tests still have to be updated

Twixes:
Yep, adjusted for that @yakkomajuri. FYI in-plugin `console` method calls are now `status`'d to stdout in dev mode (`DEBUG` truthy or `NODE_ENV` is `dev`).

Twixes:
PR reviewable.
This functionality (including its tests) is completely disabled by default currently though, configurable with setting `ENABLE_PERSISTENT_CONSOLE`. That's because we should first merge this to make PostHog/posthog#3482 work end-to-end, then merge PostHog/posthog#3482, and then remove `ENABLE_PERSISTENT_CONSOLE` when the functionality can be put into production (mind that PostHog Cloud will require a migration to be manually initiated for this to work, Infra teams knows about this).

Checks

Top green check is the end goal state - persistent console fully tested and enabled by default. That ran with PostHog/posthog#3482 as the main repo base though, so until that is merged, the version merged will have some `if (!server.ENABLE_PERSISTENT_CONSOLE) return` conditions – bottom green check.

src/shared/db.ts
```diff
@@ -580,4 +585,51 @@ export class DB {
...
+        instanceId: UUID
+    ): Promise<PluginLogEntry> {
+        const entry: PluginLogEntry = {
+            id: new UUIDT().toString(),
+            team_id: pluginConfig.team_id,
+            plugin_id: pluginConfig.plugin_id,
+            timestamp: new Date().toISOString().replace('T', ' ').replace('Z', ''),
+            type,
+            message,
+            instance_id: instanceId.toString(),
```
mariusandra, line 612: Thinking out loud: could it make sense to store `pluginConfig` id as well in the tables? Also below in `pluginDigest`. Filtering by `pluginconfig_id` feels a bit safer somehow: no risk of forgetting `team_id` and showing the wrong things.

Twixes: Why not, makes sense

src/worker/vm/extensions/console.ts
```diff
@@ -1,11 +1,34 @@
...
+        }
+
+        if (!server.ENABLE_PERSISTENT_CONSOLE) {
+            return
+        }
+
+        const message = args
+            .map((arg) => {
+                const argString = String(arg)
+                return argString === '[object Object]' ? JSON.stringify(arg, null, 4) : argString
```
yakkomajuri, line 20: Being nitpicky here again but now this will lead to arrays being represented without brackets, which can be confusing:

```
> String([1,2,3]) // "1,2,3"
> JSON.stringify([1,2,3]) // "[1,2,3]"
```

Twixes: :+1: Addressed this case.

mariusandra (approved):
LGTM. Merge whenever convenient.

PR #320, opened 2021-04-15
Add PLUGIN_SERVER_IDLE env var for development

Twixes:
## Changes

This adds env var `PLUGIN_SERVER_IDLE` enabling `AlternativeMode.Idle`, which is a simple env var-based way of turning off the plugin server, specifically for developing the plugin server in tandem with the Django server.


yakkomajuri (approved):
What's the benefit of having it idle over just not running it?

I guess it's something you'd pass to the Django start script?

Code seems straightforward to me

Twixes:
Well, to otherwise not running you have to edit `bin/start-worker`, which messes with git. With an env var you can just… pass an env var whenever you want to use an external (e.g. development) plugin server.

Twixes:
Well, to otherwise not running you have to edit `bin/start-worker`, which messes with git. With an env var you can just… pass an env var whenever you want to use an external (e.g. development) plugin server.

Issue #280, opened 2021-03-24
Don't process $snapshot via plugins


macobo:
Provocative title, I know.

Currently $snapshot events are treated like any other and processed by plugins. However:
1. Modifying the payload data could easily break the session recording feature or display misleading information on the screen.
1. The data payload is chunked and compressed - making it hard to make anything out of it.
1. The payload data contains pretty large data that we need to serialize to send back-and-forth from consumer to worker to VM and back again - this hurts performance.
1. Plugin authors might not know about treating $snapshot events in a special way.

The only usecase for plugins is syncing it to other data sources and playing back via rrweb there.

I'd propose ignoring snapshot events for now and in the future perhaps add a `processSnapshotBatch` or similar function to plugins if the syncing scenario comes up. This function would not be allowed to modify the data though.

mariusandra:
Import and export plugins are the only ones that legitimately need to access this event. Handling them in a separate `processSnapshot` function makes sense and shouldn't be too hard to implement. 👍 for skipping them in `processEvent`.

Regarding `processSnapshotBatch`, this is a side quest and better suited for #242, but we might benefit from deprecating `processEventBatch` altogether. In reality, in a well performing system, 99% time when `processEventBatch` gets called, it gets just one event. The original motivation behind this function was that you for example get 100 events at once and then upload them to S3 or wherever. In reality, this won't really work. The bigquery export for example uses some internal buffering queue to batch its uploads... and we'll need something similar for S3. (Plus a `onShutdown` function to flush it... but that's another story)

yakkomajuri:
Interestingly, I had a call with a user recently about this. He would want to actually do some processing on session recording events.

We should probably make a determination otherwise people could actually start writing plugins that target `$snapshot` events

mariusandra:
@yakkomajuri what kind of processing was this user looking at?

If nothing substantial, I'd propose we:

- remove `$snapshots` from `processEvent`. 
- implement the async `onEvent` as discussed here: https://github.com/PostHog/plugin-server/issues/328
- implement a async `onSnapshot` to complement `onEvent` that just gets session recording events. Plugin authors can choose what to target, or both (`export const onSnapshot = onEvent`)

PR #324, opened 2021-04-20
Plugin Running Times Metrics

yakkomajuri:
## Changes

Adding PostHog metrics on time taken for plugins to run. Allows for charts like this:

Screenshot 2021-04-20 at 12 01 07

This adds timers alongside StatsD, and also:

- Makes `posthog` available anywhere and more configurable
- Adds a set of events that should be ignored by plugins. I had originally added `$snapshot` to that list given #280 but I guess we need `onEvent` before we can do that, otherwise it would affect export plugins. `$plugin_running_duration` has to be ignored to prevent an infinite event generator 

## Checklist

-   [ ] Jest tests


src/shared/posthog.ts
```diff
@@ -0,0 +1,8 @@
+import * as fetch from 'node-fetch'
+import { nodePostHog } from 'posthog-js-lite/dist/src/targets/node'
+
+import { defaultConfig } from './config'
+
+const posthog = nodePostHog(defaultConfig.POSTHOG_PROJECT_API_KEY, { fetch, apiHost: defaultConfig.POSTHOG_API_HOST })
```
mariusandra, line 6: nit: `export const posthog = ...`

src/worker/ingestion/process-event.ts
```diff
@@ -52,9 +49,8 @@ export class EventsProcessor {
         this.teamManager = new TeamManager(pluginsServer.db)
         this.personManager = new PersonManager(pluginsServer)
 
-        this.posthog = nodePostHog('sTMFPsFhdP1Ssg', { fetch })
         if (process.env.NODE_ENV === 'test') {
-            this.posthog.optOut()
+            posthog.optOut()
         }
```
mariusandra, line 54: This now changes a globally exported var deep inside the event processing code. That feels a bit off... and if I look at shared/posthog, I'd never expect that I also need to look here to find some opt out logic.

Thus I think this check could easily go in shared/posthog.

src/worker/plugins/run.ts
```diff
@@ -19,8 +25,16 @@ export async function runPlugins(server: PluginsServer, event: PluginEvent): Pro
...
 
+            const timeSpentRunning = new Date().getTime() - timer.getTime()
+            server.statsd?.timing(`plugin.${pluginConfig.plugin?.name}.process_event`, timer)
+            posthog.capture('$plugin_running_duration', {
+                team: event.team_id,
+                plugin: pluginConfig.plugin?.name,
+                time_ms: timeSpentRunning,
+                time_seconds: Math.round(timeSpentRunning / 10) / 1000,
+                function: 'processEvent',
+            })
```
mariusandra, line 37: This effectively doubles the amount of events we emit. For capturing aggregated metrics, that's way too much volume.

src/worker/plugins/run.ts
```diff
@@ -1,9 +1,15 @@
...
 import { processError } from '../../shared/error'
+import { posthog } from '../../shared/posthog'
 import { PluginConfig, PluginsServer } from '../../types'
 
+const EVENTS_TO_IGNORE = ['$plugin_running_duration']
+
 export async function runPlugins(server: PluginsServer, event: PluginEvent): Promise<PluginEvent | null> {
+    if (EVENTS_TO_IGNORE.includes(event.event)) {
+        return event
+    }
```
mariusandra, line 12: Hot tip, if you ever find yourself building escape hatches around large blocks of code, it's probably a [code smell](https://en.wikipedia.org/wiki/Code_smell) and something's off with the abstraction or implementation.

In this case, we send the `$plugin_running_duration` event to app.posthog.com, not to the self-hosted instance that's processing events. Hence we don't really have much do escape here.

src/worker/plugins/run.ts
mariusandra: I'd rather aggregate this information and return it in an instance status report, perhaps once per hour or whenever. We could also add the instance ID into this report to help deduping.

src/worker/plugins/run.ts
```diff
@@ -58,11 +75,20 @@ export async function runPluginsOnBatch(server: PluginsServer, batch: PluginEven
...
                     plugin: pluginConfig.plugin?.name ?? '?',
                     teamId: teamId.toString(),
                 })
+                posthog.capture('$plugin_running_duration', {
+                    team: teamId,
+                    plugin: pluginConfig.plugin?.name,
+                    time_ms: timeSpentRunning,
+                    time_seconds: Math.round(timeSpentRunning / 10) / 1000,
+                    function: 'processEventBatch',
+                })
```
mariusandra, line 91: same feedback as above, yet also having a 7-line code block like this identical to one a few lines above (and one a few lines below) screams DRY and is another smell that something can be consolidated.

tests/postgres/worker.test.ts
```diff
@@ -22,7 +22,7 @@ jest.mock('../../src/worker/ingestion/ingest-event')
 jest.mock('../../src/worker/plugins/run')
 jest.mock('../../src/worker/plugins/setup')
 jest.mock('../../src/worker/plugins/teardown')
-jest.setTimeout(600000) // 600 sec timeout
+jest.setTimeout(1000000) // 1000 sec timeout
```
mariusandra, line 25: why the change?

README.md
```diff
@@ -67,6 +67,8 @@ There's a multitude of settings you can use to control the plugin server. Use th
 | INTERNAL_MMDB_SERVER_PORT     | port of the internal server used for IP location (0 means random) | `0`                                   |
 | DISTINCT_ID_LRU_SIZE          | size of persons distinct ID LRU cache                             | `10000`                               |
 | PLUGIN_SERVER_IDLE            | whether to disengage the plugin server, e.g. for development      | `false`                               |
+| POSTHOG_PROJECT_API_KEY       | PostHog project API key for analytics                             | `sTMFPsFhdP1Ssg`                      |
+| POSTHOG_API_HOST              | address of the PostHog instance used for analytics                | `https://app.posthog.com`             |
```
mariusandra, line 71: It's not totally clear from the descriptions what these are used for... and if or why we would want them to be anything other than the hardcoded production values, except perhaps locally when testing. Having them exposed like this makes it very easy to opt-out of telemetry... which maybe we want, but then it should be handled in a similar way as in the main posthog app

mariusandra (changes requested):
Plenty of code feedback below. Plus one bigger issue: this  PR sends a new event to track metrics per every ingested event... which is way too much volume, effectively doubling what we currently do. Instead we must batch this info and just transmit like we currently transmit the instance status reports in the main app.

src/worker/plugins/run.ts
yakkomajuri: Hmm but app will also be sending `$plugin_running_duration` to app, no?

tests/postgres/worker.test.ts
yakkomajuri: So I definitely was annoying here: https://github.com/PostHog/plugin-server/pull/314/commits/5626c5dea5ea4197b5e2b5a648f6e621fe8436cb

Sorry for not reverting that - but the worker benchmarks were failing on that PR. I guess the timeout is somewhat part of the benchmark too? The thing is that it was throwing a rather cryptic error

src/worker/plugins/run.ts
mariusandra: I guess you had the escape hatch here because every event generated another event and hence without it (and with the posthog.capture below) we would get stuck in an endless loop.

If we send this event only hourly or otherwise periodically, it won't matter if we also measure how long a few plugins run on it.

yakkomajuri:
New approach still WIP, need to smoothen some edges and update tests.

src/worker/plugins/run.ts
yakkomajuri: Yeah exactly, my comment was related to your above comment but with the new approach this is irrelevant

yakkomajuri:
The conflicts are running a bit deep so I think I'll just start fresh

Issue #322, opened 2021-04-16
$increment for user properties


yakkomajuri:
Documenting this but I could probably pick it up at some point.

I'd love to be able to increment numerical user properties using something like `$increment`, which would be part of the `$set`/`$set_once` "special properties" group.

I can see a lot of use cases for this, like total revenue for user, number of times logged in, number of people referred, total time spent on the platform, etc.

These can be done with events (+ sum of props) somewhat, but with some limitations:

- I won't see them readily available on the user page
- Most importantly, I can't filter by them: I can say `sum of property $amount_paid on event $purchase for user with email=john@doe.com`, but I can't get a view of all users with lifetime revenue greater than 10k or something

The alternative currently is to pull person props via the API (using a personal API key), add your value to them, and reset the properties.

PR #326, opened 2021-04-20
Add $increment for numerical user props

yakkomajuri:
## Changes

_Please describe._

Adds `$increment` as the new magical property on events as suggested here #322 

More context on the issue.

The behavior is:
- If the prop in `$increment` is non-numerical, don't do anything
- If it is numerical, create a new prop with the initial value or increment an existing property

## Checklist

-   [ ] Jest tests


yakkomajuri:
If this makes sense @mariusandra @Twixes, I'll go ahead and add `posthog.people.increment` in `posthog-js`

src/worker/ingestion/process-event.ts
```diff
@@ -230,8 +231,17 @@ export class EventsProcessor {
...
+
         const updatedProperties: Properties = { ...propertiesOnce, ...personFound.properties, ...properties }
 
+        for (const [key, val] of Object.entries(incrementProperties)) {
+            if (updatedProperties[key]) {
+                updatedProperties[key] += val
+            } else {
+                updatedProperties[key] = val
+            }
+        }
```
Twixes, line 243: This may result in some race conditions, would be better to do in SQL if possible.

Twixes (commented):
Makes sense to me BUT I will cheekily not approve this until there's a documentation update PR ready in the posthog.com repo (and I encourage y'all to make us write more documentation too, because it's crucial for getting customers to use extensibility features)

yakkomajuri:
I like that - thanks for pushing for more docs. Will put a PR for it.

Also I guess the problem is the increment without an existing prop, because increments on existing props will reach eventual consistency. Rubber ducking at this point but yes will look into an updated approach.

src/worker/ingestion/process-event.ts
yakkomajuri: hmm yes that's true - let me look into another approach

mariusandra: Yep, this should normally be done inside SQL if possible. Googling gave something [like this](https://stackoverflow.com/questions/25957937/how-to-increment-value-in-postgres-update-statement-on-json-key/39076637):

```sql
UPDATE foo 
SET data = jsonb_set(data, '{bar}', (COALESCE(data->>'bar','0')::int + 1)::text::jsonb)
WHERE id = 1;
```

mariusandra:
I love the feature, but we might need to HODL it until the issue with user prop merging in clickhouse is solved :/. This would massively increase the nr of duplicate users.

CC @fuziontech @EDsCODE

src/worker/ingestion/process-event.ts
yakkomajuri: Yeah I was looking into this as well - `jsonb` is great, but in CH we store properties as `varchar`...

Does give me an excuse to look at CH a bit more though

mariusandra: Yeah, there will be no property updates in clickhouse... only a new insert with a new full array... :/

mariusandra: Regarding clickhouse, since postgres is the master data store for persons, we can just do the update in postgres (inline like the query above), fetch the returned row (and its properties) and then store that in clickhouse. Theoretically nothing should override anything this way.

fuziontech:
I say let's do it and see how it effects things. Worst case we roll this back? We can always correct the issue with a sync from PG -> Clickhouse if it does catch on fire.

yakkomajuri:
On the topic of race conditions:

https://github.com/PostHog/posthog/issues/4082

yakkomajuri:
Finally had some time to work on this today.

Still working on smoothening the edges here and will explain the query tomorrow.

yakkomajuri:
Merging `master` broke the tests :D

Anyway, this is ready for a first look. The update query is probably the key part here, so here's a brieft overview of how I got there:

[This](https://github.com/PostHog/plugin-server/pull/326/files/d4906b1ed0ff9078c1a3a6780475019263722e03#r617304919) was a good starting point but unfortunately the approach only works for updating one property. 

Then I [landed here](https://stackoverflow.com/questions/40347580/postgresql-jsonb-update-multiple-keys-in-one-query), talking about updating multiple properties. However, I thought limitless nested `jsonb_set` calls would become quite a messy query to build and debug though (imagine someone tries to increment 10 props at once). 

So after a Postgres docs session I came up with this (using wrong SQL formatting so it's easier to read here):

```sql
UPDATE posthog_person
SET properties = 
    properties ||  jsonb_build_object('prop2', (COALESCE(properties->>'prop2','0')::bigint + 10))
WHERE id = 1 
RETURNING properties;
```

This does the equivalent of a JS expression like:

```js
{ ...{ "prop1": "hello", "prop2": 10 }, ...{ "prop2": 10 + 10 } }
```

Using the object concatenation with a new object  was the way I found to avoid the nested mess of `jsonb_set`.

However, this didn't handle a few edge cases well, particularly overflows and updates to non-numeric props (we can't do these checks in JS because one has to expect a new prop might be added by a separate thread after we got the "snapshot" of current props). 

Hence the query had to get a bit more complex, landing here:

```sql
UPDATE posthog_person
SET properties = 
    -- get all current properties
    properties || 
    -- check that:
    -- 1: if there's an existing prop, it can be cast to int
    -- 2: adding up the values won't cause an overflow
    CASE WHEN (COALESCE(properties->>'a', '0')~E'^\\d+$' AND (COALESCE(properties->>'a', '0')::float/(2^63-1) + 100::float/(2^63-1)) < 1)
        -- build a new object that creates and/or increments a prop
        THEN jsonb_build_object('a', (COALESCE(properties->>'a','0')::int + 10))
        ELSE '{}'
    END
WHERE id = 1 
-- return properties so we can finish the full update of props
RETURNING properties;
```

This should let us safely increment as many properties as we want at once.

Happy to get harsh feedback on this.

src/worker/ingestion/process-event.ts
```diff
@@ -237,12 +239,30 @@ export class EventsProcessor {
...
+        // CH still needs to update if there are $increment props but Postgres has already done so
+        if (arePersonsEqualExcludingIncrement && (!this.db.kafkaProducer || !areTherePropsToIncrement)) {
             return personFound
         }
 
+        if (incrementedPropertiesQueryResult && incrementedPropertiesQueryResult.rows.length > 0) {
+            for (const [key, val] of Object.entries(incrementedPropertiesQueryResult.rows[0].properties)) {
+                updatedProperties[key] = val
+            }
+        }
```
mariusandra, line 264: Could this be simplified to just `Object.assign(updatedProperties, incrementedPropertiesQueryResult.rows[0].properties)`... or then with some `{...updatedProperties, ...other}` logic?

mariusandra (commented):
This is awesome! Great work :). 

Regarding the actual implementation:
- Might be nice to add a few more tests to check what happens if you use: a) negative integers, b) floats, c) NaN
- Can we cast into the postgres `numeric` type and skip the overflow issue?

With those (and one comment inline as well) resolved, I think this is good to merge.

That's because while this PR solves the issue for $increment, we could still undo the incremented values if a `$set` is running at the same time. To get around this, we could also move all the properties in `$set` to the `jsonb_build_object ` system... or we could take a step back and denormalise the properties altogether: https://github.com/PostHog/posthog/issues/4272

src/worker/ingestion/process-event.ts
yakkomajuri: ah yes great point! This could definitely be simplified, reused some logic here

fuziontech:
>Can we cast into the postgres numeric type and skip the overflow issue?

Yes please! This will save us a ton of pain down the road.

yakkomajuri:
Feels like Cypress :D
Screenshot 2021-05-11 at 09 43 06

mariusandra:
Let's support decrementing as well. It has many uses, from corrections to data... to tracking balances... to [fill in the blank].

Twixes:
Definitely was like Cypress 😅 because of CH insertion via Kafka being async. Fixed in master with #376.

mariusandra:
I think `$increment` and `$increment_by` make sense. Manually negating decrement values is too much to explain and maintain.

neilkakkar (approved):
Looks good!

yakkomajuri:
Docs coming with the latest Array PR

Issue #269, opened 2021-03-22
Plugin audit log on events


mariusandra:
Currently if for example one export plugin stops working (export server throws 500), the plugin is ignored and we won't know what event it is unless we do a full diff for event IDs in both posthog and the export destination.

There should be an easier way. Building support for retries ([comment below the line here](https://github.com/PostHog/plugin-server/issues/156#issue-809224435), [internal discussion](https://posthog.slack.com/archives/C01F6QWTVB5/p1616390397000800)) would solve this, however just keeping a history of what plugins were applied to an event will give us enough information to make due in case of emergency.

I imagine we will just set a property on the ingested event, something like `event.properties['$plugins_ran'] = []` and `event.properties['$plugins_failed'] = []` and then push something to those arrays after plugins run. For example `event.properties['$plugins_failed'].push(`${plugin.id}-${toSlug(plugin.name)}`)...

This has not been asked by any user yet, though it might be useful to implement regardless.

Twixes:
I think this was addressed in #327.

PR #327, opened 2021-04-20
Add plugins audit log

yakkomajuri:
## Changes

Tackles #269 

I guess `$plugins_failed` will need to evolve once we have retries, but we don't yet, so...

## Checklist

-   [ ] Jest tests


src/worker/plugins/run.ts
```diff
@@ -15,9 +17,11 @@ export async function runPlugins(server: PluginsServer, event: PluginEvent): Pro
 
             try {
                 returnedEvent = (await processEvent(returnedEvent)) || null
+                pluginsRan.push(pluginConfig.plugin?.name)
```
mariusandra, line 20: I'd somehow add an ID here, so we could quickly look up what's needed. The id of a `pluginConfig` is probably the best. Something simple, just like: `${pluginConfig.plugin?.name} (${pluginConfig.id})` is probably enough.

src/worker/plugins/run.ts
```diff
@@ -27,6 +31,11 @@ export async function runPlugins(server: PluginsServer, event: PluginEvent): Pro
         }
     }
 
+    if (!returnedEvent.properties) {
+        returnedEvent.properties = {}
+    }
+    returnedEvent.properties['$plugins_ran'] = pluginsRan
+    returnedEvent.properties['$plugins_failed'] = pluginsFailed
```
mariusandra, line 38: Perhaps no need to add these if no plugins ran?

mariusandra (commented):
Thanks for making progress on this! Some comments/suggestions in the review.

yakkomajuri:
Also renamed to `$plugins_ran_successfully` because plugins that failed did technically run

src/worker/plugins/run.ts
```diff
@@ -27,6 +31,16 @@ export async function runPlugins(server: PluginsServer, event: PluginEvent): Pro
...
 
+    if (!returnedEvent.properties) {
+        returnedEvent.properties = {}
+    }
+    if (pluginsRan.length > 0) {
+        returnedEvent.properties['$plugins_ran_successfully'] = pluginsRan
+    }
+    if (pluginsFailed.length > 0) {
+        returnedEvent.properties['$plugins_failed'] = pluginsFailed
+    }
```
mariusandra, line 42: Some code feedback then :). 

1) There's no need to run `if (!returnedEvent.properties) { returnedEvent.properties = {} }` if we're not going to change anything (when lengths are 0 for both arrays)
2) I think it's better that either both keys exist or none do, otherwise you'll end up second guessing yourself if the $plugins_failed array is just not there (instead of being empty, which tells you directly that nothing failed)

You could do something like:

```js
    if (pluginsRan.length > 0 || pluginsFailed.length > 0) {
        event.properties = { 
            ...event.properties,
            '$plugins_ran_successfully': pluginsRan,
            '$plugins_failed': pluginsFailed,
        }
    }
```

src/worker/plugins/run.ts
```diff
@@ -66,6 +83,16 @@ export async function runPluginsOnBatch(server: PluginsServer, batch: PluginEven
...
         }
 
+        for (const event of returnedEvents) {
+            if (event) {
+                if (!event.properties) {
+                    event.properties = {}
+                }
+                event.properties['$plugins_ran_successfully'] = pluginsRan
+                event.properties['$plugins_failed'] = pluginsFailed
+            }
```
mariusandra, line 93: Same here that we should only add these props if any plugins actually ran.

mariusandra (changes requested):
feedback left inline :)

src/worker/plugins/run.ts
```diff
@@ -66,6 +81,16 @@ export async function runPluginsOnBatch(server: PluginsServer, batch: PluginEven
             }
         }
 
+        for (const event of returnedEvents) {
+            if ((event && pluginsRan.length > 0) || pluginsFailed.length > 0) {
```
mariusandra, line 85: this `if` is looking funny... :)

yakkomajuri: huh formatter messed up there

mariusandra: Did it now? :P 
Or did you just forget the `()` and prettier just made the default operator precedence apparent? :)
Always when mixing `&&` and `||` put as many brackets around the different parts of the query as possible to avoid any errors...

mariusandra:
A flaky test that's not related to this PR is failing. So merging and will fix that later.

mariusandra:
Ah, no, I can't :D 

image

Will re-run and hope for the best.

src/worker/plugins/run.ts
```diff
@@ -27,6 +31,14 @@ export async function runPlugins(server: PluginsServer, event: PluginEvent): Pro
         }
     }
 
+    if (pluginsRan.length > 0 || pluginsFailed.length > 0) {
+        event.properties = {
+            ...event.properties,
+            $plugins_ran_successfully: pluginsRan,
```
Twixes, line 37: How about naming this `$plugins_succeeded`?

mariusandra: I like it. Changed.

PR #331, opened 2021-04-23
Refactor event names and properties

mariusandra:
## Changes

- Instead of updating huge JSON arrays on `Team`, insert unique event names and properties to dedicated tables
- Context here: https://github.com/PostHog/posthog/issues/4087

## Checklist

-   [ ] Jest tests


src/shared/db.ts
```diff
@@ -642,4 +644,14 @@ export class DB {
 
         return entry
     }
+
+    public async fetchEventDefinitions(): Promise<EventDefinitionType[]> {
+        return (await this.postgresQuery('SELECT * FROM posthog_eventdefinition', undefined, 'fetchAllPluginLogs'))
```
yakkomajuri, line 649: I guess the tag here lingered from a paste?

src/shared/db.ts
```diff
@@ -642,4 +644,14 @@ export class DB {
...
+            .rows as EventDefinitionType[]
+    }
+
+    public async fetchPropertyDefinitions(): Promise<PropertyDefinitionType[]> {
+        return (await this.postgresQuery('SELECT * FROM posthog_propertydefinition', undefined, 'fetchAllPluginLogs'))
```
yakkomajuri, line 654: same tag comment

src/shared/db.ts
mariusandra: whoops, fixed

src/shared/db.ts
mariusandra: same reply

src/shared/db.ts
```diff
@@ -642,4 +644,15 @@ export class DB {
 
         return entry
     }
+
+    public async fetchEventDefinitions(): Promise<EventDefinitionType[]> {
```
yakkomajuri, line 648: I guess these seem to only be used for tests? 

Also a minor thing is that this has same tag as a different query (one that filters by team)

src/worker/ingestion/team-manager.ts
```diff
@@ -101,42 +100,35 @@ export class TeamManager {
...
+                    [new UUIDT().toString(), key, typeof value === 'number', team.id],
+                    'insertPropertyDefinition'
+                )
+            }
         }
+
         if (team && !team.ingested_event) {
+            await this.db.postgresQuery(
+                `UPDATE posthog_team SET ingested_event = $1 WHERE id = $2`,
+                [true, team.id],
```
yakkomajuri, line 128: comment just purely for my own learning here: is it just our own standard to have all values be passed as an array? Cause in this case `true` is not dynamic so it could be in the query directly. Not a complaint, just wondering for my own purposes - I do get the readability/standardization this brings

src/worker/ingestion/team-manager.ts
```diff
@@ -101,42 +100,35 @@ export class TeamManager {
             event: event,
             ingested: team.ingested_event,
         })
-        let shouldUpdate = this.calculateUpdates(team, event, properties)
-        if (shouldUpdate && team.__fetch_event_uuid !== eventUuid) {
```
yakkomajuri, line 105: Based on these deletions it seems this function doesn't use `eventUuid` anymore?

src/worker/ingestion/team-manager.ts
```diff
@@ -101,42 +100,35 @@ export class TeamManager {
...
+            this.eventNamesCache.get(team.id)?.add(event)
+        }
+
+        for (const [key, value] of Object.entries(properties)) {
+            if (!this.eventPropertiesCache.get(team.id)?.has(key)) {
+                await this.db.postgresQuery(
+                    `INSERT INTO posthog_propertydefinition (id, name, is_numerical, volume_30_day, query_usage_30_day, team_id) VALUES ($1, $2, $3, NULL, NULL, $4) ON CONFLICT DO NOTHING`,
+                    [new UUIDT().toString(), key, typeof value === 'number', team.id],
+                    'insertPropertyDefinition'
+                )
```
yakkomajuri, line 121: With events we seem to add to the cache following the insert query, which we don't seem to do for props?

mariusandra: Oooh, nice catch

yakkomajuri (changes requested):
Well I was just driving by to learn a bit but ended up giving this a review - few nits and questions (for my own learning as well) but mostly I think the one change needed is to set the cache after inserting a new prop like we do for events (unless I misunderstood something here - totally possible)

tests/worker/ingestion/team-manager.test.ts
```diff
@@ -157,22 +154,23 @@ describe('TeamManager()', () => {
...
+            //     { event: '$pageview', usage_count: 2, volume: 3 },
+            //     { event: 'new-event', usage_count: null, volume: null },
+            // ])
+            // expect(team?.event_properties).toEqual(['property_name', 'numeric_prop', 'number'])
+            // expect(team?.event_properties_with_usage).toEqual([
+            //     { key: 'property_name', usage_count: null, volume: null },
+            //     { key: 'numeric_prop', usage_count: null, volume: null },
+            //     { key: 'number', usage_count: null, volume: null },
+            // ])
+            // expect(team?.event_properties_numerical).toEqual(['numeric_prop', 'number'])
```
Twixes, line 168: Shouldn't have commented out code in tests I think?

src/shared/db.ts
mariusandra: yep, they're only for tests. There are a few other fetch* functions like that in the db class, like `fetchElements` or `fetchEvents` that return everything without filters. They're useless in prod, but quite nice in tests. They could possibly be refactored to another file, but... not in this PR :)

src/worker/ingestion/team-manager.ts
mariusandra: No standard, just whatever does the job :)
In this specific case, I'm not sure why it's one way and not the other... and I think it doesn't really make any difference.

src/worker/ingestion/team-manager.ts
mariusandra: indeed! Strange that it wasn't show as unused for me and I missed it...

tests/worker/ingestion/team-manager.test.ts
mariusandra: ooh, how did I forget that in :(

mariusandra:
Feedback addressed! Ready for another look.

mariusandra:
So the tests are broken in really weird ways and I suspect some cache bug with GH actions:

image

Not sure how to resolve. Tests pass locally though.

mariusandra:
All green!

yakkomajuri (approved):
LGTM

mariusandra:
Let's merge it then and be one step closer to deprecating those huge fields

PR #333, opened 2021-04-26
Register system events and errors with plugin logs

### LINKED ISSUE #254: Pipe plugin errors to stdout on DEBUG
When developing locally, it would be nice to get the errors that we show on the `ERROR` tag in the UI on the terminal as well. 

cc @kpthatsme

Twixes:
## Changes

This adds registering of plugin errors and load/unload events with plugin logs. Closes #254.

## Checklist

-   [ ] Jest tests

### DIFF
--- a/src/shared/sql.ts
+++ b/src/shared/sql.ts
@@ -1,4 +1,12 @@
-import { Plugin, PluginAttachmentDB, PluginConfig, PluginConfigId, PluginError, PluginsServer } from '../types'
+import {
+    Plugin,
+    PluginAttachmentDB,
+    PluginConfig,
+    PluginConfigId,
+    PluginError,
+    PluginLogEntryType,
+    PluginsServer,
+} from '../types'
 
 function pluginConfigsInForceQuery(specificField?: keyof PluginConfig): string {
     return `SELECT posthog_pluginconfig.${specificField || '*'}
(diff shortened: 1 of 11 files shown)


Twixes:
There's a teardown test failing, but that's in `master` unfortuantely.

src/shared/sql.ts
```diff
@@ -36,10 +36,20 @@ export async function getPluginConfigRows(server: PluginsServer): Promise<Plugin
...
 ): Promise<void> {
     await server.db.postgresQuery('UPDATE posthog_pluginconfig SET error = $1 WHERE id = $2', [
         pluginError,
         typeof pluginConfig === 'object' ? pluginConfig?.id : pluginConfig,
     ])
+    if (pluginError && server.ENABLE_PERSISTENT_CONSOLE) {
+        await server.db.createPluginLogEntry(
+            pluginConfig,
+            PluginLogEntryType.Error,
+            true,
```
mariusandra, line 49: Hmm.. Maybe I'm wrong, but I think most of the time this function is called with not "is_system" errors, but just uncaught errors from within the plugin. I don't think those should be system logs/errors, but just plugin errors. System logs are indeed the "Plugin unloaded" type things from below...

Twixes: Makes, sense I replaced this with a varchar choice field `source`, which can be "CONSOLE", "SYSTEM", or "PLUGIN".

mariusandra (approved):

PR #336, opened 2021-04-27
Annotate update for plugin config errors to include timing

fuziontech:
## Changes

Adding metrics for the sql call to update plugin configs with the error flag. I've seen this one show up a number of times in the logs

Also adding more metrics for piscina worker pool. I have a hunch that the root cause of the issues is around piscina silently running out of capacity and stalling.

## Checklist

-   [ ] Jest tests

### DIFF
--- a/tests/sql.test.ts
+++ b/tests/sql.test.ts
@@ -133,8 +133,9 @@ test('setError', async () => {
 
     const pluginError: PluginError = { message: 'error happened', time: 'now' }
     await setError(server, pluginError, pluginConfig39)
-    expect(server.db.postgresQuery).toHaveBeenCalledWith('UPDATE posthog_pluginconfig SET error = $1 WHERE id = $2', [
-        pluginError,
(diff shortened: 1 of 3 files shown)


tests/sql.test.ts
```diff
@@ -125,10 +125,11 @@ test('setError', async () => {
...
 
     await setError(server, null, pluginConfig39)
-    expect(server.db.postgresQuery).toHaveBeenCalledWith('UPDATE posthog_pluginconfig SET error = $1 WHERE id = $2', [
-        null,
-        pluginConfig39.id,
-    ])
+    expect(server.db.postgresQuery).toHaveBeenCalledWith(
+        'UPDATE posthog_pluginconfig SET error = $1 WHERE id = $2',
+        [null, pluginConfig39.id],
+        'update_pc_error'
```
fuziontech, line 131: woops

mariusandra: 🙈

PR #341, opened 2021-04-27
Autorestart first steps

mariusandra:
## Changes

- Because THE SHOW MUST GO ON... and nobody wants to manually restart servers at 3am
- If the plugin server has not ingested an event in X time units (default off), send an alert to sentry.
- This should be set to something like 5min in cloud.
- TODO: also restart the server, but I'd like to first just get the alerting live before I add the SIGTERM.

## Checklist

-   [ ] Jest tests


Twixes (approved):
Should be good

PR #342, opened 2021-04-27
Sync clear error

mariusandra:
## Changes

- This `UPDATE posthog_pluginconfig SET error = $1 WHERE id = $2` seems to time out often. 
- Let's see if doing it sync (instead of delegating to the back of the JS event loop) fixes the stability we're having with this line.

## Checklist

-   [ ] Jest tests


Twixes (approved):
:+1: I don't see an important reason to await it

PR #343, opened 2021-04-27
Revert "Sync clear error"

macobo:
Reverts PostHog/plugin-server#342

Context from slack:
- Full thread at https://posthog.slack.com/archives/C0185UNBSJZ/p1619534371059900
- Michael pointed out that this range contains an error
- I originally marked this as void not await due to some deadlocking I was seeing locally - https://github.com/PostHog/plugin-server/pull/234/commits/7ebed6feec1132e5dcdb4da33a39510e1c667c6d as part of https://github.com/PostHog/plugin-server/pull/234

Not sure _why_ this would cause errors but it seems to be related to how pisciana/workers in node work.

PRs > Issues but will let you take it from here

### DIFF
--- a/src/shared/error.ts
+++ b/src/shared/error.ts
@@ -32,5 +32,8 @@ export async function processError(
 }
 
 export async function clearError(server: PluginsServer, pluginConfig: PluginConfig): Promise<void> {
-    await setError(server, null, pluginConfig)
+    // running this may causes weird deadlocks with piscina and vms, so avoiding if possible
+    if (pluginConfig.error) {
+        await setError(server, null, pluginConfig)
+    }
 }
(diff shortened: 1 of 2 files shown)


mariusandra:
Ouch and thanks for the context. That's a very weird deadlock. 

With `void` this was sometimes showing up as just a really long postgres query. I guess it was long for the same reason why it deadlocks otherwise. From the logs I only saw the `clearError` query have issues. Thus I added an `if` around it to hopefully prevent it from running more often than it should.

We're probably going to move away from this to relying more on plugin logs anyway, so this will eventually get handled better. :)

mariusandra (approved):

macobo:
Note not 100% sure this is the cause - worth deploying this and verifying. Just seems suspect since it has caused issues in the past

mariusandra:
Had to move the guard around to avoid manually fixing too many tests.

PR #346, opened 2021-04-27
Run kafkajs with eachMessage

mariusandra:
## Changes

- Currently we have a problem when our batches take longer than 30 seconds to ingest. Kafka might invalidate our session and since we only `commit()` at the end of a batch, the commit will "bounce" (we're rebalancing after all) and will be retried some other time. The batch will fail.
- This splits each ingested batch into a smaller batch of 40 events and then runs plugins on them and ingests them in parallel.
- After each batch offsets are committed (pending autoCommitInterval's approval), storing the state. If at this point the server decides to do rebalancing, the batch is cleanly halted at that moment. 
- Before we were ingest 300 events sequentially in parallel batches of 40 anyway, and just running all plugins totally in parallel before that. Now plugins will also be ran in groups of 40.
- Basically we'll be tearing down the orange boxes from [this graph](https://github.com/PostHog/plugin-server/issues/227).

## Checklist

-   [ ] Jest tests

### DIFF
--- a/src/main/ingestion/kafka-queue.ts
+++ b/src/main/ingestion/kafka-queue.ts
@@ -6,7 +6,7 @@ import { PluginsServer, Queue } from 'types'
 
 import { timeoutGuard } from '../../shared/ingestion/utils'
 import { status } from '../../shared/status'
-import { getPiscinaStats, groupIntoBatches, killGracefully, sanitizeEvent } from '../../shared/utils'
+import { groupIntoBatches, killGracefully, sanitizeEvent } from '../../shared/utils'
 
 export class KafkaQueue implements Queue {
     private pluginsServer: PluginsServer
@@ -73,7 +73,6 @@ export class KafkaQueue implements Queue {
         if (processedEvent) {
             const singleIngestionTimeout = timeoutGuard('After 30 seconds still ingesting event', {
                 event: JSON.stringify(processedEvent),
-                piscina: JSON.stringify(getPiscinaStats(this.piscina)),
             })
             const singleIngestionTimer = new Date()
             try {
--- a/src/main/ingestion/kafka-queue.ts
+++ b/src/main/ingestion/kafka-queue.ts
@@ -141,6 +141,12 @@ export class KafkaQueue implements Queue {
                         await this.eachBatch(payload)
                     } catch (error) {
                         status.info('💀', `Kafka batch of ${payload.batch.messages.length} events failed!`)
+                        if (error.type === 'UNKNOWN_MEMBER_ID') {
+                            status.info(
+                                '💀',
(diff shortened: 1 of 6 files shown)


mariusandra:
I went for a different approach than in the PR's title. Updated the original description.

Basically using `eachMessage` won't work, as it'll only parallelise once it gets more than one partition, meaning each 1MB of data will be processed sequentially and only the next one in parallel. This won't utilise all our cores, so swapped to a different approach.

mariusandra:
Not sure why, but seeing really poor performance locally, and some issues with the benchmark here in github.

Locally:

```
    ℹ️️ [Kafka & ClickHouse] Ingested 1000 events in 163.15s (6.13 events/sec, 163.15ms per event)
```

Going to shelf it for today.

mariusandra:
Ready for review.

This slows down ingestion just a tiny bit. Before:

```
95fc853a704d4db4e37acb7eb788ef81790de15c  Bump version to 0.16.20
    [MAIN] 09:42:15 🧩 Kafka batch of 1551 events completed in 995ms (plugins: 45ms, ingestion: 950ms)
    [MAIN] 09:42:15 🧩 Kafka batch of 1449 events completed in 825ms (plugins: 40ms, ingestion: 785ms)
```

After:

```
  console.info
    [MAIN] 10:01:18 🧩 Kafka batch of 1551 events completed in 1113ms
    [MAIN] 10:01:19 🧩 Kafka batch of 1449 events completed in 987ms
```

... but the ability to bail after every 40 ingested events and not cause cluster disruptions is one I'd sacrifice many more milliseconds for. We can even fine tune it a bit further to be a bit faster as well if need be.

src/main/ingestion/kafka-queue.ts
```diff
@@ -1,111 +1,81 @@
...
+        const timer = new Date()
+        let processedEvent: PluginEvent
+        try {
+            processedEvent = await this.processEvent(event)
+        } catch (error) {
+            status.info('🔔', error)
+            Sentry.captureException(error)
+            throw error
+        } finally {
+            this.pluginsServer.statsd?.timing('kafka_queue.single_event_batch', timer)
```
Twixes, line 66: This doesn't refer to a batch anymore, right?

mariusandra: Nope. I just kept the name to have continuity in the graphs...

Twixes (approved):

src/main/ingestion/kafka-queue.ts
Twixes: I'd rename to something different though, will be confusing for people without the history of this

mariusandra: ok, why not. removed `_batch`

PR #347, opened 2021-04-28
Do not run getPiscinaStats 3x per event

mariusandra:
## Changes

Calling `getPiscinaStats` 3x per event caused a 100x slowdown in ingestion.

from:

```
    [MAIN] 09:43:28 🧩 Kafka batch of 1551 events completed in 906ms (plugins: 49ms, ingestion: 857ms)
```

to:

```
    [MAIN] 09:40:04 🧩 Kafka batch of 1551 events completed in 90306ms (plugins: 78ms, ingestion: 90228ms)
```

The funny thing, this value would make sense when the timeoutGuard fires, not while creating the guard. So it was a bug in any case!

## Checklist

-   [ ] Jest tests

### DIFF
--- a/src/main/ingestion/kafka-queue.ts
+++ b/src/main/ingestion/kafka-queue.ts
@@ -6,7 +6,7 @@ import { PluginsServer, Queue } from 'types'
 
 import { timeoutGuard } from '../../shared/ingestion/utils'
 import { status } from '../../shared/status'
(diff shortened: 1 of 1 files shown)


mariusandra:
Actually, let's close this and go for https://github.com/PostHog/plugin-server/pull/346 directly.

PR #349, opened 2021-04-29
Kafka producer retries

mariusandra:
## Changes

- Bump Kafka producer retries from 5x to 10x
- Actually kill the server if no event ingested in 5min if the `STALENESS_RESTART_SECONDS` env is set. Should help with getting paged.
- Related: https://github.com/PostHog/plugin-server/issues/348

## Checklist

-   [ ] Jest tests

### DIFF
--- a/src/main/pluginsServer.ts
+++ b/src/main/pluginsServer.ts
@@ -188,7 +188,7 @@ export async function startPluginsServer(
                         `Plugin Server has not ingested events for over ${serverConfig.STALENESS_RESTART_SECONDS} seconds! Rebooting.`,
                         extra
                     )
-                    setTimeout(() => process.kill(process.pid, 'SIGINT'), 1000)
+                    setTimeout(() => process.kill(process.pid, 'SIGTERM'), 1000)
                     setTimeout(() => process.kill(process.pid, 'SIGTERM'), 60000)
                     setTimeout(() => process.kill(process.pid, 'SIGKILL'), 120000)
                 }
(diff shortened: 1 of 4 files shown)


src/main/pluginsServer.ts
```diff
@@ -176,17 +176,21 @@ export async function startPluginsServer(
...
                         }
                     )
                     console.log(
-                        `Plugin Server has not ingested events for over ${serverConfig.STALENESS_RESTART_SECONDS} seconds!`,
+                        `Plugin Server has not ingested events for over ${serverConfig.STALENESS_RESTART_SECONDS} seconds! Rebooting.`,
                         extra
                     )
+                    setTimeout(() => process.kill(process.pid, 'SIGINT'), 1000)
+                    setTimeout(() => process.kill(process.pid, 'SIGTERM'), 60000)
+                    setTimeout(() => process.kill(process.pid, 'SIGKILL'), 120000)
```
Twixes, line 193: We should log this shutdown to StatsD. BTW isn't SIGINT the same as SIGTERM here?

mariusandra: Sure, changed to:

```js
                    server.statsd?.increment(`alerts.stale_plugin_server_restarted`)
                    setTimeout(() => process.kill(process.pid, 'SIGTERM'), 1000)
                    setTimeout(() => process.kill(process.pid, 'SIGTERM'), 60000)
                    setTimeout(() => process.kill(process.pid, 'SIGKILL'), 120000)
```

No need to mess around, we're sending a clear signal that it's time to terminate 🕶️

mariusandra:
The only problem now is that the staleness test terminates jest :D

mariusandra:
Should be fixed. 🤞 tests...

I also fixed another leaky timeout. Might as well go in with the same PR...

mariusandra:
@Twixes mind giving it a quick review when you get the change?

Twixes (approved):
LGTM

PR #350, opened 2021-04-30
Big refactor

mariusandra:
## Changes

It's starting to be increasingly hard to work with the plugin server codebase since I can't seem to find anything the first time I look for it :)

This is the first of a few refactors that I'd like to push out to solve that problem.

This PR moves around a lot of files and changes a few imports/exports inside them. It doesn't change anything inside the files themselves... and it's just the first big move. The next refactoring PRs should be a lot less dramatic.

After the refactor, this is what the codebase will look like:

Screenshot 2021-04-30 at 09 43 49

I kept the "main" / "worker" split, yet split the "shared" folder into "config" and "utils" to better describe how they connect to the main/worker.

Going deeper:

Screenshot 2021-04-30 at 09 57 45
Screenshot 2021-04-30 at 10 00 33

The lines are still slightly blurred (src/main/job-queue-manager is called from src/utils/db/server and so on), but taking into account future plans, this is all one step in the right direction.

The... only... problem is that this PR is based on https://github.com/PostHog/plugin-server/pull/325 , as the refactor was heavily influenced by the addition of these retry queues. I'd prefer not to redo it, so would be great to merge that in, and adjust/fix after. That PR by default has all retry queues disabled, so merging it shouldn't really change anything. 😄 🤞 

## Checklist

-   [ ] Jest tests


Twixes (approved):
Makes sense to me

fuziontech:
I commented on #325 about how we should probably have two queues in the long run - one global, one local (retry queue). I think for now we can just spin up another postgres instance for this and have a unified centralized queue. Just keep in mind that eventually we'll want to split out the retry queue because if something goes down we don't want to burn up our single point of failure and have a cascade.

PR #352, opened 2021-05-03
Catch unhandled promises in workers, match those to plugin configs

mariusandra:
## Changes

- Adds a callback into workers to catch unhandled promise rejections. 
- This prevents the app from crashing by default on **NodeJS 15+** when a plugin does `new Promise(() => throw new Error('lol'))`. 
- Node 14 (and thus all docker images) were fine: the error would be just swallowed. Not ideal, but not disruptive.
- Give each anonymous function in `vm.ts` a name to be able to better follow stack traces, as line numbers are usually off.
- Adds a system to detect which plugin config an uncaught error belongs to by abusing names of "anonymous" functions in `vm.ts` (`function __bindMeta__PluginConfig_24_SECRETHASH(...)`...).
- Adds a secret hash system to prevent malicious plugins from spoofing other team IDs and potentially causing them to be punished for throwing an error.
- We will unfortunately not always have the information to detect the right plugin config, but this is at least one level of defence against accidental throws.

## Checklist

-   [ ] Jest tests


src/shared/utils.ts
```diff
@@ -506,3 +507,23 @@ export function getPiscinaStats(piscina: Piscina): Record<string, number> {
         'runTime.p50': piscina.runTime.p50,
     }
 }
+
+export function pluginConfigIdFromStack(
+    stack: string,
+    pluginConfigSecretLookup: Map<string, PluginConfigId>
+): PluginConfigId | void {
+    const regexp = /at __[a-zA-Z0-9]+__PluginConfig_([0-9]+)_([0-9a-f]+) \(vm\.js\:/
```
yakkomajuri, line 515: Could we get an example in a comment for future reference, so that we can easily know what this matches?

yakkomajuri (commented):
This is definitely going to be very useful. 

Benchmarks test is failing with the existential error, unrelated to this I believe.

src/shared/utils.ts
mariusandra: Clarified it a bit.

The benchmarks test is indeed annoyingly flakey. It'll bother me enough to look at it soon enough :).

yakkomajuri: great, thanks!

mariusandra:
I'm going to merge this. The test that's broken will be fixed by https://github.com/PostHog/plugin-server/pull/351

PR #357, opened 2021-05-05
Graphile worker connection handling

### LINKED ISSUE #355: Graphile queue stalls the app if it can't connect
We need to add some error handling. Right now if the DB for graphile is misconfigured, it just stalls the app forever.

Because some plugin servers will definitely need a connected job queue before ingestion can start, I'll also add an env like `CRASH_IF_NO_PERSISTENT_JOB_QUEUE=0` (off by default, for now).

mariusandra:
## Changes

- This adds some error handling to the graphile job queue. Before everything would just freeze to a halt on a connection error. Now the server either skips this queue and starts, or crashes, depending on an env.
- Adds a `CRASH_IF_NO_PERSISTENT_JOB_QUEUE` env. Defaults to false (no crash) for now. This should help debug the connection issue with aurora for cloud. We might remove it later and always require at least one persistent queue. Normally we just use the main postgres db as the graphile db, and that should work... Unless you can't create tables.
- Closes #355

## Checklist

-   [ ] Jest tests


mariusandra:
Would you imagine that everything works locally 😅

mariusandra:
@Twixes Excluding any last broken tests (will fix ASAP), this PR is ready to be reviewed. I'm not floored with the JobQueue interface as it is now. I think it's mixing way too many concerns and is not handling failures (and fallbacks) well enough. So there will be plenty of work and refactoring still to come.

In the mean while, this will at least make the graphile worker somewhat usable... and at least provide legible errors if it can't connect.

I'd like to get it in cloud asap to finally hook up the Aurora db. :)

Twixes (approved):
All good for me

PR #359, opened 2021-05-06
onEvent & onSnapshot

### LINKED ISSUE #280: Don't process $snapshot via plugins

### LINKED ISSUE #328: Add `onEvent`
There's a case to be made for splitting `processEvent` into `processEvent` and `onEvent`. I'm still not 100% convinced introducing a larger API surface is worth it, but the semantics around retries makes it rather appealing.

We would make this distinction:

- Plugins which use `processEvent` modify events and must run in order. For example: adding GeoIP information.
- Plugins which use `onEvent` do not modify the event. They export the processed event to another service or do some other async background processing. For example batching and uploading to S3. All `onEvent` functions can run in parallel, no sequential ordering needed. We can also run them in parallel with the actual event ingestion code.

This naming scheme `onEvent` could fit with other "event types" like `onAction`, `onSnapshot` (you don't `processAction`). 

Retries are different in both cases. With `processEvent`, a retry would halt the entire event processing pipeline, and delay event ingestion until a retry passes. With `onEvent`, a retry would just retry that async background job without ceremony. The event is already ingested.

Hence we might event code retry support for just `onEvent` for simplicity and leave `processEvent` out of it.

mariusandra:
## Changes

- Adds `onSnapshot`  and `onEvent` that get run for snapshots and everything else respectively
- Removes snapshots from processEvent
- Closes #328 
- Closes #280 

## Checklist

-   [ ] Jest tests

### DIFF
--- a/src/main/ingestion-queues/kafka-queue.ts
+++ b/src/main/ingestion-queues/kafka-queue.ts
@@ -2,7 +2,7 @@ import Piscina from '@posthog/piscina'
 import { PluginEvent } from '@posthog/plugin-scaffold'
 import * as Sentry from '@sentry/node'
 import { Consumer, EachBatchPayload, EachMessagePayload, Kafka, KafkaMessage } from 'kafkajs'
-import { PluginsServer, Queue } from 'types'
+import { PluginsServer, Queue, WorkerMethods } from 'types'
 
 import { timeoutGuard } from '../../utils/db/utils'
 import { status } from '../../utils/status'
@@ -14,26 +14,19 @@ export class KafkaQueue implements Queue {
     private kafka: Kafka
     private consumer: Consumer
     private wasConsumerRan: boolean
-    private processEvent: (event: PluginEvent) => Promise<PluginEvent>
-    private ingestEvent: (event: PluginEvent) => Promise<void>
+    private workerMethods: WorkerMethods
 
     // used for logging aggregate stats to the console
     private messageLogDate = 0
     private messageCounter = 0
 
-    constructor(
-        pluginsServer: PluginsServer,
-        piscina: Piscina,
-        processEvent: (event: PluginEvent) => Promise<any>,
-        ingestEvent: (event: PluginEvent) => Promise<void>
-    ) {
+    constructor(pluginsServer: PluginsServer, piscina: Piscina, workerMethods: WorkerMethods) {
         this.pluginsServer = pluginsServer
         this.piscina = piscina
         this.kafka = pluginsServer.kafka!
         this.consumer = KafkaQueue.buildConsumer(this.kafka)
         this.wasConsumerRan = false
-        this.processEvent = processEvent
-        this.ingestEvent = ingestEvent
+        this.workerMethods = workerMethods
         this.messageLogDate = new Date().valueOf()
     }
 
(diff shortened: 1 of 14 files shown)


src/main/ingestion-queues/kafka-queue.ts
```diff
@@ -51,39 +44,40 @@ export class KafkaQueue implements Queue {
...
+                          func: (event) => this.workerMethods.onSnapshot(event),
+                          statsKey: 'kafka_queue.single_on_snapshot',
+                          timeoutMessage: 'After 30 seconds still running onSnapshot',
+                      })
+                    : this.runInstrumentedFunction({
+                          event: processedEvent,
+                          func: (event) => this.workerMethods.onEvent(event),
+                          statsKey: 'kafka_queue.single_on_event',
+                          timeoutMessage: 'After 30 seconds still running onEvent',
+                      }),
```
Twixes, line 79: Nit: I think it'd be a bit more readable to do the ternary operaration on arg _in_ the `this.runInstrumentedFunction(arg)` call.

src/main/ingestion-queues/kafka-queue.ts
```diff
@@ -51,39 +44,40 @@ export class KafkaQueue implements Queue {
...
-        } catch (error) {
-            status.info('🔔', error)
-            Sentry.captureException(error)
-            throw error
-        } finally {
-            this.pluginsServer.statsd?.timing('kafka_queue.single_event', timer)
-            clearTimeout(processingTimeout)
+        let processedEvent: PluginEvent | null = event
+
+        if (processedEvent.event !== '$snapshot') {
```
Twixes, line 49: Absolute nit: since this is used in two places, the logic would be nicer to follow if the condition was a bool const `isSnapshot`

src/worker/vm/vm.ts
```diff
@@ -152,6 +152,8 @@ export async function createPluginConfigVM(
             const __methods = {
                 setupPlugin: __asyncFunctionGuard(__bindMeta('setupPlugin')),
                 teardownPlugin: __asyncFunctionGuard(__bindMeta('teardownPlugin')),
+                onEvent: __asyncFunctionGuard(__bindMeta('onEvent')),
+                onSnapshot: __asyncFunctionGuard(__bindMeta('onSnapshot')),
```
Twixes, line 156: `onEvent` conveys the read-only nature of this feature well, but maybe not that this is ran at the very end of processing, when the even has been processed by all plugins (as opposed to a raw event from the endpoint for instance). So maybe something like `onCompletedEvent`, `onFinishedEvent`, `onSavedEvent`?

src/main/ingestion-queues/kafka-queue.ts
mariusandra: Indeed it was! I got carried away and refactored this ingestion dance into a separate `ingest-event.ts` simple [POJSF](https://gist.github.com/klaaspieter/12cd68f54bb71a3940eae5cdd4ea1764) as well, and shared it between Celery and Kafka. [Instrumentation must be free!](https://www.openbsd.org/lyrics.html#35)

src/main/ingestion-queues/kafka-queue.ts
mariusandra: agreed

src/worker/vm/vm.ts
mariusandra: I'd really like to fight entropy here again :). If we add some modifier, then it's another thing for people to remember, and  the entire `on` namespace will get annoyingly repetitive. `onCreatedEvent`, `onCreatedAction`, etc.

People who will know that `onEvent` happens after `processEvent` because in the interface we'll have a visual representation of `processEvent` plugins that happen before `onEvent` plugins that can't be reordered are at the end.

src/main/ingestion-queues/kafka-queue.ts
Twixes: Nice, good refactor

src/worker/vm/vm.ts
Twixes: The UI is not an issue, but this may be unclear from an API perspective for plugin developers. But I don't have a strong opinion on this, so if you definitely see plain `onEvent` as the best option, then let's go for that.

mariusandra: Let's just document this well :).

mariusandra: With the UI, I was getting at the idea that if you're a plugin user, you will see from the UI that there are some plugins that happen in order and some that happen after them. This would lead to some understanding of how to write plugins. However, thinking further, indeed the users and developers won't be the same people (see rudderstack export), so scratch that reasoning. 🤷

PR #361, opened 2021-05-07
Enable Graphile as the default job queue

mariusandra:
## Changes

- Since not that long ago we added job queues to the plugin server. One of them is with a tool called "[graphile/worker](https://github.com/graphile/worker)", which is a postgres-based queue. 
- I'd like to enable this queue for everyone by default in 1.25.0, so there's at least some fallback for broken events. There's also the env `JOB_QUEUE_GRAPHILE_URL` for specifying a custom database for the job queue, like we have with Aurora. If that env is blank, we use `DATABASE_URL`.
- The only issue is that graphile does its migrations after connecting. This means, everyone who's using posthog, should have their database credentials set in such a way to allow the plugin server to make migrations on their `DATABASE_URL`.

Do you think this is fine? Just in case, I added small script and now if you run the plugin server with `--migrate`, so currently as simple as `cd plugins && yarn start --migrate`, it'll apply the graphile migrations in case it's a configured job queue.

Should we add this in with the other migrations somehow? The plugin server might get more migrations eventually (not sure what and when), though I can always make it work by running migration on boot. Except if there's a connection/permission error.

Also, if graphile can't connect to the database, it might stall the plugin server a bit (timeouts, ...), but it'll then bail and start normally. That's because `CRASH_IF_NO_PERSISTENT_JOB_QUEUE` is still set to "no crash" by default.

## Checklist

-   [ ] Jest tests

### DIFF
--- a/benchmarks/vm/memory.benchmark.ts
+++ b/benchmarks/vm/memory.benchmark.ts
@@ -32,6 +32,7 @@ const mockPlugin: Plugin = {
     archive: null,
     error: undefined,
     is_global: false,
+    is_preinstalled: false,
     created_at: new Date().toISOString(),
     updated_at: new Date().toISOString(),
 }
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -61,6 +61,7 @@ export function getDefaultConfig(): PluginsServerConfig {
         PLUGIN_SERVER_IDLE: false,
         JOB_QUEUES: 'graphile',
         JOB_QUEUE_GRAPHILE_URL: '',
+        JOB_QUEUE_GRAPHILE_SCHEMA: 'graphile_worker',
         CRASH_IF_NO_PERSISTENT_JOB_QUEUE: false,
         STALENESS_RESTART_SECONDS: 0,
     }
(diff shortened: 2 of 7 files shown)


macobo:
> The only issue is that graphile does its migrations after connecting. This means, everyone who's using posthog, should have their database credentials set in such a way to allow the plugin server to make migrations on their DATABASE_URL.

Couple of questions to clarify: 
1) What happens if two different plugin servers try to migrate at the same time
2) Do you foresee needing to do data migrations (e.g. updating existing data rather than just adding to the schema)

mariusandra:
1. I'm not 100% sure. Graphile seems pretty sturdy though, and made in a way that seems to be used in a "set it and forget it" fashion. Worse case, the plugin server will fail during first installation and then will restart. 
2. I think not, but if so, I assume the guy behind Graphile will have made sure it doesn't cause downtime. It's used in a lot of projects.

macobo:
>The only issue is that graphile does its migrations after connecting. This means, everyone who's using posthog, should have their database credentials set in such a way to allow the plugin server to make migrations on their DATABASE_URL.

AFAIK this is a safe assumption to make. When calling migrations in any deployments we support we pass in the same credentials as to web so things should work.

As for whether migrating-at-start is OK: the short answer is ¯\_(ツ)_/¯ 

Slightly longer: I think it depends on how things behave when e.g. you start 5 plugin servers at the same time and each of them tries to run migrations. Based on what you're saying above, I get that it's behaving well for now which means it should be okay to roll out as-is. 

Main things you want to avoid is that it doesn't get stuck in e.g. a state where the migrations can't be applied due to conflicting changes and duplicate data migrations. If we start seeing the first problem OR need to do data migrations we should move migrations to happen in the same place django migrations currently occur (which are guaranteed to happen in one container at a time).

mariusandra:
Yeah, I'm quite convinced the migrations will run OK. They have been going well so far on cloud.

There's actually one other point to consider. I've been asking this before, but never got a direct answer. Graphile uses a different schema to store its data: `graphile_worker` by default, not `public` as posthog. We might get cases, where the postgres user is only able to modify tables in the `public` schema. We fail gracefully then, so it should be OK this time around. Just the queue won't be started and some plugins might fail.

Graphile creates 4 tables: `job_queues`, `jobs`, `known_crontabs` and `migrations`. With names as generic as that, it's wise to keep them in a different schema. However I've exposed a separate `JOB_QUEUE_GRAPHILE_SCHEMA` env to set this to `public` if someone really needs to.

PR #368, opened 2021-05-07
Plugin Running Duration Metrics

yakkomajuri:
## Changes

Supersedes #324

## Checklist

-   [ ] Jest tests


Twixes:
What's the state of this? Some tests still red 😳

yakkomajuri:
They don't look related...

src/utils/posthog.ts
```diff
@@ -0,0 +1,11 @@
+import * as fetch from 'node-fetch'
+import { nodePostHog } from 'posthog-js-lite/dist/src/targets/node'
+
+export const posthog = nodePostHog('sTMFPsFhdP1Ssg', {
```
neilkakkar, line 4: should this be coming from a config / env var?

yakkomajuri: It shouldn't - I originally had it as such but Marius had a good point that people deploying the plugin server shouldn't be able to configure this (as it sends metrics to us). 

It also is a public token anyway

yakkomajuri:
Let's get this in @neilkakkar @Twixes

mariusandra:
Now that https://github.com/PostHog/plugin-server/pull/432 is in, is there something we should alter here? Somehow both features are similar in spirit.

yakkomajuri:
Are they though? #432 aims to provide self-hosted users metrics from their instance to their instance, whereas this sends metrics from everyone to us, some data that we were discussing might even underpin a billing strategy in the future.

Could be wrong in not seeing a deeper link. Also happy to consolidate approaches though if you feel they make sense.

yakkomajuri:
@mariusandra @neilkakkar

neilkakkar (approved):
LGTM. I'd request one more test that integrates with `runProcessEvent` or the like (which ensures not just that capture works fine, but works fine for the `run*` functions )

yakkomajuri:
I think your point is valid @neilkakkar 

Merging though as I don't have time today and this has been lingering for so long. We'll get feedback rather soon about how it works live :D

mariusandra:
Fine for me to merge :). Regarding this and the instance status report, I think it would also be valuable to eventually see this data in the instance status report. However not urgent.

neilkakkar:
Perfectly okay with doing the tests in a separate PR, but I'd really like to have them in some time, even if things work fine for now 😉 . Since, that's how we know if we're not regressing this functionality when future changes come in.

PR #373, opened 2021-05-10
Add a few metrics

### LINKED ISSUE #354: Metrics for JS event bus
Under https://github.com/PostHog/plugin-server/issues/264 Marius wrote:

>because there are too many other unresolved promises jamming the JS event bus, but the single threaded event loop is stuck behind something sync and stupid, like running getPiscinaStats 3x per event.

If this is a factor in under-load operations of plugin-server, we should also have internal metrics for it to point to when there are issues.

### LINKED ISSUE #366: Add metrics for plugin logs to grafana
Let's get some graphs for log entries.

Twixes:
## Changes

Metrics for plugin log entry creation (resolves #366), for event loop lag (not sure how effective this will be, but resolves #354), also will allow fixing below graph.

No data

### DIFF
--- a/src/utils/db/db.ts
+++ b/src/utils/db/db.ts
@@ -642,6 +642,12 @@ export class DB {
             console.error(e)
         }
 
+        this.statsd?.increment(`logs.entries_created`, {
+            source,
+            team_id: pluginConfig.team_id.toString(),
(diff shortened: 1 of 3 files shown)


mariusandra (approved):
Makes sense

Issue #279, opened 2021-03-24
Disable plugins after they fail to load enough times / Retry loading plugins?


mariusandra:
This my dev EE console every time I start it:

```
[MAIN] ✅ Kafka consumer connected!
[__14] 🔌 Loaded plugin "MaxMind GeoIP" (organization 01778ae6-967a-0000-449a-270bb4205419)
[MAIN] ✅ Kafka consumer joined group clickhouse-ingestion!
[MAIN] ✅ Connected to Redis!
[MAIN] 🚀 All systems go.
[__15] 🔌 Loaded plugin "MaxMind GeoIP" (organization 01778ae6-967a-0000-449a-270bb4205419)
[__16] 🔌 Loaded plugin "MaxMind GeoIP" (organization 01778ae6-967a-0000-449a-270bb4205419)
[__15] ⚠️ Failed to load plugin "Gitlab aasd" (organization 01778ae6-967a-0000-449a-270bb4205419)
[__16] ⚠️ Failed to load plugin "Gitlab aasd" (organization 01778ae6-967a-0000-449a-270bb4205419)
[__14] ⚠️ Failed to load plugin "Gitlab aasd" (organization 01778ae6-967a-0000-449a-270bb4205419)
[__14] ⚠️ Failed to load plugin "GitLab Release Tracker" (organization 01778ae6-967a-0000-449a-270bb4205419)
[__15] ⚠️ Failed to load plugin "GitLab Release Tracker" (organization 01778ae6-967a-0000-449a-270bb4205419)
[__16] ⚠️ Failed to load plugin "GitLab Release Tracker" (organization 01778ae6-967a-0000-449a-270bb4205419)
```

Right now if a plugin fails during initialisation, we'll just try again next time the server restarts. It might make sense to punish plugins that fail on boot by disabling them. Perhaps with some retry strategy?

Somewhat related to #273 and #72

Does not seem urgent right now, just documenting.

mariusandra:
Here's an example of a plugin that fails in a different way: https://sentry.io/organizations/posthog/issues/2375532863/?project=5592816&query=is%3Aunresolved&statsPeriod=14d

After a few of these the plugin should somehow be taken out of the pipeline.

yakkomajuri:
#380 fixed this

PR #380, opened 2021-05-11
Punish Plugins >:D

yakkomajuri:
## Changes

Disables plugins that errored during setup. Addresses #279 

This is somewhat a WIP from the perspective of approach. It works, but isn't particularly sophisticated.

I guess I confused you @mariusandra on Slack because I think you had this simple approach in mind whereas I was thinking of even blocking users from re-enabling plugins we "know" are broken. Hence I was talking about hashing the content and all.

Nevertheless, this is already an improvement, and I'm putting it out to spark discussion on a few points (PRs over issues):

- Do we want a retry strategy for `setupPlugin`? How many retries?
- Do we want to allow users to re-enable plugins that failed on setup even though they haven't changed at all? i.e. prevent a user from keeping on enabling a plugin that we know will fail.

## Checklist

-   [ ] Jest tests

### DIFF
--- a/src/utils/db/sql.ts
+++ b/src/utils/db/sql.ts
@@ -3,10 +3,12 @@ import {
     PluginAttachmentDB,
     PluginConfig,
     PluginError,
+    PluginId,
     PluginLogEntrySource,
     PluginLogEntryType,
     PluginsServer,
 } from '../../types'
+import { TeamId } from './../../types'
 
 function pluginConfigsInForceQuery(specificField?: keyof PluginConfig): string {
     return `SELECT posthog_pluginconfig.${specificField || '*'}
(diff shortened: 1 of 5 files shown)


yakkomajuri:
Green :O

neilkakkar (approved):
LGTM.

About the retry strategies for setup: I'm guessing that's for plugins that do something over the network during setup - which might fail on a network partition and thus get disabled. 

I think solving this while keeping the code as simple as possible is a big win. So, I'd actually vote for no retries on setup. This is because:

1. We can't be sure if the error on setup is due to a user input error, or an actual connectivity issue.
2. In either case, since the plugin user needs to check their settings, and test upstream connectivity, disabling on one try and letting them re-enable when they're convinced everything is working is good enough.

which kinda answers the 2nd question as well: yes, let people re-enable plugins, because we don't actually *know* if the same plugin code would run this time or not?

Thoughts? Am I making a wrong assumption here?

yakkomajuri:
Well so I agree with keeping it simple. But there would be a way (a convoluted one though), to check the same plugin is running.

Plugins that are downloaded have a `latest_tag` which specifies the version of the plugin. We'd also need to check that config  values have changed, because those can affect it too.

And for plugins that are user-built, we could hash the config schema and plugin code to determine if it has changed. Definitely don't think it's necessary, but I guess it just came to mind when I saw this issue.

yakkomajuri:
Sorry for the accidental minor bump :(

PR #384, opened 2021-05-17
Plugin Capabilities implementation

### LINKED ISSUE #379: Plugin Capabilities
Splitting this from #165

> We need to cache what each plugin supports. Basically if it has support for 1) scheduled functions runEveryX 2) event processing functions processEvent / processEventBatch. This lets us separate VMs better in the future. For example having a separate pool of piscina workers for scheduled tasks, starting there only the VMs that are relevant. This caching could be run when a plugin is installed.

It would help with distributing and optimising worker pools inside the plugin server, if we knew what each plugin can do. Basically, what it exports. This knowledge can be used to help make a nicer interface (no need to reorder export plugins that run at the same time) and it'll help us split processEvent/onEvent tasks (which need to finish very fast) from the rest (scheduled jobs and other async tasks)

neilkakkar:
## Changes

Adds support for Plugins Capabilities. Resolves #379 

Companion PostHog/posthog PR (https://github.com/PostHog/posthog/pull/4371) to be merged first.

~~Will write tests now, but just wanted to get a PoC out.~~

## Checklist

-   [x] Jest tests

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -5,6 +5,7 @@ import { StatsD } from 'hot-shots'
 import { Redis } from 'ioredis'
 import { Kafka } from 'kafkajs'
 import { DateTime } from 'luxon'
+import { JobQueueManager } from 'main/job-queues/job-queue-manager'
 import { Pool } from 'pg'
 import { VM } from 'vm2'
 
@@ -100,7 +101,7 @@ export interface PluginsServer extends PluginsServerConfig {
     pluginConfigSecretLookup: Map<string, PluginConfigId>
     // tools
     eventsProcessor: EventsProcessor
-    jobQueueManager: JobQueue
+    jobQueueManager: JobQueueManager
     // diagnostics
     lastActivity: number
     lastActivityType: string
--- a/src/types.ts
+++ b/src/types.ts
@@ -80,7 +80,7 @@ export interface PluginsServerConfig extends Record<string, any> {
     STALENESS_RESTART_SECONDS: number
 }
 
-export interface PluginsServer extends PluginsServerConfig {
+export interface Hub extends PluginsServerConfig {
     instanceId: UUID
     // active connections to Postgres, Redis, ClickHouse, Kafka, StatsD
     db: DB
--- a/src/types.ts
+++ b/src/types.ts
@@ -161,7 +161,13 @@ export interface Plugin {
     from_web?: boolean
     created_at: string
     updated_at: string
-    capabilities?: string[]
+    capabilities?: Capabilities
+}
+
+export interface Capabilities {
+    jobs: string[]
+    scheduled_tasks: string[]
+    methods: string[]
 }
 
 export interface PluginConfig {
--- a/src/types.ts
+++ b/src/types.ts
@@ -165,9 +165,9 @@ export interface Plugin {
 }
 
 export interface Capabilities {
-    jobs: string[]
-    scheduled_tasks: string[]
-    methods: string[]
+    jobs?: string[]
+    scheduled_tasks?: string[]
+    methods?: string[]
 }
 
 export interface PluginConfig {
--- a/src/types.ts
+++ b/src/types.ts
@@ -261,13 +261,13 @@ export type WorkerMethods = {
 export interface PluginConfigVMReponse {
     vm: VM
     methods: {
-        setupPlugin: (() => Promise<void>) | undefined
-        teardownPlugin: (() => Promise<void>) | undefined
-        onEvent: ((event: PluginEvent) => Promise<void>) | undefined
-        onSnapshot: ((event: PluginEvent) => Promise<void>) | undefined
-        processEvent: ((event: PluginEvent) => Promise<PluginEvent>) | undefined
+        setupPlugin?: () => Promise<void>
+        teardownPlugin?: () => Promise<void>
+        onEvent?: (event: PluginEvent) => Promise<void>
+        onSnapshot?: (event: PluginEvent) => Promise<void>
+        processEvent?: (event: PluginEvent) => Promise<PluginEvent>
         // DEPRECATED
-        processEventBatch: ((batch: PluginEvent[]) => Promise<PluginEvent[]>) | undefined
+        processEventBatch?: (batch: PluginEvent[]) => Promise<PluginEvent[]>
     }
     tasks: Record<PluginTaskType, Record<string, PluginTask>>
 }
--- a/src/types.ts
+++ b/src/types.ts
@@ -402,6 +402,7 @@ export interface ClickHousePerson {
     team_id: number
     properties: string
     is_identified: number
+    is_deleted: number
     timestamp: string
 }
 
--- a/src/types.ts
+++ b/src/types.ts
@@ -253,13 +253,13 @@ export type WorkerMethods = {
 export interface PluginConfigVMReponse {
     vm: VM
     methods: {
-        setupPlugin: () => Promise<void>
-        teardownPlugin: () => Promise<void>
-        onEvent: (event: PluginEvent) => Promise<void>
-        onSnapshot: (event: PluginEvent) => Promise<void>
-        processEvent: (event: PluginEvent) => Promise<PluginEvent>
+        setupPlugin: () => Promise<void> | undefined
+        teardownPlugin: () => Promise<void> | undefined
+        onEvent: (event: PluginEvent) => Promise<void> | undefined
+        onSnapshot: (event: PluginEvent) => Promise<void> | undefined
+        processEvent: (event: PluginEvent) => Promise<PluginEvent> | undefined
         // DEPRECATED
-        processEventBatch: (batch: PluginEvent[]) => Promise<PluginEvent[]>
+        processEventBatch: (batch: PluginEvent[]) => Promise<PluginEvent[]> | undefined
     }
     tasks: Record<PluginTaskType, Record<string, PluginTask>>
 }
--- a/src/types.ts
+++ b/src/types.ts
@@ -161,10 +161,10 @@ export interface Plugin {
     from_web?: boolean
     created_at: string
     updated_at: string
-    capabilities?: Capabilities
+    capabilities?: PluginCapabilities
 }
 
-export interface Capabilities {
+export interface PluginCapabilities {
     jobs?: string[]
     scheduled_tasks?: string[]
     methods?: string[]
--- a/src/types.ts
+++ b/src/types.ts
@@ -161,6 +161,7 @@ export interface Plugin {
     from_web?: boolean
     created_at: string
     updated_at: string
+    capabilities?: string[]
 }
 
 export interface PluginConfig {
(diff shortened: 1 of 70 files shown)


mariusandra:
Having given it a quick glance, I think we should store a bit more metadata with the capabilities, mainly adding the type of the capability. Basically storing this in a JSON like: `{ jobs: ['runSomthing'], tasks: ['processEvent', 'setupPlugin'] }` or something like that.

tests/helpers/plugins.ts
```diff
@@ -48,7 +48,7 @@ export const plugin60: Plugin = {
     is_preinstalled: false,
     created_at: new Date().toISOString(),
     updated_at: new Date().toISOString(),
-    capabilities: {},
+    capabilities: { jobs: [], scheduled_tasks: [], methods: [] }, // inferred on setup
```
mariusandra, line 51: Maybe we should keep this here, or somewhere, as the `{}` default that is set in postgres? To test that `jobs` and `methods` might not be set on the object.

src/utils/db/sql.ts
```diff
@@ -51,6 +52,25 @@ export async function getPluginConfigRows(server: PluginsServer): Promise<Plugin
...
+        [capabilities, pluginConfig.plugin_id],
+        'setPluginCapabilities'
+    )
+    await server.db.createPluginLogEntry(
+        pluginConfig,
+        PluginLogEntrySource.System,
+        PluginLogEntryType.Info,
+        `Set plugin capabilities (instance ID ${server.instanceId}).`,
+        server.instanceId
+    )
```
mariusandra, line 71: I think if the capabilities haven't changed, there's no need to update them nor log that we set them. That would be needless logspam :)

src/worker/plugins/loadPlugin.ts
```diff
@@ -94,3 +98,40 @@ export async function loadPlugin(server: PluginsServer, pluginConfig: PluginConf
     }
     return false
 }
+
+async function inferPluginCapabilities(server: PluginsServer, pluginConfig: PluginConfig) {
+    // infer on load implies there's no lazy loading, but all workers get
+    // these properties loaded
```
mariusandra, line 104: We could lazy load them. We won't do anything with the loaded capabilities for the moment, and eventually set them in a separate "install" step anyway. No need to block for that.

mariusandra (commented):
Very cool! Few inline comments plus probably could use a merge from master.

src/worker/plugins/loadPlugin.ts
```diff
@@ -94,3 +102,51 @@ export async function loadPlugin(server: PluginsServer, pluginConfig: PluginConf
     }
     return false
 }
+
+async function inferPluginCapabilities(server: PluginsServer, pluginConfig: PluginConfig, prevConfig?: PluginConfig) {
+    // infer on load implies there's no lazy loading, but all workers get
+    // these properties loaded
+    const vm = await pluginConfig.vm?.resolveInternalVm
+    const capabilities: Capabilities = { scheduled_tasks: [], jobs: [], methods: [] }
```
mariusandra, line 110: Tip: we should normally avoid using `!` in real (non-test) code. It's a place for subtle bugs, and better to just avoid. Often times, the solution is as simple as replacing `!` with `?`. Here though I'd suggest casting this object to `Required<Capabilities>` instead of just `Capabilities`. Then all the `methods`, `jobs`, etc will be there.

src/worker/plugins/loadPlugin.ts
```diff
@@ -94,3 +102,51 @@ export async function loadPlugin(server: PluginsServer, pluginConfig: PluginConf
...
+        capabilities.jobs?.sort().toString() == prevCapabilities.jobs?.sort().toString() &&
+        capabilities.scheduled_tasks?.sort().toString() == prevCapabilities.scheduled_tasks?.sort().toString() &&
+        capabilities.methods?.sort().toString() == prevCapabilities.methods?.sort().toString()
+    ) {
+        // pass - no change in capabilities
+    } else {
+        await setPluginCapabilities(server, pluginConfig, capabilities)
+    }
+
+    pluginConfig.plugin!.capabilities = capabilities
```
mariusandra, line 151: Here again the `!` could cause subtle issues. After all, we're not sure this function will always be called with pluginConfigs that contain a plugin. It might be the case when we follow the current trail that leads to this function, but it might not be the case in the future.

Often the solution is to use a guard, such as `if (!pluginConfig.plugin) { throw new Error('how the hell did you get here?') }`, or just `return`, depends on the case. Guards are best placed at the top of a function as well.

src/worker/plugins/run.ts
```diff
@@ -58,7 +58,7 @@ export async function runProcessEvent(server: PluginsServer, event: PluginEvent)
 
             try {
                 returnedEvent = (await processEvent(returnedEvent)) || null
-                if (returnedEvent.team_id != teamId) {
+                if (returnedEvent?.team_id != teamId) {
```
mariusandra, line 61: First, in JS we should always use the `!==` and `===` operators instead of `!=` and `==`. The shorter ones equate all falsy fields. Basically, `"" == false == 0`, but `"" !== false !== 0`. 

Then, if `returnedEvent` is null (plugin skipped it) and hence its `team_id` is undefined, this will throw the error.

mariusandra (commented):
Few more flyby comments and suggestions :)

src/types.ts
```diff
@@ -255,13 +261,13 @@ export type WorkerMethods = {
...
-        setupPlugin: () => Promise<void>
-        teardownPlugin: () => Promise<void>
-        onEvent: (event: PluginEvent) => Promise<void>
-        onSnapshot: (event: PluginEvent) => Promise<void>
-        processEvent: (event: PluginEvent) => Promise<PluginEvent>
+        setupPlugin: (() => Promise<void>) | undefined
+        teardownPlugin: (() => Promise<void>) | undefined
+        onEvent: ((event: PluginEvent) => Promise<void>) | undefined
+        onSnapshot: ((event: PluginEvent) => Promise<void>) | undefined
+        processEvent: ((event: PluginEvent) => Promise<PluginEvent>) | undefined
```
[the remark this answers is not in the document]
neilkakkar: much better, thanks!

src/worker/plugins/loadPlugin.ts
neilkakkar: Hmm, I could unblock them, but that still doesn't mean the VMs are lazy loaded right? since `inferPluginCapabilities()` will load the VM on startup, before the worker gets any event.

Maybe I should just remove this comment, just realised it's confusing w.r.t what is being lazy loaded.

Twixes (changes requested):
Nice! Just some minor comments

tests/plugins.test.ts
```diff
@@ -473,6 +483,120 @@ test('plugin config order', async () => {
...
+
+    await setupPlugins(mockServer)
+    const { pluginConfigs } = mockServer
+
+    const pluginConfig = pluginConfigs.get(39)!
+
+    expect(pluginConfig.plugin!.capabilities!.methods!.sort()).toEqual([
+        'onSnapshot',
+        'processEvent',
+        'processEventBatch',
```
Twixes, line 576: It seems a bit quirky that `processEventBatch` is listed as a capability when it's only an autogenerated version based on actually provided `processEvent`, but that seems OK

Twixes (approved):
Stamp of approval

src/worker/plugins/loadPlugin.ts
```diff
@@ -94,3 +103,53 @@ export async function loadPlugin(server: PluginsServer, pluginConfig: PluginConf
...
+    const vm = await pluginConfig.vm?.resolveInternalVm
+    const capabilities: Required<PluginCapabilities> = { scheduled_tasks: [], jobs: [], methods: [] }
+
+    const tasks = vm?.tasks
+    const methods = vm?.methods
+
+    if (methods) {
+        for (const [key, value] of Object.entries(methods)) {
+            if (value) {
+                capabilities.methods!.push(key)
```
mariusandra, line 128: nit: `!`

src/worker/plugins/loadPlugin.ts
mariusandra: The funny code here is this:

```ts
void pluginConfig.vm?.initialize!(server, pluginConfig, plugin.source, pluginDigest(plugin))
await inferPluginCapabilities(server, pluginConfig, prevConfig)
```

The first function is run in the background, the second one we await for. The second function awaits for a promise (`resolveInternalVm`) that gets set when the first function that's running in the background finishes.

In essence, the plugin is not lazy anymore, but we block until the VM setup has completed.

To get around it, we could just `void` the `inferPluginCapabilities` call and let them run in the background, however that also feels slightly hacky. 

A possibly better idea is to put this `inferPluginCapabilities` inside the `LazyVM.initialize` function. There you probably won't need `prevConfig` either. Just check if the current `pluginConfig.plugin.capabilities` matches reality or not. If not, update.

tests/plugins.test.ts
```diff
@@ -473,6 +483,120 @@ test('plugin config order', async () => {
...
+    await setupPlugins(mockServer)
+    const pluginConfig = mockServer.pluginConfigs.get(39)!
+
+    pluginConfig.updated_at = new Date().toISOString()
+    // config is changed, but capabilities haven't changed
+
+    await setupPlugins(mockServer)
+    const newPluginConfig = mockServer.pluginConfigs.get(39)!
+
+    expect(newPluginConfig.plugin!.capabilities).toBe(pluginConfig.plugin!.capabilities)
```
mariusandra, line 596: Let's also check that:

```ts
    expect(newPluginConfig.plugin).not.toBe(pluginConfig.plugin)
```

otherwise the capabilities are the same because it's literally the same object :)

mariusandra (changes requested):
Still some inline comments to check :)

src/worker/plugins/loadPlugin.ts
neilkakkar: that.. makes a lot of sense. Ha, would've been nice if I made this leap myself when thinking about the implementation.

Come to think of it, I don't need the `prevConfig` in the current setup either, for similar reasons - I already have a `pluginConfig.plugin.capabilities` to check.

tests/plugins.test.ts
```diff
@@ -38,7 +39,7 @@ afterEach(async () => {
 })
 
 test('setupPlugins and runProcessEvent', async () => {
-    getPluginRows.mockReturnValueOnce([plugin60])
+    getPluginRows.mockReturnValueOnce([{ ...plugin60 }])
```
neilkakkar, line 42: it took me more than an hour on Friday night to figure out *this* was the issue with tests! :/ (since it affected plugin60 directly, and wouldn't show up if I simply ran the tests in isolation) - only when things run in sequence did this bork things up. 🤦 .

mariusandra: Ooh... that's a nice rabbit hole :D

mariusandra (approved):
Looks good to me!

PR #392, opened 2021-05-18
Add node-postgres to supported libraries

yakkomajuri:
## Changes

_Please describe._

I'll be using this for the Redshift and subsequently Postgres plugins

## Checklist

-   [ ] Jest tests


mariusandra:
Code wise this should be fine, but there's another issue, security. More context in this thread: https://github.com/PostHog/plugin-repository/issues/11

yakkomajuri:
About to jump on my 5h of calls so just quickly mocked up this WIP at approach 2 outlined here #395 to spark a discussion and move forward iteratively with Redshift at least.

Curious to hear thoughts @mariusandra

mariusandra:
Yikes, this approach is both amazing and terrifying :D.

There's going to be a lot of maintenance overhead if we do this. Instead some simple wrapper that injects parameters would be good enough. However, then there's no guarantee the someone can't use some internal and unexposed "changeConnectionUrl" function... ES Proxies might help or this might be a losing battle.

Also, just replied here: https://github.com/PostHog/plugin-server/issues/395#issuecomment-844401356

mariusandra (approved):

yakkomajuri:
Oh! I was just writing some wrappers

mariusandra:
Whoops :)

PR #393, opened 2021-05-18
Delete person using insertion

EDsCODE:
## Changes

_Please describe._
- as a result of this [pairing session](https://github.com/PostHog/product-internal/issues/47), we're looking to use a new `deleted` field to track deleted users (similar to how deletion everywhere else behind django is managed)
- queries will account for this and filter out deleted persons

## Checklist

-   [ ] Jest tests

### DIFF
--- a/src/utils/db/db.ts
+++ b/src/utils/db/db.ts
@@ -324,7 +324,7 @@ export class DB {
                                     team_id: teamId,
                                     is_identified: isIdentified,
                                     id: uuid,
-                                    deleted: 0
+                                    is_deleted: 0
                                 })
                             ),
                         },
@@ -436,7 +436,7 @@ export class DB {
                                     team_id: person.team_id,
                                     is_identified: person.is_identified,
                                     id: person.uuid,
-                                    deleted: 1
+                                    is_deleted: 1
                                 })
                             ),
                         },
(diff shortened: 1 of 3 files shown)


src/utils/db/db.ts
```diff
@@ -422,7 +423,27 @@ export class DB {
...
+                    messages: [
+                        {
+                            value: Buffer.from(
+                                JSON.stringify({
+                                    created_at: castTimestampOrNow(person.created_at, TimestampFormat.ClickHouse),
+                                    properties: JSON.stringify(person.properties),
+                                    team_id: person.team_id,
+                                    is_identified: person.is_identified,
+                                    id: person.uuid,
+                                    deleted: 1
```
fuziontech, line 439: I'd make this `is_deleted` to conform with how most schemas do soft deletes. Also to conform with `is_identified` in this table. Just for consistency.

src/utils/db/db.ts
```diff
@@ -246,9 +246,12 @@ export class DB {
     public async fetchPersons(database: Database = Database.Postgres): Promise<Person[] | ClickHousePerson[]> {
         if (database === Database.ClickHouse) {
             const query = `
-                SELECT * FROM person FINAL JOIN (
+                SELECT * FROM person
+                FINAL
```
EDsCODE, line 250: Do we need final here especially with the join on group by latest query that follows?

macobo: Good question, not sure, was there before.

Note this function is only used in tests with clickhouse.

mariusandra: This is indeed only used in tests... and tests were flakey without it.

EDsCODE: cool, shouldn't be a problem

macobo (approved):

mariusandra (approved):

PR #394, opened 2021-05-18
Ensure correct event team ID instead of throwing error

Twixes:
## Changes

#381 made changing an event's team ID by plugins an illegal operation, which is good for security. Here I propose that instead of throwing an error, we just ensure that the ID is never changed, in the background – the former option is especially troublesome because in the current approach it'd be hard for a user to identify which plugin is causing the issue, as the error is thrown after _all_ plugins have ran.

Also, this fixes the behavior for batch processing, which did not gain this protection in #381.

## Checklist

-   [x] Jest tests

### DIFF
--- a/src/worker/plugins/run.ts
+++ b/src/worker/plugins/run.ts
@@ -58,10 +58,6 @@ export async function runProcessEvent(server: PluginsServer, event: PluginEvent)
 
             try {
                 returnedEvent = (await processEvent(returnedEvent)) || null
-                if (returnedEvent.team_id != teamId) {
-                    returnedEvent = null // don't try to ingest events with modified teamIDs
-                    throw new Error('Illegal Operation: Plugin tried to change teamID')
-                }
                 pluginsSucceeded.push(`${pluginConfig.plugin?.name} (${pluginConfig.id})`)
             } catch (error) {
                 await processError(server, pluginConfig, error, returnedEvent)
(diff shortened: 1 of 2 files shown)


yakkomajuri:
I do prefer the approach of handling things in the background, but #381 has a bit of discussion about this:

> Do wonder though about an approach where we don't even expose team ID to plugins, therefore preventing plugin devs from shooting themselves in the foot by changing a prop that was indeed exposed to them.

I think it can be a bit annoying as a plugin dev if I'm given an object that I'm supposedly able to modify but suddenly my change doesn't work and I have to spend some time figuring out why.

Docs can ofc go some way here but anyway.

Not strongly opposed to this (or opposed at all actually), but just leaving some thoughts

neilkakkar:
Interesting! Regarding inability to figure out which plugin caused the issue: this isn't actually true. The error is thrown/processed on the pluginConfig, which means users can know exactly which plugin threw the error:

image

image


.. About the batch, definitely my bad, I assumed that had reached end of life, but it has only been deprecated.

neilkakkar:
Thinking this through: the part that seems weird is the order in which plugins run: since `team_id` is exposed as a readable property, and we have a plugin in between that changes the `team_id`, any plugin afterwards that depends/ wants to read it is borked.

Twixes:
Ah, that's true, may be good to leave the error in then @neilkakkar I made #396 as an alternative.

PR #396, opened 2021-05-19
Slightly change the way changing `team_id` by plugins is illegal

Twixes:
## Changes

So _this_ PR adds the feature to batch processing like #394, but instead of silently restoring the correct `team_id` (unlike master which nullifies the event), it restores it AND throws an IllegalOperationError. This way the impact of illegally operating plugins is minimized, but the problem is still communicated as an error. Closes #394.

## Checklist

-   [x] Jest tests

### DIFF
--- a/src/worker/plugins/run.ts
+++ b/src/worker/plugins/run.ts
@@ -3,6 +3,14 @@ import { PluginEvent } from '@posthog/plugin-scaffold'
 import { PluginConfig, PluginsServer, PluginTaskType } from '../../types'
 import { processError } from '../../utils/db/error'
 
+export class IllegalOperationError extends Error {
+    name = 'IllegalOperationError'
+
+    constructor(operation: string) {
+        super(operation)
+    }
+}
+
 export async function runOnEvent(server: PluginsServer, event: PluginEvent): Promise<void> {
     const pluginsToRun = getPluginsForTeam(server, event.team_id)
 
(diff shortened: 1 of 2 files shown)


src/worker/plugins/run.ts
```diff
@@ -115,6 +123,17 @@ export async function runProcessEventBatch(server: PluginsServer, batch: PluginE
...
                     returnedEvents = (await processEventBatch(returnedEvents)) || []
+                    let wasChangedTeamIdFound = false
+                    for (const returnedEvent of returnedEvents) {
+                        if (returnedEvent.team_id != teamId) {
+                            returnedEvent.team_id = teamId
+                            wasChangedTeamIdFound = true
+                        }
+                    }
+                    if (wasChangedTeamIdFound) {
+                        console.log(new IllegalOperationError('dds').name, 'xxx')
```
neilkakkar, line 134: leftover from debug?

Twixes: Indeed 😳

src/worker/plugins/run.ts
```diff
@@ -3,6 +3,14 @@ import { PluginEvent } from '@posthog/plugin-scaffold'
 import { PluginConfig, PluginsServer, PluginTaskType } from '../../types'
 import { processError } from '../../utils/db/error'
 
+export class IllegalOperationError extends Error {
+    name = 'IllegalOperationError'
+
+    constructor(operation: string) {
```
yakkomajuri, line 9: not even sure it's a nit, but I don't think you need the constructor if it just calls super. The constructor of the parent class should be called anyway if this is omitted. Fine if you like the explicitness though.

mariusandra:
Great work everyone in getting to this solution! LGTM.

Issue #395, opened 2021-05-19
Securely exposing packages that leverage external connections


yakkomajuri:
Sparked by a discussion here: https://github.com/PostHog/plugin-repository/issues/11

We currently allow plugin users to leverage `fetch` and would also want to add support for libraries such as `pg`, which make connections via HTTP, TCP, etc.

However, as said by @mariusandra: "the connections will be made from our VPC, so they could theoretically find their way towards sensitive data".

While the issue came up when looking to add `pg`, @macobo pointed out that `fetch` is just as problematic: "...`clickhouse` exposes a HTTP api which can be used for evil things there." I guess the difference is there's no plugin that makes an HTTP request via `fetch` to an arbitrary host (not on Cloud at least), while plugins to send data to a Postgres instance for example would need to accept an arbitrary host.

## How to solve?

Well, this is the million dollar question. Some of this can be solved on an individual use-case basis, but it would be great to have a mechanism that applies to all plugins.

There's many possible levels to this:

### 1. One-off (individual plugin) basis

Ensuring all plugins that make it to Cloud have limitations on user inputs for connection parameters. This would be handled at the plugin level (e.g. doing a check on the host, port, etc.). This can be done easily now if we want to ship some plugins fast but is not a scalable approach.

### 2. Limitations at the library level

Another approach is to patch/fork the libraries we expose to plugin devs. This means they could safely take user input and make connections but the library would prevent anything that's dangerous. This is also not scalable if we want to do #286.

### 3. Server-level 

Ultimately (in an ideal world) we would want to get to a point where plugin devs can do as they wish, with all available packages, and we'd still be safe. This would require some configuration at the top level (preventing connections within the VPC). Quoting @macobo: "I guess next thing to dig into is how networking in node works in general and see whether there's a central hook to monkey-patch or figure out how to do some DNS trickery so plugins can't access instances."

## Discussion

So, what next? Do we want to start with 1 or 2 to ship certain plugins (e.g. Redshift)? Or do we want to take a step back and figure this all out at once?

Twixes:
I'm pretty sure there's no way to do something bad network-side from a Node VM (aside from plain dumb denial of service). File system access, that's definitely a concern regarding libraries and something we must avoid. Arbitrary requests, should be safe.

mariusandra:
Oh well, let's just add `node-postgres` then. 

[...]

PR #397, opened 2021-05-19
Validate hosts on available libs

yakkomajuri:
## Changes

**This is not scalable!**

We definitely shouldn't have to write wrappers around all of these. _However_, as we currently stand, this is a temp fix.

There are plugins out there using `fetch` on user inputs, so this at least gives us more time to dig into #395 

Only exporting `pg` `Client` for now because I decided to use that instead of `Pool` for Redshift. I think the logic should work for `Pool` too but I'm clocking off for the day.

## Checklist

-   [ ] Jest tests


mariusandra:
I think this hostname-based validation is actually a dead end. 

You can just register a `server.thisistotallylegit.com` dns record and point it to the IP of the database instead of just the domain. 

If you're lazy, you can use a service like `nip.io` and create a domain that dynamically maps to a domain:
- 10.0.0.1.nip.io maps to 10.0.0.1
- 192-168-1-250.nip.io maps to 192.168.1.250
- 0a000803.nip.io maps to 10.0.8.3

src/worker/vm/imports.ts
```diff
@@ -2,14 +2,76 @@ import { BigQuery } from '@google-cloud/bigquery'
...
+import nodePostgres from 'pg'
+import { parse } from 'pg-connection-string'
 import snowflake from 'snowflake-sdk'
+import url from 'url'
 import * as zlib from 'zlib'
 
 import { writeToFile } from './extensions/test-utils'
 
+// 93.184.216.34 = example.com
+const RESTRICTED_IPS = ['93.184.216.34']
```
yakkomajuri, line 17: would now be a matter of configuring this

yakkomajuri:
Going down the list on @macobo's suggestions in https://github.com/PostHog/plugin-server/issues/395, this now uses Node's built-in `dns.lookup` to get the IP for a host and check if that is in a list of restricted IPs we maintain.

src/worker/vm/imports.ts
macobo: This will be hard to maintain - we'll be adding services without this blacklist.

Instead we could blacklist the whole block of IPs in our VPC which is `172.31.0.0/16`

src/worker/vm/imports.ts
```diff
@@ -2,14 +2,76 @@ import { BigQuery } from '@google-cloud/bigquery'
...
+    )
+
+const validateHostOrUrl = async (hostOrUrl: any) => {
+    if (typeof hostOrUrl != 'string') {
+        throw new Error(`Invalid host/URL ${hostOrUrl}`)
+    }
+    if (
+        (hostOrUrl.includes('amazonaws.com') && !hostOrUrl.includes('redshift.amazonaws.com')) ||
+        hostOrUrl.includes('localhost') ||
+        hostOrUrl.includes('posthog.net') ||
```
macobo, line 35: I suggest doing these `hostOrUrl` lookups on the `parsedHost` parameter below. This way we can use exact match and avoid issues with `posthog.netsomeclient.com`

macobo (commented):
This looks pretty amazing!

src/worker/vm/imports.ts
yakkomajuri: Hmm this lingered from the previous changes but I guess amazon checks could go below. IP checks and localhost should happen here though because of `url.parse`:

```
url.parse('localhost:8000').hostname // '8000'
url.parse('127.0.0.1').hostname // null
```
Having written this though I guess IPs could be checked via a null hostname?

yakkomajuri: Ok so IP addresses following IPv4 return null hostnames, IPv6 addresses don't.

Kept `localhost` and IPv6 here and moved the rest under

src/worker/vm/imports.ts
```diff
@@ -2,14 +2,70 @@ import { BigQuery } from '@google-cloud/bigquery'
...
+const isIpv6 = (str: string) =>
+    /(([0-9a-fA-F]{1,4}:){7,7}[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,7}:|([0-9a-fA-F]{1,4}:){1,6}:[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,5}(:[0-9a-fA-F]{1,4}){1,2}|([0-9a-fA-F]{1,4}:){1,4}(:[0-9a-fA-F]{1,4}){1,3}|([0-9a-fA-F]{1,4}:){1,3}(:[0-9a-fA-F]{1,4}){1,4}|([0-9a-fA-F]{1,4}:){1,2}(:[0-9a-fA-F]{1,4}){1,5}|[0-9a-fA-F]{1,4}:((:[0-9a-fA-F]{1,4}){1,6})|:((:[0-9a-fA-F]{1,4}){1,7}|:)|fe80:(:[0-9a-fA-F]{0,4}){0,4}%[0-9a-zA-Z]{1,}|::(ffff(:0{1,4}){0,1}:){0,1}((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])|([0-9a-fA-F]{1,4}:){1,4}:((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9]))/.test(
+        str
+    )
+
+const validateHostOrUrl = async (hostOrUrl: any) => {
+    if (typeof hostOrUrl != 'string') {
+        throw new Error(`Invalid host/URL ${hostOrUrl}`)
+    }
+    if (hostOrUrl.includes('localhost') || isIpv6(hostOrUrl)) {
```
Twixes, line 29: Loopback IP addresses (starting with `127.`) should be banned too by default, in addition to `localhost` (which is a synonym for `127.0.01`.

BTW why disallow all IPv6?

src/worker/vm/imports.ts
```diff
@@ -2,14 +2,70 @@ import { BigQuery } from '@google-cloud/bigquery'
...
+        throw new Error(`Host ${hostOrUrl} is not allowed`)
+    }
+
+    const parsedHost = url.parse(hostOrUrl).hostname
+
+    if (
+        !parsedHost || // IPv4 addresses return a null host, IPv6 addresses do not
+        (parsedHost.includes('amazonaws.com') && parsedHost !== 'redshift.amazonaws.com') ||
```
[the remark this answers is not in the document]
yakkomajuri: Have been gathering everyone's input into this. 

This came from here:

https://github.com/PostHog/plugin-server/issues/395#issuecomment-844417191

src/worker/vm/imports.ts
```diff
@@ -2,14 +2,70 @@ import { BigQuery } from '@google-cloud/bigquery'
...
 
 import { writeToFile } from './extensions/test-utils'
 
+// 93.184.216.34 = example.com
+const RESTRICTED_IPS = ['93.184.216.34']
+
+const isIpv6 = (str: string) =>
+    /(([0-9a-fA-F]{1,4}:){7,7}[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,7}:|([0-9a-fA-F]{1,4}:){1,6}:[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,5}(:[0-9a-fA-F]{1,4}){1,2}|([0-9a-fA-F]{1,4}:){1,4}(:[0-9a-fA-F]{1,4}){1,3}|([0-9a-fA-F]{1,4}:){1,3}(:[0-9a-fA-F]{1,4}){1,4}|([0-9a-fA-F]{1,4}:){1,2}(:[0-9a-fA-F]{1,4}){1,5}|[0-9a-fA-F]{1,4}:((:[0-9a-fA-F]{1,4}){1,6})|:((:[0-9a-fA-F]{1,4}){1,7}|:)|fe80:(:[0-9a-fA-F]{0,4}){0,4}%[0-9a-zA-Z]{1,}|::(ffff(:0{1,4}){0,1}:){0,1}((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])|([0-9a-fA-F]{1,4}:){1,4}:((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9]))/.test(
+        str
+    )
```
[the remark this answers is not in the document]
yakkomajuri: DuckDuckGo results :D

src/worker/vm/imports.ts
yakkomajuri: All IPs are actually blocked, not only IPv6. It's just that a few lines down checks for IPv4s. 

Also input came from here:

https://github.com/PostHog/plugin-server/issues/395#issuecomment-844417191

src/worker/vm/imports.ts
```diff
@@ -2,14 +2,70 @@ import { BigQuery } from '@google-cloud/bigquery'
...
+    if (hostOrUrl.includes('localhost') || isIpv6(hostOrUrl)) {
+        throw new Error(`Host ${hostOrUrl} is not allowed`)
+    }
+
+    const parsedHost = url.parse(hostOrUrl).hostname
+
+    if (
+        !parsedHost || // IPv4 addresses return a null host, IPv6 addresses do not
+        (parsedHost.includes('amazonaws.com') && parsedHost !== 'redshift.amazonaws.com') ||
+        parsedHost.includes('posthog.net')
```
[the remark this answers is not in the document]
yakkomajuri: Yeah, was thinking of this too but didn't do anything about it. Let me update it

src/worker/vm/imports.ts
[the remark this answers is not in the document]
yakkomajuri: not necessary anymore with the range check though I guess

src/worker/vm/imports.ts
Twixes: Ah, "IPv4 addresses return a null host", that's a tiny bit hacky.
As was mentioned somewhere, for a malicious actor it's trivial to bypass this using any domain, while regular users could just get frustrated (especially with a vague error message), so I'd like to push back on this.

src/worker/vm/imports.ts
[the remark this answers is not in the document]
Twixes: Ah, sorry I didn't chime in earlier. I don't see a benefit here really. amazonaws.com is not an internal PostHog hostname at all.

yakkomajuri:
Ok so having taken everyone's input into account, here's where we're at:

- We check inputs going to `fetch` but not `pg` (databases have their own security mechanisms)
- For `fetch`, we:
    - Disallow `localhost`
    - Allow IPs, except those starting with `127`
    - Allow all hosts by default, unless they resolve to an IP within our VPC

So, going once, going twice....

yakkomajuri:
Actually I guess now we should check hosts for `pg` too

yakkomajuri:
Now if we want to check hosts passed to `pg` as well we'll need to change the API and do something like make our own functions that return an instance of `Pool` or `Client` rather than extending the class since we can't have the constructor `await` for the validation

mariusandra (changes requested):
Some thoughts inline as comments, and then:

- If we already release this PR for `fetch`, it's a win. We can iterate on top of that.
- For postgres, I think it's fine to release `Pool` in a wrapper. We shouldn't support non-pooled connections anyway, as we can have 10 tasks running the same VM in a worker at the same time, so already then we need some pooling logic, even to just await until the connection is free again.
- Could even then there be some way to do `pool.getConnection().url = "new url"` or something? I haven't checked this, just putting it out there.
- What about the S3, BigQuery and Snowflake plugins? Wrapping every call in the AWS SDK sounds like a nightmare. Same for BigQuery. Snowflake seems easier somehow as it's a very simple wrapper they have.
- The simplest temporary solution for all of those is to expose the `validateHostOrUrl` function for inclusion inside plugins. Then we can at least have an extra layer of security for cloud by just requiring this in all globally installed plugins.

src/worker/vm/imports.ts
```diff
@@ -2,14 +2,70 @@ import { BigQuery } from '@google-cloud/bigquery'
...
+        throw new IllegalOperationError(`${hostOrUrl} is not allowed for security reasons`)
+    }
+
+    const parsedHost = url.parse(hostOrUrl).hostname
+    if (!parsedHost) {
+        throw new IllegalOperationError(`Could not find a hostname for ${hostOrUrl}`)
+    }
+
+    const lookupResult = await dns.promises.lookup(parsedHost)
+    if (isIpInRestrictedRange(lookupResult.address)) {
```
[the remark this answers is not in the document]
yakkomajuri: Not checking IPv6 against the `127` equivalent, but the lib for range checking handles the range checking - from their docs:

```
// IPv4-mapped IPv6 addresses are automatically converted back to IPv4 addresses
// and will match against IPv4 CIDR/IP's.
ipRangeCheck("0:0:0:0:0:FFFF:222.1.41.90", "222.1.41.0/24")
// > true
```

mariusandra:
Also, what happens if we `fetch` a redirect to an URL that we're trying to prevent the user from accessing?

Something like `https://bagofchips.org/redirect.php?url=http://localhost:3212/clickhouse/` that just sets the http redirect header.

yakkomajuri:
One solution is to set `node-fetch` to error on a redirect. I think that's reasonable to ask of plugin devs/users.

yakkomajuri:
@mariusandra made all the updates. Also set node-fetch to error on redirects. Seems reasonable + it's nicer than manual follow and check. If we get complaints could change my mind.

yakkomajuri:
Well, still need to get this in `plugin-contrib`

mariusandra:
I think we need redirects out of the box. They are so common (someone mistypes http, some service changed their domain and redirects traffic, who knows what else) that without them `fetch` will feel broken.

yakkomajuri:
Will do a separate PR for getting this into contrib along with some other helpers

mariusandra:
What if it's an endless redirect loop? Those happen :)

yakkomajuri:
This should be good to go, let me fix conflicts

yakkomajuri:
Testing `fetch` has become a bit hard with the manual redirect follows and now the retry queue tests. This is passing locally though...

yakkomajuri:
Pretty weird that everything keeps passing for me locally
Screenshot 2021-06-02 at 15 54 27

yakkomajuri:
Uff, well this took me a while of running in circles only to find that short delays were the reason why tests passed locally and not remotely

cc @mariusandra for the green check

src/worker/vm/imports.ts
```diff
@@ -3,14 +3,90 @@ import * as contrib from '@posthog/plugin-contrib'
...
+    const res = await nodeFetch(url, { redirect: 'manual' })
+    if (res.headers && res.headers.get('location')) {
+        setOfRedirectsFollowed.add(res.headers.get('location')!)
+        await checkRedirectChain(
+            res.headers.get('location')!,
+            originalUrl,
+            numberOfRedirectsFollowed + 1,
+            setOfRedirectsFollowed
+        )
+    }
```
mariusandra, line 87: Won't we potentially need to pass along some cookies for completeness? Maybe not, I'm not sure here.

Probably fine to ignore that, but did you check the spec (or the implementation in a polyfill) or just implement what was needed to get it working?

src/worker/vm/imports.ts
```diff
@@ -3,14 +3,90 @@ import * as contrib from '@posthog/plugin-contrib'
...
+import nodeFetch, { RequestInit } from 'node-fetch'
+import pg from 'pg'
 import snowflake from 'snowflake-sdk'
+import url from 'url'
 import * as zlib from 'zlib'
 
+import { IllegalOperationError } from './../plugins/run'
 import { writeToFile } from './extensions/test-utils'
 
+interface URLLike {
```
mariusandra, line 19: This `vm` file is definitely not the right place for all this IP and `fetch` code (the next ~80 lines).

src/worker/vm/imports.ts
```diff
@@ -3,14 +3,90 @@ import * as contrib from '@posthog/plugin-contrib'
...
+        throw new IllegalOperationError(`Could not find a hostname for ${hostOrUrl}`)
+    }
+
+    const lookupResult = await dns.promises.lookup(parsedHost)
+    if (isIpInRestrictedRange(lookupResult.address)) {
+        throw new IllegalOperationError(`IP ${lookupResult.address} is not allowed for security reasons`)
+    }
+}
+
+export const fetch = async (url: string | URLLike, init?: RequestInit) => {
```
mariusandra, line 57: TS error

tests/postgres/vm.test.ts
```diff
@@ -14,6 +14,20 @@ jest.mock('../../src/utils/celery/client')
...
+// expect.anything() will not match null and undefined
+const passOnAnything = () => ({
+    message: () => 'Ok',
+    pass: true,
+})
+
+const extendedExpect = expect as ExtendedExpect
+extendedExpect.extend({
+    passOnAnything,
+})
```
mariusandra, line 27: Why is any of this needed? Can't we just use `expect.any()` or then explicitly check for `null` or `undefined` when needed?

mariusandra (changes requested):
Some more thoughts.

tests/postgres/vm.test.ts
yakkomajuri: I actually had a problem where the tests were all testing `fetch` calls, and with the manual redirects testing, the test call would have with an options object `{ redirect: 'manual' }` and the actual call would have `undefined`. `expect.any()` nor `expect.anything()` were suitable. 

Another approach is to do `init || {}` to the actual fetch call to prevent it from being undefined. Then I guess `expect.any` could be used. 

Some discussion on solutions for matching `undefined` and "any object" https://stackoverflow.com/questions/47027011/using-expect-anyobject-or-expect-anything-does-not-work-to-match-undefi

src/worker/vm/extensions/fetch.ts
```diff
@@ -0,0 +1,80 @@
...
+const RESTRICTED_IP_RANGES =
+    process.env.NODE_ENV === 'test'
+        ? ['127.0.0.1/8', '116.203.255.68']
+        : (process.env.RESTRICTED_IP_RANGES || '').split(',')
+
+const isIpInRestrictedRange = (ip: string): boolean => {
+    return ipRangeCheck(ip, RESTRICTED_IP_RANGES)
+}
+
+const validateHostOrUrl = async (hostOrUrl: any) => {
```
Twixes, line 23: Could we have _unit_ tests for this function? That way we could cover cases really cleanly (in addition the existing tests checking that this is indeed raised in VMs)

src/worker/vm/extensions/fetch.ts
```diff
@@ -0,0 +1,80 @@
...
+    if (isIpInRestrictedRange(lookupResult.address)) {
+        throw new IllegalOperationError(`IP ${lookupResult.address} is not allowed for security reasons`)
+    }
+}
+
+export const fetch = async (url: string | URLLike, init?: RequestInit): Promise<Response> => {
+    if (typeof url === 'object' && 'href' in url) {
+        url = url.href
+    }
+    await checkRedirectChain(url, init?.headers)
```
Twixes, line 51: I'm worried that this could result in bad issues if we ping a non-idempotent URL more than once this way (and discarding the result too). I think it'd be better to forgo this check.

Twixes (changes requested):

yakkomajuri:
I'm closing this as this was a quick change that has stretched out too far and we all seem to have diverging opinions. 

Let's revisit when the council convenes to agree on a more clear approach to #454

mariusandra:
I think that's a wise choice. My main gripe with this PR was that it improves *some* things, but not nearly enough to make things as secure as we want. Thus it's easy to walk away with the illusion of security, and that's dangerous. Some form of network level isolation is the right answer here, and we can't address it just inside the plugin server.

Plus the PR introduces new loose parts that will need watering over time. If the extensibility team is understaffed, it's best to have less code :).

PR #398, opened 2021-05-20
Fix tests for plugin-server, borked by #4366

neilkakkar:
## Changes

Ref: https://github.com/PostHog/posthog/pull/4366

## Checklist

-   [x] Jest tests

### DIFF
--- a/tests/helpers/sql.ts
+++ b/tests/helpers/sql.ts
@@ -108,6 +108,7 @@ export async function createUserTeamAndOrganization(
         updated_at: new Date().toISOString(),
         personalization: '{}',
         setup_section_2_completed: true,
+        for_internal_metrics: false,
     })
     await insertRow(db, 'posthog_organizationmembership', {
         id: organizationMembershipId,


Twixes (approved):
Such is life

tests/helpers/sql.ts
```diff
@@ -108,6 +108,7 @@ export async function createUserTeamAndOrganization(
         updated_at: new Date().toISOString(),
         personalization: '{}',
         setup_section_2_completed: true,
+        for_internal_metrics: false,
```
tiina303, line 111: Why is this needed if the module defaults to false? https://github.com/PostHog/posthog/blob/master/posthog/models/organization.py#L83

neilkakkar: .. because Django sets the defaults on the Model object, not the database, so anyone else using raw SQL will bork things :/

neilkakkar: wow, it's been on their minds (in some shape or form) for ~15 years: https://code.djangoproject.com/ticket/470

Issue #364, opened 2021-05-07
S3 job queue


mariusandra:
Build a S3 job queue as a backup in case Graphile/Postgres is down.

PR #399, opened 2021-05-20
S3 Job Queue

mariusandra:
## Changes

- Still WIP
- Implements a simple redlocked job queue based on S3. 
- Needs tests
- Closes https://github.com/PostHog/plugin-server/issues/364

## Checklist

-   [ ] Jest tests


mariusandra:
Here's something that could use a look. :)

I grouped the different job queues into folders based on functionality (local, concurrent, redlocked), but didn't yet implement different functionalities depending on the type. This means currently all job queues are still redlocked. I'd suggest tackling this in another PR.

Twixes (approved):
Looks reasonable to me

src/main/job-queues/redlocked/s3-queue.ts
```diff
@@ -0,0 +1,180 @@
...
+        await this.connectS3()
+    }
+
+    async enqueue(retry: EnqueuedJob): Promise<void> {
+        if (!this.s3Wrapper) {
+            throw new Error('S3 object not initialized')
+        }
+        const date = new Date(retry.timestamp).toISOString()
+        const [day, time] = date.split('T')
+        const dayTime = `${day.split('-').join('')}-${time.split(':').join('')}`
```
neilkakkar, line 37: was there a specific reason for formatting things like this?

mariusandra: Basically just keeping it sortable to the second while removing special symbols that might interfere with it being a filename.

PR #403, opened 2021-05-21
In-memory action definitions synced with Django

### LINKED ISSUE #235: Match actions to events on the fly
We should probably match actions to events on the fly in the plugin server.

At the highest level:
1. On the plugin server, keep in memory all actions and all action steps for all teams. 
2. After an event is ingested, see if any of the action steps apply to it. If so, create the action.

There are three different types of ActionSteps that we can match directly:
- Autocapture: event == '$autocapture' && match elements via selector && url contains/regex && ...
- Custom Event: event == 'custom event'
- Pageview: event == '$pageview' && url contains/regex

Filters on all three of them add some complexity:
- Event property equals/contains/etc something --> we have the event's properties, so easy to do all possible comparisons
- User property contains something --> we need the `Person` for this. We are `select`ing it for all posthog-js events due to some automatic `$set` properties, so it's probably fine to also fetch the person (if not already fetched) to match a property for an action
- User belongs to a cohort --> add one extra query to get all cohorts for a person? ... or inlined in the fetch above?

In addition:
- We will need a system to keep all these ActionSteps in memory in the plugin server and reload them when changed. Similar to `PluginConfig`s now.

However:
- We still need the django async action-event mapping code to clean up after changes to actions (calculating action...)
- Should we find a way to run webhooks on late-matched or back-matched actions?

Shouldn't be that hard to pull off?

--- 

Next step after this: automatically update cohorts after ingesting events.

Twixes:
## Changes

Part of #235.

- [x] In-memory synced list of action steps for all teams, reloading/adding/removing data for a specific action based on pubsub dispatched when the action is updated in the app, using Django model signals
- Action matching logic:
	- Action step type:
		1. Autocapture:
			- [ ] Link href equals
			- [ ] Text equals
			- [ ] HTML selector match
			- [ ] URL contains
			- [ ] URL matches regex
			- [ ] URL matches exactly
		2. Custom view
			- [ ] Event name matches exactly
		3. Page view
			- [ ] Event name matches exactly (`$pageview`)
			- [ ] URL contains
			- [ ] URL matches regex
			- [ ] URL matches exactly
	- General filtering:
		1. Property (event & person & element):
			  - [ ] Equals (or not)
			  - [ ] Contains (or not)
			  - [ ] Matches (or not) regex
			  - [ ] Greater than/lower than
			  - [ ] Set (or not)
		2. Cohort:
			- [ ] Person belonging
- [ ] `onAction` Piscina task
- [ ] Insertion of new action occurrence into Postgres `posthog_action_events` table (skipped for ClickHouse)
- [ ] Webhooks firing
- [ ] REST hooks (Zapier) firing

## Checklist

-   [x] Jest tests


Twixes:
This is a diff of almost 500 lines already, would you prefer to merge just this action syncing feature alone (with it tested, but disabled in ingestion), or merge all of action matching at once (will be a big diff)?

src/worker/ingestion/process-event.ts
```diff
@@ -38,22 +41,40 @@ export class EventsProcessor {
...
         this.clickhouse = pluginsServer.clickhouse
         this.kafkaProducer = pluginsServer.kafkaProducer
         this.celery = new Client(pluginsServer.db, pluginsServer.CELERY_DEFAULT_QUEUE)
         this.teamManager = new TeamManager(pluginsServer.db)
         this.personManager = new PersonManager(pluginsServer)
+        this.actionManager = new ActionManager(pluginsServer.db)
+        this.pubSub = new PubSub(pluginsServer, {
+            'fetch-action': async (message) => await this.actionManager.fetchAction(parseInt(message)),
+            'delete-action': (message) => this.actionManager.deleteAction(parseInt(message)),
+        })
```
mariusandra, line 60: The only problem with this is that it'll add another redis connection per worker thread, which can add up. I guess even if we have officially given up on Heroku free redis tiers, it might be good to reload these inside the workers with a similar `broadcastTask` system that plugin reloads use.

src/worker/ingestion/action-manager.ts
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    public getAction(id: Action['id']): Action | undefined {
+        if (!this.ready) {
+            throw new Error('ActionManager is not ready! Run actionManager.prepare() before this')
+        }
+        return this.actionCache[id]
+    }
+
+    public async fetchAction(id: Action['id']): Promise<void> {
```
mariusandra, line 30: The name of this function is misleading. It's more `reloadAction` than "fetch (and return) from the database"

tests/worker/ingestion/action-manager.test.ts
```diff
@@ -0,0 +1,85 @@
...
+            })
+
+            await server.db.postgresQuery(
+                `UPDATE posthog_actionstep SET properties = jsonb_set(properties, '{0,key}', '"baz"') WHERE id = 911`,
+                undefined,
+                'testKey'
+            )
+
+            // This is normally done by Django async in such a situation
+            await actionManager.fetchAction(67)
```
mariusandra, line 57: Yup, echoing my point from above, the name is misleading here. What happens to the fetched action? Not obvious it actually caches it in a map... :)

mariusandra: Other than calling this here, should we also have a bigger and more E2E test to make sure actions actually reload in the workers? Fine to push that for a later PR.

mariusandra (changes requested):
Few thoughts inline. Otherwise, I think it's good to merge this feature chunk by chunk.

src/worker/ingestion/process-event.ts
Twixes: Agreed, used similar system as `reloadPlugins`, avoiding increase of Redis connections with per-thread pubsub instances

src/worker/ingestion/action-manager.ts
Twixes: Yeah, was definitely on the fence about this, went for `reloadAction` and `dropAction`

tests/worker/ingestion/action-manager.test.ts
Twixes: At this stage it's hard to test this E2E with the pubsub only in the main thread (whole `startPluginsServer` needed instead of just `createServer`), I don't know about reaching into `actionManagers` in worker threads, and there's no actual functionality (like `onAction`) to test here yet. But definitely a goal to have E2E tests when more parts are in place.

src/main/pluginsServer.ts
```diff
@@ -141,17 +141,19 @@ export async function startPluginsServer(
...
                 status.info('⚡', 'Reloading plugins!')
-
                 await piscina?.broadcastTask({ task: 'reloadPlugins' })
                 await scheduleControl?.reloadSchedule()
-            }
+            },
+            'reload-action': async (message) =>
+                await piscina?.broadcastTask({ task: 'reloadAction', args: { actionId: parseInt(message) } }),
+            'drop-action': async (message) =>
+                await piscina?.broadcastTask({ task: 'dropAction`', args: { actionId: parseInt(message) } }),
```
mariusandra, line 154: There's some "`" here after "dropAction". Can you make sure this code path is also tested?

src/worker/tasks.ts
```diff
@@ -48,6 +45,12 @@ export const workerTasks: Record<string, TaskRunner> = {
     reloadSchedule: async (server) => {
         await loadSchedule(server)
     },
+    reloadAction: async (server, args: { actionId: Action['id'] }) => {
+        return await server.eventsProcessor.actionManager.reloadAction(args.actionId)
+    },
+    dropAction: (server, args: { actionId: Action['id'] }) => {
+        return server.eventsProcessor.actionManager.dropAction(args.actionId)
+    },
```
mariusandra, line 53: Can you add these to worker.test.ts

mariusandra (changes requested):
I'm sure this will work well (once all code paths are tested/covered, see below :P), however as a general observation, I'm afraid of "eventual inconsistency" here. In case we miss some drop/reload signal from the app, we'll never reach consistency. 

This dropped signal might seem unlikely, but experience shows that things breaking randomly is more likely than not, given enough time :). Even if not today, then someone could accidentally introduce a 30sec "await" after the actions load, but before the pubsub started, or ... whatever imaginary case.

With reloads, this is less of an issue since reloads happen frequently, and always read in all plugins and diff changes. However here I'd add a "every 15min do a manual sync" just in case.

src/types.ts
```diff
@@ -423,6 +424,115 @@ export interface CohortPeople {
...
+    selector: string | null
+    url: string | null
+    url_matching: ActionStepUrlMatching | null
+    name: string | null
+    event: string | null
+    properties: ActionStepProperties[] | null
+}
+
+/** Raw Action row from database. */
+export interface RawAction {
```
neilkakkar, line 518: duplicate?

src/utils/status.ts
```diff
@@ -22,6 +22,10 @@ export class Status implements StatusBlueprint {
     }
 
     buildMethod(type: keyof StatusBlueprint): StatusMethod {
+        if (process.env.NODE_ENV?.toLowerCase() === 'test') {
```
neilkakkar, line 25: what's this for? O.o

src/worker/ingestion/action-manager.ts
```diff
@@ -0,0 +1,63 @@
...
+    private actionCache: ActionCache
+
+    constructor(db: DB) {
+        this.ready = false
+        this.db = db
+        this.actionCache = {}
+    }
+
+    public async prepare(): Promise<void> {
+        this.actionCache = await this.db.fetchAllActionsMap()
```
neilkakkar, line 19: Not sure how exactly action manager will be used yet, so:

If something goes wrong with fetching actions, such that this cache isn't populated, do we want to stop ingestion completely? or continue?

tests/helpers/sql.ts
```diff
@@ -144,6 +157,32 @@ export async function createUserTeamAndOrganization(
...
+        action_id: teamId + 67,
+        tag_name: null,
+        text: null,
+        href: null,
+        selector: null,
+        url: null,
+        url_matching: null,
+        name: null,
+        event: null,
+        properties: JSON.stringify([{ type: 'event', operator: PropertyOperator.Exact, key: 'foo', value: ['bar'] }]),
```
neilkakkar, line 184: now that my PR is in, you shouldn't need to `JSON.stringify` arrays here!

src/types.ts
```diff
@@ -423,6 +424,115 @@ export interface CohortPeople {
...
+    NotIContains = 'not_icontains',
+    Regex = 'regex',
+    NotRegex = 'not_regex',
+    GreaterThan = 'gt',
+    LessThan = 'lt',
+    IsSet = 'is_set',
+    IsNotSet = 'is_not_set',
+}
+
+/** Sync with posthog/frontend/src/types.ts */
```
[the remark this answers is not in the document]
Twixes: I think what you mean is added in https://github.com/PostHog/posthog/pull/4436 (to be merged after this)

src/types.ts
Twixes: Indeed

src/utils/db/db.ts
```diff
@@ -705,14 +708,53 @@ export class DB {
...
+        const rawActions: RawAction[] = (
+            await this.postgresQuery(`SELECT * FROM posthog_action WHERE id = $1`, [id], 'fetchActions')
+        ).rows
+        if (!rawActions.length) {
+            return null
+        }
+        const steps: ActionStep[] = (
+            await this.postgresQuery(`SELECT * FROM posthog_actionstep WHERE action_id = $1`, [id], 'fetchActionSteps')
+        ).rows
+        const action: Action = { ...rawActions[0], steps }
```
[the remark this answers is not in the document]
Twixes: No, it's a primary key

src/worker/ingestion/action-manager.ts
Twixes: Well, in that case an error will be thrown (if e.g. Postgres connection fails) and indeed ingestion will be stopped

neilkakkar: Right, I'm asking if that's what we ought to do, or not?

Twixes: Ah, yes, we do, since otherwise data integrity will be at risk (action matching wouldn't work). Besides, this is not likely to fail for internal reasons, and if Postgres (the external dependency) fails, then we are in catastrophic failure territory anyway

Twixes:
Done @mariusandra resyncing all every 5 minutes + added taskRunner tests.

src/utils/db/db.ts
```diff
@@ -727,10 +727,16 @@ export class DB {
 
     public async fetchAllActionsMap(): Promise<Record<Action['id'], Action>> {
         const rawActions: RawAction[] = (
-            await this.postgresQuery(`SELECT * FROM posthog_action`, undefined, 'fetchActions')
+            await this.postgresQuery(`SELECT * FROM posthog_action WHERE deleted = FALSE`, undefined, 'fetchActions')
```
mariusandra, line 730: nice catch!

src/utils/status.ts
mariusandra, line 25: nit: I'm not sure why the `.toLowerCase()` is needed.

src/utils/status.ts
```diff
@@ -22,6 +22,10 @@ export class Status implements StatusBlueprint {
...
+            // TODO: use determineNodeEnv() here
+            return () => {} // eslint-disable-line @typescript-eslint/no-empty-function
+        }
```
mariusandra, line 28: I get the frustration with this, though sometimes when coding locally logging is good to have. It's somewhat unintuitive to find this piece of code to re-enable it. 

I think there's on alternative to this, adding [`LOG_LEVEL=none`](https://github.com/PostHog/plugin-server/blob/b3dbc21dc6574fc129d0d492c750c56ad03554aa/src/config/config.ts#L50) as a default for test mode.

It might have a bug, in that the existing console patching mechanism will also swallow messages sent by jest that would be good to still see. If so, the solution would be to just add this log level filtering inside `status`.

src/utils/pubsub.ts
```diff
@@ -0,0 +1,54 @@
...
+        const channels = Object.keys(this.taskMap)
+        await this.redis.subscribe(channels)
+        this.redis.on('message', (channel: string, message: string) => {
+            const task: PubSubTask | undefined = this.taskMap[channel]
+            if (!task) {
+                throw new Error(
+                    `Received a pubsub message for unassociated channel ${channel}! Associated channels are: ${Object.keys(
+                        this.taskMap
+                    ).join(', ')}`
+                )
```
mariusandra, line 38: I wonder if this can backfire? E.g. through some upgrade django starts sending `runCommand` via pubsub before the plugin servers have restarted to receive it. Failing silently to sentry would be better probably.

mariusandra (approved):
That's a nice bit of code! Let's see what it does 👀 

I added a few comments still inline, feel free to fix or defer.

src/main/pluginsServer.ts
```diff
@@ -141,23 +141,29 @@ export async function startPluginsServer(
...
         })
+        await pubSub.start()
 
         if (hub.jobQueueManager) {
             const queueString = hub.jobQueueManager.getJobQueueTypesAsString()
             await hub!.db!.redisSet('@posthog-plugin-server/enabled-job-queues', queueString)
         }
 
+        // every 5 minutes all ActionManager caches are reloaded for eventual consistency
+        pingJob = schedule.scheduleJob('*/5 * * * *', async () => {
```
neilkakkar, line 164: think this is overridden by the `pingJob` below?

src/utils/status.ts
Twixes: Eh, removed this path altogether as it had circular import problems with `config`, maybe another PR

src/utils/pubsub.ts
Twixes: Possibly, yeah, moved to just `captureException`

src/main/pluginsServer.ts
Twixes: Good catch

PR #408, opened 2021-05-22
Add `exportEvents` function

### LINKED ISSUE #404: Create `exportEvents` function in plugins that abstracts away all the queueing, batching, retrying logic
We're now manually writing a lot of queueing and retrying/jobs logic into events. 

Queuing: https://github.com/PostHog/posthog-plugin-replicator/blob/974854579a7a51f815f9d3601e9c5b0f075200d8/index.ts#L19-L27

Retrying: https://github.com/PostHog/s3-export-plugin/blob/91e4813e84237e88142107c00929dc6b147963b7/index.ts#L37-L41

This is super powerful stuff, but our export plugins could be better. Something like:

```js
export async function exportEvents (events, { global, config }) {
    try {
        await fetch(`https://${config.host}/e`, {
            method: 'POST',
            body: JSON.stringify(events),
            headers: { 'Content-Type': 'application/json' },
        })
    } catch (error) {
         throw new RetryError() // ask to retry
    }
}
```

It would just be an abstraction over existing tools, and it would take care of 1) retry queues, 2) batching, 3) async processing . 

You literally write the data upload call and we take care of the rest.

Plus, this will unlock one really strong feature: "start from the beginning". More on that later. :D

Open questions: I'm not sure what's the best way to configure this (e.g. some queues can take 1MB of buffer, some want 100kb, some go just one by one)

mariusandra:
## Changes

```js
import { RetryError } from '@posthog/plugin-scaffold' // also a global

export async function exportEvents (events, { global, config }) {
    try {
        await fetch(`https://${config.host}/e`, {
            method: 'POST',
            body: JSON.stringify(events),
            headers: { 'Content-Type': 'application/json' },
        })
    } catch (error) {
         throw new RetryError() // ask to retry
    }
}
```

- Closes #404
- This PR adds support for `exportEvents` by using existing plugin server tools as building blocks. 
- That function receives a batch of events (configurable size) and either exports them (`await fetch`, etc) or throws `RetryError`. If the latter happens, we retry exporting the entire batch again.
- Basically it abstracts away [this code](https://github.com/PostHog/bigquery-plugin/blob/020c0536b6b264b36d3c3144b67fd1bb8a969363/index.ts#L143-L209) from an export plugin into a very similar structure inside the plugin server.
- After the VM is initialized, if an `exportEvents` function is exported, this creates a `buffer` via `createBuffer`, and adds or patches an existing `onEvent` function to add incoming events to that buffer.
- We also add a `exportEventsWithRetry` job to retry events.
- The config for this export comes from `meta.config`. As a plugin author, you can ask these from your users, or stick to the defaults:
   - exportEventsBufferBytes (1MB default)
   - exportEventsBufferSeconds (10 sec default)
   - exportEventsToIgnore (empty default)
- It might feel like a weird place for this, but it turns out to be a nice compromise between 1) having defaults, 2) having a chance for users to specify these fields for certain plugins, 3) not having some `meta.config` parsing boilerplate in `setupPlugin`.

Please rip and nit this to shreds :).

Tests ~~might~~will still fail, will fix if so.

## Checklist

-   [ ] Jest tests


benchmarks/vm/memory.benchmark.ts
```diff
@@ -1,6 +1,6 @@
 import { PluginEvent } from '@posthog/plugin-scaffold/src/types'
 
-import { Plugin, PluginConfig, PluginConfigVMReponse } from '../../src/types'
+import { Plugin, PluginConfig, PluginConfigVMResponse } from '../../src/types'
```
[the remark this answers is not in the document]
mariusandra: Yeah... agreed 😊

src/worker/vm/upgrades/export-events.ts
```diff
@@ -0,0 +1,113 @@
...
+                    const nextRetrySeconds = 2 ** payload.retriesPerformedSoFar * 3
+                    console.log(`Enqueued batch ${payload.batchId} for retry in ${Math.round(nextRetrySeconds)}s`)
+
+                    await meta.jobs
+                        .exportEventsWithRetry({ ...payload, retriesPerformedSoFar: payload.retriesPerformedSoFar + 1 })
+                        .runIn(nextRetrySeconds, 'seconds')
+                } else {
+                    console.log(
+                        `Dropped batch ${payload.batchId} after retrying ${payload.retriesPerformedSoFar} times`
+                    )
```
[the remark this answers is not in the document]
mariusandra: Ok, agreed. Sent to system. I originally implemented this as a way to give the user some status about exports, mainly to see that everything is working well. However yes, this is more system level information.

Because of https://github.com/PostHog/plugin-server/issues/424 I also fear having too many console.logs, so I'm happy to remove them for stability as well.

mariusandra:
Some things to improve still:
- Add `statsd`
- I'd still like to support `throw new RetryError()` in `onEvent`, `onSnapshot` and perhaps also `processEvent`. In the future as well `onRequest` (incoming webhooks), `onAction`, etc.
- I'm not happy with how `onEvent` gets overridden to a function that runs two functions (`batch.add` and `onEvent`). This would break the separate `onEvent` retry dynamic as it stands now.

Twixes:
Looks okay for now. Though merg conflict and red tests.

mariusandra:
I added some statsd and made a new issue to track the other stuff: https://github.com/PostHog/plugin-server/issues/429

Twixes:
One more question: does the buffer get flushed when the plugins gets unloaded?

mariusandra:
Very good question. Answer: yes. It does. Now.. :)

src/worker/vm/upgrades/export-events.ts
```diff
@@ -0,0 +1,152 @@
+import { createBuffer } from '@posthog/plugin-contrib'
```
neilkakkar, line 1: what's the idea behind the `upgrades` folder?

src/worker/vm/upgrades/export-events.ts
```diff
@@ -0,0 +1,152 @@
...
+        exportEventsWithRetry: (payload: ExportEventsJobPayload, meta: PluginMeta<ExportEventsUpgrade>) => Promise<void>
+    }
+    config: {
+        exportEventsBufferBytes: string
+        exportEventsBufferSeconds: string
+        exportEventsToIgnore: string
+    }
+    jobs: {
+        exportEventsWithRetry: ExportEventsJobPayload
+    }
```
neilkakkar, line 29: This is a bit confusing to me: aren't jobs supposed to be callables?

src/worker/vm/upgrades/export-events.ts
neilkakkar: .. ok, this makes sense, since it's overriding existing functionality.

tests/postgres/vm.test.ts
```diff
@@ -939,3 +943,310 @@ test('onSnapshot', async () => {
...
+        `
+        await resetTestDatabase(indexJs)
+        const vm = await createPluginConfigVM(
+            hub,
+            {
+                ...pluginConfig39,
+                config: {
+                    ...pluginConfig39.config,
+                    exportEventsBufferBytes: 10000,
+                    exportEventsBufferSeconds: 1,
```
neilkakkar, line 962: think this has type string above
```suggestion
                    exportEventsBufferSeconds: '1',
```

tests/postgres/vm.test.ts
```diff
@@ -939,3 +943,310 @@ test('onSnapshot', async () => {
...
+            }
+        `
+        await resetTestDatabase(indexJs)
+        const vm = await createPluginConfigVM(
+            hub,
+            {
+                ...pluginConfig39,
+                config: {
+                    ...pluginConfig39.config,
+                    exportEventsBufferBytes: 10000,
```
neilkakkar, line 961:
```suggestion
                    exportEventsBufferBytes: '10000',
```

neilkakkar:
A question about the general flow of data here:

`meta.global` is local to the VM, right? So, we have buffers on each thread which are being filled and flushed as and when events come in for those specific plugins on those specific threads.

This is separate from the runAt / runWhenever functionality which is redlocked.

So, it's entirely possible we're sending X requests together, where X is number of threads over all instances? We may have only 1 instance processing a job at a time, but that doesn't stop the buffer flushes on timeout, since this is separate from the jobs locking mechanism? *until it gets to a retry

(Correct me if I'm wrong somewhere please)

tests/postgres/vm.test.ts
mariusandra: makes sense, fixed

tests/postgres/vm.test.ts
mariusandra: fixed here and elsewhere as well

mariusandra:
@neilkakkar indeed, `meta.global` is vm-specific. I've been thinking it might make sense to call this `meta.shared` instead, but not sure. And indeed, that means a buffer is per-thread, and in the worst case, every thread per server might make a request at the same time. In reality, this might just mean we need to increase the size of the buffers. In any case, it'll still be better than without a buffer, just something to keep in mind going forward :)

neilkakkar:
Agreed, just wanted to (1) ensure I understood correctly, and (2) make my concern explicit. 

It's probably not worrisome right now, and definitely better to have a buffer than not, but as you say, something to keep in mind :)

neilkakkar (approved):
Awesome work!

Twixes (approved):

PR #426, opened 2021-05-25
Support default exports

mariusandra:
## Changes

Makes it possible to write:

```ts
```

## Checklist

-   [ ] Jest tests


Twixes (approved):

mariusandra:
Last time only the VM performance benchmark failed, so I'm calling that good enough and merging!

PR #432, opened 2021-05-26
Capture info on events ingested for consumption in posthog

macobo:
This will go into the internal /instance/status dashboard, used to debug
other posthog installations.

I was really conservative adding these metrics to avoid cascades where
capturing a metric would trigger another metric being captured. If as we
scale this out we find more metrics that help users manage their
instance, we can implement using e.g. in-memory histograms for timers.

## Checklist

-   [ ] Jest tests


mariusandra (approved):
LGTM!

PR #282, opened 2021-03-24
Remove Fastify and web server

Twixes:
## Changes

We've realized it's not feasible to run a plugin web server due to the limitations of Heroku (which is an important deployment platform for us): https://github.com/PostHog/plugin-server/issues/165#issuecomment-805759098
Therefore this PR removes the Fastify web server that has been dead code for a while.
Closes #25.

## Checklist

-   [x] Updated Settings section in README.md, if settings are affected
-   [x] Jest tests

PR #433, opened 2021-05-26
Reorient `ActionManager` to group by `teamId` for practicality

Twixes:
## Changes

Followup to #403, makes `ActionManager` more practical for the actual matching engine.

## Checklist

-   [x] Jest tests


src/worker/ingestion/action-manager.ts
```diff
@@ -20,48 +22,57 @@ export class ActionManager {
...
             throw new Error('ActionManager is not ready! Run actionManager.prepare() before this')
         }
-        return this.actionCache[id]
+        return this.actionCache[teamId] || null
     }
 
     public async reloadAllActions(): Promise<void> {
-        this.actionCache = await this.db.fetchAllActionsMap()
+        this.actionCache = Object.fromEntries(
+            Object.entries(groupBy(await this.db.fetchAllActions(), 'team_id')).map(([teamId, actions]) => [
```
neilkakkar, line 34: Do you expect `db.fetchAllActions()` to be used elsewhere? If not, I think it makes things a lot simpler to just have `db.fetchAllActionsByTeam()` which returns them grouped by team?

Twixes: :+1: That's sensible

neilkakkar: Sorry if I was a bit unclear: I meant `fetchAllActionsGroupedByTeam` to replace `fetchAllActions` (which means getting rid of all the groupBy logic, since we're building the grouping from scratch, just like in `fetchAllActions()` right now.

Thoughts?

Twixes: Ah, I see what you mean, done. Though maybe I'll leave `groupBy` in `utils` since maybe and will come in useful and I've already written it

neilkakkar: Works!

neilkakkar (approved):
A few red tests, but otherwise LGTM

src/worker/ingestion/action-manager.ts
mariusandra: > "maybe it'll come in useful" 
... is a dangerous thing to say :D. See for example the HTTP server (let's remove it) :).

Twixes: LOL, well, at least it's isolated _and_ tested. Unlike the HTTP server… pending removal in #282 for some time

Issue #430, opened 2021-05-25
State of benchmarks


Twixes:
We have a CI suite called "Benchmarks". It's been added a while ago, and now my question is: what state are these benchmarks in really?

Personally, I've just never used them in any real way once they've been added and measured for the first time.
My only interaction has been that they occasionally fail due to a cosmic ray flipping a bit somewhere (not actually) and I have to rerun the GH Actions jobs.

So I was wondering:
@mariusandra Have you been finding the benchmarks useful?
Would making them more prominent (for example by having @posthog-bot post results on PRs) help their case?
Are their results relevant in their current form?

mariusandra:
These benchmarks could have prevented one bug one time, had we seen the slowdown somewhere visually. A few times they as well caught some E2E-ish bugs. If we're running out of action runners, I'd be OK with disabling them. I think I fixed the flakey one with one of the last PRs, so if they don't bother us much, I'd let them stay.

PR #435, opened 2021-05-27
Test for running the built package and remove benchmarks

mariusandra:
## Changes

- I recently had `@posthog/plugin-scaffold` in `devDependencies` instead of `dependencies` and that caused a 30 minutes event ingestion lag on cloud.
- Turns out we don't really test `dist` code, so this test tries to fix that.
- Also deleting the relatively useless benchmark tests. Let's give the CI some breathing room.

## Checklist

-   [ ] Jest tests


Twixes:
I guess this closes #430 then?

mariusandra:
Yep. Obviously still something flaking in `jobs.test.ts`, so re-running tests. In a future PR we could split up the existing tests even more. In my experience if tests run under 5min, it's more or less instant feedback. Anything above that is approaching drudgery. We're at ~8min best case right now.

Twixes (approved):

PR #436, opened 2021-05-27
Action matching juice

### LINKED ISSUE #235: Match actions to events on the fly

Twixes:
## Changes

Resolves #235.

- [x] ~~In-memory synced list of action steps for all teams, reloading/adding/removing data for a specific action based on pubsub dispatched when the action is updated in the app, using Django model signals~~ #403 & #433
- Action matching logic:
	- Action step type:
		1. Autocapture:
			- [x] Link href equals
			- [x] Text equals
			- [x] HTML selector match
			- [x] URL contains
			- [x] URL matches regex
			- [x] URL matches exactly
		2. Custom view
			- [x] Event name matches exactly
		3. Page view
			- [x] Event name matches exactly (`$pageview`)
			- [x] URL contains
			- [x] URL matches regex
			- [x] URL matches exactly
	- General filtering:
		1. Property:
			  - Types:
			      - [x] Event
				  - [x] Person
				  - [x] Element
			  - Operators:
				  - [x] Equals (or not)
				  - [x] Contains (or not)
				  - [x] Matches (or not) regex
				  - [x] Greater than/lower than
				  - [x] Set (or not)
		2. Cohort:
			- [x] Person belonging to dynamic cohort
			- [ ] Person belonging to static cohort (ClickHouse)
- [x] Insertion of new action occurrence into Postgres `posthog_action_events` table (Postgres)
- [ ] Webhooks firing
- [ ] REST hooks (Zapier) firing

## Checklist

-   [x] Jest tests


Twixes:
All matching options are functional (selector matching for the trickiest). Any optimization tips appreciated.
There's a bunch of tests but I'm still adding more for more combinations and edge case (though impossible to cover all).
Webhooks/Zapier will be a smaller followup PR.

Twixes:
Also, I'll push rollout of this to 1.27, as we want to be 100% sure this is reliable and works (a bit similar to the earlier ingestion migration), so rushing this out now wouldn't be a good idea. The action matching code will run in production but not do anything with results yet.

src/worker/ingestion/action-matcher.ts
```diff
@@ -0,0 +1,498 @@
...
+     */
+    private async checkEventAgainstCohortFilter(
+        event: PluginEvent,
+        person: Person | undefined,
+        filter: CohortPropertyFilter
+    ): Promise<boolean> {
+        if (!person?.properties) {
+            return false
+        }
+        return await this.db.doesPersonBelongToCohort(parseInt((filter.value as unknown) as any), person.id)
```
yakkomajuri, line 275: Should we maybe check for `NaN` here?

src/worker/ingestion/action-matcher.ts
```diff
@@ -0,0 +1,498 @@
...
+     */
+    private checkEventAgainstEventFilter(event: PluginEvent, filter: EventPropertyFilter): boolean {
+        return this.checkPropertiesAgainstFilter(event.properties, filter)
+    }
+
+    /**
+     * Sublevel 4 of action matching.
+     */
+    private checkEventAgainstPersonFilter(
+        event: PluginEvent,
```
yakkomajuri, line 243: event is not being used here, I guess the comment below is lingering?

src/worker/ingestion/action-matcher.ts
```diff
@@ -0,0 +1,498 @@
...
+        const foundValue = properties[filter.key]
+        let foundValueLowerCase: string // only calculated if needed for a case-insensitive operator
+
+        const requiredValueType = propertyOperatorToRequiredValueType[filter.operator]
+        if (requiredValueType && typeof foundValue !== requiredValueType) {
+            return false // MISMATCH DUE TO VALUE TYPE INCOMPATIBLE WITH OPERATOR SUPPORT
+        }
+
+        let test: (possibleValue: any) => boolean
+        switch (filter.operator) {
```
yakkomajuri, line 299: nit (personal preference?): this could be an object mapping `filter.operator` to `test` func?

src/worker/ingestion/action-matcher.ts
```diff
@@ -0,0 +1,498 @@
...
+                    // Otherwise we just move the base down
+                    baseElementIndex--
+                }
+            }
+        }
+        return wasPartMatched
+    }
+
+    private checkElementAgainstSelectorPartRequirements(element: Element, requirements: Partial<Element>): boolean {
+        if (requirements['text'] && element.text !== requirements.text) {
```
yakkomajuri, line 408: nit: in some places this check is done with `a.b` and others `a['b']`

src/worker/ingestion/action-matcher.ts
```diff
@@ -0,0 +1,498 @@
...
+    }
+}
+
+class SelectorPart {
+    directDescendant: boolean
+    uniqueOrder: number
+    requirements: Partial<Element>
+
+    constructor(tag: string, directDescendant: boolean, escapeSlashes: boolean) {
+        const SELECTOR_ATTRIBUTE_REGEX = /([a-zA-Z]*)\[(.*)=[\'|\"](.*)[\'|\"]\]/
```
yakkomajuri, line 455: A comment here would be useful

src/worker/ingestion/process-event.ts
```diff
@@ -25,13 +24,16 @@ import { DB } from '../../utils/db/db'
 import { KafkaProducerWrapper } from '../../utils/db/kafka-producer-wrapper'
 import { elementsToString, personInitialAndUTMProperties, sanitizeEventName, timeoutGuard } from '../../utils/db/utils'
 import { status } from '../../utils/status'
-import { castTimestampOrNow, filterIncrementProperties, UUID, UUIDT } from '../../utils/utils'
-import { ActionManager } from './action-manager'
+import { castTimestampOrNow, extractElements, filterIncrementProperties, UUID, UUIDT } from '../../utils/utils'
 import { PersonManager } from './person-manager'
 import { TeamManager } from './team-manager'
 
+export interface IEventX extends IEvent {
```
yakkomajuri, line 31: What's does the X mean?

yakkomajuri (commented):
This is amazing work!

Initial first pass - didn't compare anything with the Python code yet, and didn't look over everything just yet. 

This is great stuff though.

src/worker/ingestion/action-matcher.ts
Twixes: We couldn't do the optimization of only calculating `foundValueLowerCase` in case of case-insensitive operators. Well, we could, but it'd be uglier.

src/worker/ingestion/action-matcher.ts
Twixes: Yeah, leftover from some code ported from Python where dicts require brackets.

src/worker/ingestion/action-matcher.ts
```diff
@@ -0,0 +1,498 @@
...
+    /**
+     * Sublevel 2 of action matching.
+     *
+     * Return whether the event is a match for
+     * the step's "Link href equals", "Text equals" and "HTML selector matches" constraints.
+     * Step properties: `tag_name`, `text`, `href`, `selector`.
+     */
+    private checkStepElement(elements: Element[], step: ActionStep): boolean {
+        // CHECK CONDITIONS, OTHERWISE SKIPPED
+        if (step.href || step.text || (step.href && elements.length)) {
```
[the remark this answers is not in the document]
Twixes: It would but it would be kind of nonsensical anyway. 🤔 I reworked this to make sense.

src/worker/ingestion/action-matcher.ts
Twixes: What should this comment convey?

src/worker/ingestion/process-event.ts
Twixes: Renamed to `IEventMatchable`. Though still kind of an ugly way of passing this data, I will rework it to something a bit nicer

src/worker/ingestion/action-matcher.ts
```diff
@@ -0,0 +1,498 @@
...
+        if (step.url_matching && step.url) {
+            const eventUrl = event.properties?.$current_url
+            if (!eventUrl || typeof eventUrl !== 'string') {
+                return false // URL IS UNKNOWN
+            }
+            let isUrlOk: boolean
+            switch (step.url_matching) {
+                case ActionStepUrlMatching.Contains:
+                    // Simulating SQL LIKE behavior (_ = any single character, % = any zero or more characters)
+                    // TODO: reconcile syntax discrepancies between SQL and JS regex
```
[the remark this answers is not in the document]
Twixes: Actually I now added RE2, which is used by ClickHouse for regex, so no discrepancy in that regard. There's still a _possible_ discrepancy between POSIX regex used by Postgres and RE2 by ClickHouse, but that's minor enough for me to remove this comment

src/worker/ingestion/action-manager.ts
```diff
@@ -35,7 +35,19 @@ export class ActionManager {
 
     public async reloadAction(teamId: Team['id'], actionId: Action['id']): Promise<void> {
         const refetchedAction = await this.db.fetchAction(actionId)
-        const wasCachedAlready = teamId in this.actionCache && actionId in this.actionCache[teamId]
+
+        let wasCachedAlready = true
+        if (!this.actionCache) {
```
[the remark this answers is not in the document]
Twixes: Never actually

src/worker/ingestion/action-matcher.ts
```diff
@@ -0,0 +1,498 @@
...
+
+    /**
+     * Sublevel 3 of action matching.
+     */
+    private async checkEventAgainstFilter(
+        event: PluginEvent,
+        elements: Element[],
+        filter: PropertyFilter
+    ): Promise<boolean> {
+        const person = await this.db.fetchPerson(event.team_id, event.distinct_id)
```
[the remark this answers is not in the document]
Twixes: Done though indeed a minor optimization. A better one will be using person already fetched in event ingestion

src/worker/ingestion/action-matcher.ts
yakkomajuri: I think with longer regexes an example of what it should match is always useful

src/utils/db/db.ts
```diff
@@ -772,6 +783,14 @@ export class DB {
         return action
     }
 
+    public async registerActionOccurrence(actionId: Action['id'], eventId: Event['id']): Promise<void> {
+        await this.postgresQuery(
+            `INSERT INTO posthog_action_events_2 (action_id, event_id) VALUES ($1, $2)`, // TODO: remove _2
```
mariusandra, line 788: TODO

src/utils/utils.ts
```diff
@@ -606,3 +606,16 @@ export function stringClamp(value: string, def: number, min: number, max: number
     const nanToNull = (nr: number): null | number => (isNaN(nr) ? null : nr)
     return clamp(nanToNull(parseInt(value)) ?? def, min, max)
 }
+
+export function extractElements(elements: Record<string, any>[]): Element[] {
```
mariusandra, line 610: A bunch of element code is in `src/utils/db/utils.ts`. Perhaps this makes sense to put there as well?

src/worker/ingestion/action-matcher.ts
```diff
@@ -0,0 +1,490 @@
...
+            case 'event':
+                return this.checkEventAgainstEventFilter(event, filter)
+            case 'person':
+                person = await this.db.fetchPerson(event.team_id, event.distinct_id)
+                return this.checkEventAgainstPersonFilter(person, filter)
+            case 'element':
+                return this.checkEventAgainstElementFilter(elements, filter)
+            case 'cohort':
+                person = await this.db.fetchPerson(event.team_id, event.distinct_id)
+                return await this.checkEventAgainstCohortFilter(person, filter)
```
mariusandra, line 218: This `db.fetchPerson` could be fetching the same db rows multiple times per `ActionMatcher.match` in case we have a lot of steps that check user properties. 

It would be great if we could either recycle the person that we got while ingesting, or then just cache it here somehow.

mariusandra (approved):
This is some really nice code. Well done! :)

I found one TODO to fix, and have some thoughts regarding caching person. Otherwise LGTM.

src/worker/ingestion/process-event.ts
Twixes: Removed this extended interface in favor of much nicer `EventProcessingResult`.

PR #441, opened 2021-05-27
Set precedence to DATABASE_URL over POSTHOG_DB_* fields

neilkakkar:
## Changes

This reconciles precedence order to be the same as on PostHog/posthog. This is important, because otherwise cases are possible where the Django app connects to postgres without issues, but the plugin server doesn't: so events are shown on the UI, but aren't ingested properly.

Specially, have seen this happen with SSL enabled DB urls, like in this case: https://posthogusers.slack.com/archives/C01GLBKHKQT/p1621416865003200

## Checklist

-   [ ] Jest tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -105,33 +105,45 @@ jobs:
                   DATABASE_URL: 'postgres://postgres:postgres@localhost:5432/test_posthog'
                   REDIS_URL: 'redis://localhost'
               run: |
+                  echo "🔗 Linking this plugin server into posthog(master)/plugins/"
                   cd posthog/plugins
                   yarn link "@posthog/plugin-server"
                   yarn
 
-                  # start this in the background
-                  yarn start &> tmplog.txt &
+                  echo "🏁 Starting the plugin server dist/ (compiled JS) via posthog(master)/plugins/"
+                  mkdir -p tmp
+                  yarn start &> tmp/plugin-server.log &
                   pid=$!
 
+                  echo "⏳ Waiting 10 seconds to see if it runs"
                   sleep 10
 
-                  # check if it's still running, as it should be
+                  echo "🤔 Checking if it's still running, as it should be"
                   if ! kill $pid > /dev/null 2>&1; then
-                      echo "Could not send SIGTERM to process $pid" >&2
+                      echo "😵 It was not running!"
+                      echo "🪵 Here's the log:"
                       echo ""
-                      cat tmplog.txt
+                      cat tmp/plugin-server.log
                       exit 1
                   fi
 
-                  str=`cat tmplog.txt | grep "All systems go"`
+                  echo '🤔 Checking if it logged "All systems go"'
+                  str=`cat tmp/plugin-server.log | grep "All systems go"`
                   if [ ! "$str" ];then
-                      echo 'Did not find "All systems go" in plugin server log output!'
+                      sleep 5
+                      echo '😵 Did not find "All systems go" in plugin server log output!'
+                      echo "🪵 Here's the log:"
                       echo ""
-                      cat tmplog.txt
+                      cat tmp/plugin-server.log
                       exit 1
                   fi
 
-                  rm tmplog.txt
+                  sleep 5
+                  echo '✅ All systems went!'
+                  echo "🪵 Here's the complete log:"
+                  echo ""
+                  cat tmp/plugin-server.log
+                  rm -f tmp/plugin-server.log
 
     tests-postgres-1:
         name: Tests / Postgres + Redis (1)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -118,8 +118,6 @@ jobs:
                   echo "⏳ Waiting 10 seconds to see if it runs"
                   sleep 10
 
-                  cat tmp/plugin-server.log
-
                   echo "🤔 Checking if it's still running, as it should be"
                   if ! kill $pid > /dev/null 2>&1; then
                       echo "😵 It was not running!"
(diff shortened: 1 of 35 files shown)


src/utils/db/hub.ts
```diff
@@ -135,7 +135,8 @@ export async function createHub(
         timeStr ? DateTime.fromSQL(timeStr, { zone: 'utc' }).toISO() : null
     )
 
-    const postgres = createPostgresPool(serverConfig)
+    const configOrDatabaseUrl = serverConfig.DATABASE_URL ? serverConfig.DATABASE_URL : serverConfig
```
Twixes, line 138: This will always result in `DATABASE_URL` being used, because it has a non-empty default value (while `POSTHOG_DB_NAME` is null by default).

Twixes (changes requested):
This should not be needed for any setup though 🤔 `POSTHOG_DB_NAME` – AFAIK – has to be explicitly set by the user, in which case we can assume it's more deliberate than `DATABASE_URL` – which practically always has a value, because it has a default.

src/utils/db/hub.ts
neilkakkar: Oh, right, this is annoying :$ - looks like it's not required on main PostHog: https://github.com/PostHog/posthog/blob/master/posthog/settings.py#L369-L376 - and hence the precedence order there.

If we use `POSTHOG_DB_NAME`, while PostHog/posthog uses DATABASE_URL, that's.. not good.

I wonder if we can get around this by removing the default value for `DATABASE_URL` ? Does that break things / workflows elsewhere? (I'm guessing probably yes)

.. or maybe it's a documentation thing? We explicitly mention to use one or the other? Although, I'd prefer to fix this ourselves so things are consistent.

Twixes: Yeah, def not good to have a discrepancy. I suppose the fix would then be removing the default value of `DATABASE_URL` here in the non-test non-dev case, and adjusting `createPostgresPool` for same control flow as in the main repo.

src/utils/utils.ts
```diff
@@ -416,22 +416,27 @@ export function createPostgresPool(
...
                   connectionString: configOrDatabaseUrl,
               }
-            : configOrDatabaseUrl.POSTHOG_DB_NAME
+            : configOrDatabaseUrl.DATABASE_URL
             ? {
-                  database: configOrDatabaseUrl.POSTHOG_DB_NAME,
+                  connectionString: configOrDatabaseUrl.DATABASE_URL,
+              }
+            : {
+                  database: configOrDatabaseUrl.POSTHOG_DB_NAME || undefined,
```
neilkakkar, line 434: Is there a better way to convert `null` to `undefined` for types?

Twixes: You can also use `??` which is the _nullish_ coalescing operator – where `||` returns right side when left side is _falsy_, `??` returns right side only when left side is `null` or `undefined`. So depends on whether we want the plugin server to treat `''` as a valid (if empty) string value (→ `??`), or as a lack of value i.e. `undefined` (→ `||`).

Twixes (commented):
Nice, discrepancy is no more. A few extra edge cases

tests/helpers/sql.ts
```diff
@@ -33,7 +33,7 @@ export async function resetTestDatabase(
     extraRows: ExtraDatabaseRows = {}
 ): Promise<void> {
     const config = { ...defaultConfig, ...extraServerConfig }
-    const db = new Pool({ connectionString: config.DATABASE_URL })
+    const db = new Pool({ connectionString: config.DATABASE_URL || undefined })
```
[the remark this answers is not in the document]
neilkakkar: Fair

tests/helpers/graphile.ts
```diff
@@ -4,7 +4,7 @@ import { Pool } from 'pg'
 import { PluginsServerConfig } from '../../src/types'
 
 export async function resetGraphileSchema(serverConfig: PluginsServerConfig): Promise<void> {
-    const graphileUrl = serverConfig.JOB_QUEUE_GRAPHILE_URL || serverConfig.DATABASE_URL
+    const graphileUrl = serverConfig.JOB_QUEUE_GRAPHILE_URL || serverConfig.DATABASE_URL || undefined
```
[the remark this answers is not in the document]
neilkakkar: Can't be `serverConfig`, since it's directly going into the pool, how about I use the default it was supposed to be for tests? (since this is in tests?)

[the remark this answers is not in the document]
neilkakkar: Err, I meant using the non-null assertion operator

src/utils/utils.ts
neilkakkar: Will do what PostHog main does: null is not ok, falsy is ok.

tests/helpers/graphile.ts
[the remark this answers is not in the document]
Twixes: Ah, right, I missed that this is a helper for tests. In that case yep, good to just assert non-null

.github/workflows/ci.yml
```diff
@@ -118,6 +118,8 @@ jobs:
                   echo "⏳ Waiting 10 seconds to see if it runs"
                   sleep 10
 
+                  cat tmp/plugin-server.log
```
Twixes, line 121: This way we have this line once always and also a second time if the plugin server failed somehow (down below) 🤔

Twixes (approved):

.github/workflows/ci.yml
neilkakkar: <Sigh>, Thanks!

PR #450, opened 2021-06-01
Insert events and properties into enterprise definitions tables

liyiy:
## Changes

Follow up for https://github.com/PostHog/posthog/pull/4486

This PR adds the event and property definitions into the enterprise version tables.

## Checklist

-   [ ] Jest tests

### DIFF
--- a/src/worker/ingestion/team-manager.ts
+++ b/src/worker/ingestion/team-manager.ts
@@ -103,20 +103,32 @@ export class TeamManager {
         await this.cacheEventNamesAndProperties(team.id)
 
         if (!this.eventNamesCache.get(team.id)?.has(event)) {
+            const newId = new UUIDT().toString()
             await this.db.postgresQuery(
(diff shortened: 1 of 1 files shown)


src/worker/ingestion/team-manager.ts
```diff
@@ -103,21 +103,33 @@ export class TeamManager {
...
 
         if (!this.eventNamesCache.get(team.id)?.has(event)) {
+            const newId = new UUIDT().toString()
             await this.db.postgresQuery(
                 `INSERT INTO posthog_eventdefinition (id, name, volume_30_day, query_usage_30_day, team_id) VALUES ($1, $2, NULL, NULL, $3) ON CONFLICT DO NOTHING`,
-                [new UUIDT().toString(), event, team.id],
+                [newId, event, team.id],
                 'insertEventDefinition'
             )
+            await this.db.postgresQuery(
```
liyiy, line 112: I don't want to have to add definitions to an enterprise version table for non-enterprise users, but if they choose to become enterprise, they'll be missing these definitions if I don't send it along with the non-enterprise ones now, how should I go about this?

liyiy:
never mind going to implement this a different way without touching the plugin server!

PR #451, opened 2021-06-01
Clean up Jobqueues, minor fixes for S3 Queue

neilkakkar:
## Changes

Ensures S3 queue consumer runs.

Makes GraphileQueue adhere to JobQueueBase.

## Checklist

-   [x] Jest tests

### DIFF
--- a/src/main/job-queues/job-queue-base.ts
+++ b/src/main/job-queues/job-queue-base.ts
@@ -36,9 +36,11 @@ export class JobQueueBase implements JobQueue {
     startConsumer(onJob: OnJobCallback): void
     // eslint-disable-next-line @typescript-eslint/require-await
     async startConsumer(onJob: OnJobCallback): Promise<void> {
-        this.started = true
         this.onJob = onJob
-        await this.syncState()
+        if (!this.started) {
+            this.started = true
+            await this.syncState()
+        }
     }
 
     stopConsumer(): void
@@ -62,8 +64,10 @@ export class JobQueueBase implements JobQueue {
     resumeConsumer(): void
     // eslint-disable-next-line @typescript-eslint/require-await
     async resumeConsumer(): Promise<void> {
-        this.paused = false
-        await this.syncState()
+        if (this.paused) {
+            this.paused = false
+            await this.syncState()
+        }
     }
 
     // eslint-disable-next-line @typescript-eslint/require-await
@@ -77,8 +81,8 @@ export class JobQueueBase implements JobQueue {
                 clearTimeout(this.timeout)
             }
             // eslint-disable-next-line @typescript-eslint/await-thenable
-            const hadSomething = await this.readState()
-            this.timeout = setTimeout(() => this.syncState(), hadSomething ? 0 : this.intervalSeconds * 1000)
+            await this.readState()
+            this.timeout = setTimeout(() => this.syncState(), this.intervalSeconds * 1000)
         } else {
             if (this.timeout) {
                 clearTimeout(this.timeout)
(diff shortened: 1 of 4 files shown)


src/main/job-queues/job-queue-base.ts
```diff
@@ -62,8 +64,10 @@ export class JobQueueBase implements JobQueue {
     resumeConsumer(): void
     // eslint-disable-next-line @typescript-eslint/require-await
     async resumeConsumer(): Promise<void> {
```
neilkakkar, line 66: Resume is called on `drain` events emitted by Piscina, which results in multiple calls to sync state. This is okay, as long as dequeue code is idempotent (which it isn't, thus leading to race conditions).

We didn't face this with graphile because the runner object is unchanged over multiple sync state calls.

src/main/job-queues/job-queue-base.ts
```diff
@@ -77,8 +81,8 @@ export class JobQueueBase implements JobQueue {
                 clearTimeout(this.timeout)
             }
             // eslint-disable-next-line @typescript-eslint/await-thenable
-            const hadSomething = await this.readState()
-            this.timeout = setTimeout(() => this.syncState(), hadSomething ? 0 : this.intervalSeconds * 1000)
+            await this.readState()
```
neilkakkar, line 84: Given existing semantics, `readState` is supposed to finish everything required with the jobs seen so far. So, it doesn't make sense to poll again right after, if it found some jobs to process.

src/main/job-queues/redlocked/s3-queue.ts
```diff
@@ -20,6 +20,7 @@ export class S3Queue extends JobQueueBase {
         this.serverConfig = serverConfig
         this.s3Wrapper = null
         this.runner = null
```
neilkakkar, line 22: runner isn't used anymore, remove

tests/jobs.test.ts
```diff
@@ -284,5 +308,44 @@ describe('job queues', () => {
                 Key: `prefix/2020-01-01/20200101-123456.123Z-deadbeef.json.gz`,
             })
         })
+
+        test('polls for new jobs', async () => {
```
neilkakkar, line 312: these tests would fail after 60s (jest timeout) if there's no polling.

src/main/job-queues/job-queue-base.ts
mariusandra: The idea here is that if you have 10k jobs to run, you shouldn't wait 5 sec between them (or between every 100 of them) :). So if we got a job with the last run, immediately see if there's another one. Otherwise wait a few seconds and then try again.

tests/jobs.test.ts
```diff
@@ -151,6 +151,30 @@ describe('job queues', () => {
                 await waitForLogEntries(2)
                 expect(testConsole.read()).toEqual([['processEvent'], ['reply', 'runIn']])
             })
+
+            test.only('polls for jobs in future', async () => {
```
mariusandra, line 155: probably no `only` here?

tests/jobs.test.ts
```diff
@@ -151,6 +151,30 @@ describe('job queues', () => {
...
+                const DELAY = 10000 // 10s
```
mariusandra, line 156: probably OK to wait with a smaller timeout to make tests go a bit faster

mariusandra (commented):
some thoughts below, still haven't ran it locally

tests/jobs.test.ts
neilkakkar: 🤦

src/main/job-queues/job-queue-base.ts
neilkakkar: Ah, that makes sense... assuming there's batching involved. I haven't seen it so far (am I missing something?) - so didn't assume such a state was possible.

Will revert.

tests/jobs.test.ts
neilkakkar: Ha, definitely. Needed a big one only for S3.

mariusandra (approved):
These changes make sense

PR #455, opened 2021-06-02
Build healthcheck into index.js

### LINKED ISSUE #372: Official method to health check the plugin server when installed via yarn
It seems, that the existing [healthcheck.ts](https://github.com/PostHog/plugin-server/blob/master/src/healthcheck.ts)  is only exposed via the docker container defined [here](https://github.com/PostHog/plugin-server/blob/master/Dockerfile). 
The posthog main repo also exposes the plugin server and installs it via [yarn in the dockerfile](https://github.com/PostHog/posthog/blob/master/production.Dockerfile#L18).

It would be great if the package could expose a binary `healthcheck` that calls [healthcheck.ts](https://github.com/PostHog/plugin-server/blob/master/src/healthcheck.ts).

This would enable the implementation of a healthcheck when the main image is executed with [bin/plugin-server](https://github.com/PostHog/posthog/blob/master/bin/plugin-server)

Twixes:
## Changes

Closes #372. This way healthcheck is ran with the same executable as the usual mode of the plugin server: `posthog-plugin-server --healtcheck`.

## Checklist

### DIFF
--- a/src/utils/utils.ts
+++ b/src/utils/utils.ts
@@ -15,6 +15,8 @@ import { status } from './status'
 
 /** Time until autoexit (due to error) gives up on graceful exit and kills the process right away. */
 const GRACEFUL_EXIT_PERIOD_SECONDS = 5
+/** Number of Redis error events until the server is killed gracefully. */
+const REDIS_ERROR_COUNTER_LIMIT = 10
 
 export function killGracefully(): void {
     status.error('⏲', 'Shutting plugin server down gracefully with SIGTERM...')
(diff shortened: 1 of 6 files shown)


yakkomajuri (approved):

src/utils/utils.ts
```diff
@@ -386,10 +388,17 @@ export async function createRedis(serverConfig: PluginsServerConfig): Promise<Re
...
     redis
         .on('error', (error) => {
+            errorCounter++
             Sentry.captureException(error)
-            status.error('🔴', 'Redis error encountered! Trying to reconnect...\n', error)
+            if (errorCounter > REDIS_ERROR_COUNTER_LIMIT) {
+                status.error('😡', 'Redis error encountered! Enough of this, I quit!\n', error)
+                killGracefully()
+            } else {
+                status.error('🔴', 'Redis error encountered! Trying to reconnect...\n', error)
```
yakkomajuri, line 400: The only "nit" (it's not even that) I'd have is that this could tell me how many more retries to connect it'll perform. Might not even be super useful though.

Twixes: I'll table that for now

Issue #405, opened 2021-05-21
Plugin installation step


mariusandra:
Currently plugins are installed from Django. The main posthog app in [`plugin.py`](https://github.com/PostHog/posthog/blob/master/posthog/api/plugin.py). The code there either 1) gets a github repo URL, parses it, fetches a `.zip` file with the plugin code, does basic validation (version check) and stores the archive in postgres, or 2) Opens a source editor and stores its contents in postgres.

In both cases, we won't know if the plugin actually works before we try it out for real.

Instead, I'd like there to be a separate "install plugin" step and "installing..."/"updating..." states. This might mean pushing some of the logic away from Python into NodeJS (e.g. downloading the archives?) and just dispatching tasks from the plugin server, and awaiting their responses.

Eventually we the plugin server would spin up a VM for the "to install" plugin, capture its capabilities, and if all loaded well, mark the plugin as installed and ready to use. I'd argue the VM should be spin up in a separate worker pool (with just one thread) for isolation. Either via piscina or some other system. This could be the basis for plugin testing as well https://github.com/PostHog/plugin-server/issues/165

mariusandra:
This unlocks:
- Properly saving plugin capabilities: https://github.com/PostHog/plugin-server/issues/379
- Plugin source testing: https://github.com/PostHog/plugin-server/issues/165
- GeoIP download could be done in an installation step perhaps? https://github.com/PostHog/plugin-server/issues/391
- Split onEvent and other (e.g. scheduled job) plugins into separate worker pools
- Multiple JS files inside plugins https://github.com/PostHog/plugin-server/issues/367

neilkakkar:
Thinking this through: There's 2 ways for `posthog` to share data with the `plugin-server`: the postgres database, and the celery queue.

Given the discussion in PostHog/plugin-server#165 , I think we don't want to go the webserver route to open a channel of communication between these two.

Now, the question becomes how installation works. There's the UI presentation side, and the backend implementation side. I'm thinking of reusing most of what we already have, in both cases.

Here's how I imagine things would work: 
1. The `pluginConfig` gets an installation status field (that goes into the DB)
2. Python/Django does exactly what it does right now, except instead of being enabled on installation, it sets status to installing (( need to think a bit more here whether I can modify `enabled` field instead, but that feels _very_ messy)
3.  This moves the plugin into the Installed tab (name to be changed): 
[...]

Issue #165, opened 2021-02-18
Plugin Capabilities & Source Testing


mariusandra:
There are two things that could be fixed at the same time:

1. The "Plugin Source Editor" will eventually need a "test mode", which would let us run random JS code and a) verify it's not broken and b) see how it processes events. This untrusted code should be run on the plugin server and in a separate VM and possibly in a separate worker thread. How do we ask the plugin server to run these one off tests? Celery?

2. We need to cache what each plugin supports. Basically if it has support for 1) scheduled functions `runEveryX` 2) event processing functions `processEvent` / `processEventBatch`. This lets us separate VMs better in the future. For example having a separate pool of piscina workers for scheduled tasks, starting there only the VMs that are relevant. This caching could be run when a plugin is installed.

macobo:
Been thinking about this.

# Web servers

It would seem reasonable to solve it like this:

1. We set up a web server for plugins which can respond to this arbitrary code execution request. Specifically, user can upload source, test(s) and a task name and get results back. Proposal: `POST /api/test`?
2. In app, we expose the plugins API by proxying requests. Proposal: `POST /api/plugin_server/X` gets forwarded to plugin servers `/api/X`.

This web API solution can later be extended to e.g. allow [extending posthog.js via plugins](https://github.com/PostHog/posthog/issues/3504) or [UI plugins](https://github.com/PostHog/posthog/issues/2965), both of which could request javascript files from the plugin server.

The tricky part for me is whether we can reliably set up and load balance the web server across all of our deployment options:
1. For cloud, we'd need to set up an ELB and expose its URI to posthog core web. I couldn't find any terraform/other files that set up the current ELB though.
2. For docker-compose updating should be easy - the web server will be started on localhost
3. For heroku, ??? since workers can't receive web traffic
4. For cloudformation we could also set up an internal ELB and route traffic there. We'd need everyone to update their cloudformation config though, which is a PITA
5. Didn't investigate digitalocean or any other options

Because of this, the web server approach seems like it would be operationally hard to get this live and working everywhere. 

cc @fuziontech @mariusandra @Twixes - am I missing something simple here?

----

# Alternative: celery-based RPC

An alternative is to implement a celery-based RPC system, [similar to how funnels are calculated in the main codebase](https://github.com/PostHog/posthog/blob/eb7b361/posthog/api/insight.py#L203-L230).

[...]

PR #456, opened 2021-06-03
Install Plugin Step: Plugin Server side

neilkakkar:
## Changes

Part of #405 

So, this turned out to be a lot more involved than I imagined. Have decided to break it down into steps.

This is step 1: Create a new worker pool, with it's own vm/setup/everything.

PR is a proof of concept, very rough, need to clean everything up (and repeat myself less) - but wanted to get this out so you can see what's going on and comment on the approach.

For installation, our setup makes things such that there's only 1 plugin on the hub, which allows us to freely re-use other tasks as they stand. (as long as we don't mess up on reload - I'm thinking of banning reload on this thread).

Everytime a new task comes in, the install thread sets up the plugin, and returns if there's no problems. I've set things up so that this will be re-usable for #165 -> basically every time you want to test something, it begins with a fresh installation, running setup -> task -> teardown.

Some repetition is inevitable, but I think I've whittled it down to the minimum (except vm.ts, which will go away).

Also, *naming is very much a WIP too*

---

To investigate:
1. Troubles with multiple pools
    2. Pools missing logs 
2. DRY
3.

## Checklist

-   [ ] Jest tests


neilkakkar:
How I'm thinking of splitting the steps:

1. Create celery task on Plugin server to accept install commands. (this PR)
2. Update main PostHog to send install tasks on install (not upgrades - those come later, add on). Figure what to do on frontend when install fails (and how to show it)
3. Update schemas / introduce Installation upgrades and testing with different source archives. (see #405 for more)

PR #457, opened 2021-06-03
Add `url` as a library

macobo:
We can do some url-based analysis/changes based on this.

## Checklist

-   [ ] Jest tests

### DIFF
--- a/src/worker/vm/imports.ts
+++ b/src/worker/vm/imports.ts
@@ -7,12 +7,14 @@ import * as genericPool from 'generic-pool'
 import fetch from 'node-fetch'
 import * as pg from 'pg'
 import snowflake from 'snowflake-sdk'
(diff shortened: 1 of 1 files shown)


Twixes (approved):

mariusandra:
Could we also add `url` to package.json to explicitly pin it to a version? It's now installed randomly by `aws-sdk` and that could drop on any upgrade.

Twixes:
As far as I understand this is the [Node `url` module](https://nodejs.org/api/url.html), is it not? `url` from `npm` is a polyfill for browsers.

macobo:
AFAIK, that was my own takeaway as well based on https://nodejs.org/api/url.html

mariusandra:
Ah, then please ignore me :)

PR #460, opened 2021-06-03
Retry plugin initialization before punishing

### LINKED ISSUE #387: Retries when punishing plugins
New issue for thoughts and future work (retries) on merged PR https://github.com/PostHog/plugin-server/pull/380

One change (nit?) I would have done is to pass `pluginConfig` to `disablePlugin` directly, instead of via `plugin_id` and `team_id`. This would also simplify the SQL query, since we would just need to check for `id` = `pluginConfig.id`.

Regarding the points in the PR:

> Do we want to allow users to re-enable plugins that failed on setup even though they haven't changed at all? i.e. prevent a user from keeping on enabling a plugin that we know will fail.

If the API my plugin talks to fails for reasons outside of my control (or in control, but external, e.g. not enough credits for an API), I'd like to re-enable the plugin without needing to reinstall it :). So yes.

> Do we want a retry strategy for setupPlugin? How many retries?

I think we do. With the merged implementation we will accidentally punish good plugins. Knowing how the system works after that PR, I would be very careful with making any network requests inside `setupPlugin`, as any accidental network glitch (api is down for 5min) will disable the plugin completely. Basically, if my plugin won't start because "Your account is suspended due to lack of payment method.", I'd like it to try again at a later date. However if my plugin won't start because "invalid api key", I'd like it to be totally disabled. 

I think the answer is to put this on plugin authors. You either throw a `RetryError` (will be retried) or anything else (will be disabled). That's what Segment does at least.

yakkomajuri:
## Changes

Retrying plugin VM initialization with exponential backoff before ultimately disabling the plugin.

Resolves #387 

The approach here is simply to reset the `resolveInternalVm` promise with timeouts using an exponential backoff mechanism such that the plugin will continue to be enabled, but the user will see an error, and its methods will return null (i.e. none of its methods will run or affect anything) for the whole time while we retry until that promise gets resolved to a `vm` or we cap out at 15 retries.

I followed the suggestion in #387 but I'm thinking ~24h (15 retries) is a bit much - thoughts?

## Checklist

-   [ ] Jest tests


src/worker/vm/lazy.ts
```diff
@@ -14,38 +15,72 @@ import { clearError, processError } from '../../utils/db/error'
...
-    resolveInternalVm: Promise<PluginConfigVMResponse | null>
+    resolveInternalVm!: Promise<PluginConfigVMResponse | null>
+    totalAttemptsToInitialize: number
 
     constructor() {
+        this.totalAttemptsToInitialize = 0
+        this.initVM()
+    }
+
+    private initVM() {
```
Twixes, line 29: Nit: let's stick to `Vm` instead of `VM`

src/worker/vm/lazy.ts
```diff
@@ -14,38 +15,72 @@ import { clearError, processError } from '../../utils/db/error'
...
+                            `Failed to load ${logInfo}. Retrying to initialize it in ${nextRetrySeconds}.`
+                        )
+                        await createPluginLogEntry(
+                            `Plugin failed to load but its initialization will be retried in ${nextRetrySeconds} (instance ID ${hub.instanceId}).`,
+                            PluginLogEntryType.Error
+                        )
+                        setTimeout(() => {
+                            this.initVM()
+                            void this.initialize?.(hub, pluginConfig, indexJs, logInfo)
+                        }, nextRetryMs)
```
Twixes, line 69: Q: what if the plugin is reloaded (e.g. after it's updated) before this timeout callback runs?

src/worker/vm/lazy.ts
```diff
@@ -14,38 +15,72 @@ import { clearError, processError } from '../../utils/db/error'
...
-                    await hub.db.createPluginLogEntry(
-                        pluginConfig,
-                        PluginLogEntrySource.System,
-                        PluginLogEntryType.Error,
-                        `Plugin failed to load and was disabled (instance ID ${hub.instanceId}).`,
-                        hub.instanceId
+                    const isRetryError = error instanceof RetryError
+                    status.warn('⚠️', error.message)
+                    if (isRetryError && this.totalAttemptsToInitialize < 15) {
+                        const nextRetryMs = 2 ** this.totalAttemptsToInitialize * 3000
```
Twixes, line 56: Suggestion: `INITIALIZATION_RETRY_MULTIPLIER ** (this.totalAttemptsToInitialize - 1) * INITIALIZATION_RETRY_BASE_MS`. That way we'd have the initial retry timeout and its multiplier plainly in constants, as right know it's not obvious that the first retry will be 6000 ms later.

src/worker/vm/lazy.ts
```diff
@@ -14,38 +15,72 @@ import { clearError, processError } from '../../utils/db/error'
...
                 } catch (error) {
-                    await hub.db.createPluginLogEntry(
-                        pluginConfig,
-                        PluginLogEntrySource.System,
-                        PluginLogEntryType.Error,
-                        `Plugin failed to load and was disabled (instance ID ${hub.instanceId}).`,
-                        hub.instanceId
+                    const isRetryError = error instanceof RetryError
+                    status.warn('⚠️', error.message)
+                    if (isRetryError && this.totalAttemptsToInitialize < 15) {
```
Twixes, line 55: Let's put the maximum in number of attempts in a const.

src/worker/vm/lazy.ts
yakkomajuri: Ah yes agreed

src/worker/vm/lazy.ts
yakkomajuri: It wouldn't impact the plugin because a new VM is created on reload anyway. 

However, we may indeed end up with a loose timeout. I'll store the timeout and clear it on teardown

yakkomajuri:
Should I merge? @Twixes

Twixes (approved):

PR #462, opened 2021-06-04
$set precedence

mariusandra:
## Changes

- Fixes https://github.com/PostHog/posthog/issues/4592 , more context in the issue
- Previously if you had $set in both `event.$set` and `event.properties.$set` we would override the properties. Now we merge them.

## Checklist

-   [ ] Jest tests

### DIFF
--- a/src/worker/piscina.js
+++ b/src/worker/piscina.js
@@ -6,12 +6,47 @@ if (isMainThread) {
     const { createConfig } = require('./config')
     module.exports = {
         makePiscina: (serverConfig) => {
-            const piscina = new Piscina(createConfig(serverConfig, __filename))
-            piscina.on('error', (error) => {
-                Sentry.captureException(error)
-                console.error('⚠️', 'Piscina worker thread error:\n', error)
-            })
-            return piscina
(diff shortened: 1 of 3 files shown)


src/worker/piscina.js
```diff
@@ -6,12 +6,47 @@ if (isMainThread) {
     const { createConfig } = require('./config')
     module.exports = {
         makePiscina: (serverConfig) => {
-            const piscina = new Piscina(createConfig(serverConfig, __filename))
-            piscina.on('error', (error) => {
-                Sentry.captureException(error)
-                console.error('⚠️', 'Piscina worker thread error:\n', error)
-            })
-            return piscina
+            const config = createConfig(serverConfig, __filename)
```
neilkakkar, line 9: Can we remove piscina concurrency from this PR? 😋

mariusandra: whoopie :D

neilkakkar (approved):

yakkomajuri (approved):
Some test failures I haven't looked into - otherwise LGTM

PR #465, opened 2021-06-07
Make timeout errors more understandable inside workers

neilkakkar:
## Changes

Add a caller parameter to async guard

This will soon help make it easier to understand what's going on with errors like these: https://sentry.io/organizations/posthog/issues/2358370433/?project=5592816

... and what went wrong with Bigquery in https://github.com/PostHog/bigquery-plugin/pull/9 

## Checklist

-   [x] Jest tests


src/worker/ingestion/action-manager.ts
```diff
@@ -37,12 +37,6 @@ export class ActionManager {
         const refetchedAction = await this.db.fetchAction(actionId)
         const wasCachedAlready = teamId in this.actionCache && actionId in this.actionCache[teamId]
         if (refetchedAction) {
-            status.info(
-                '🍿',
-                wasCachedAlready
-                    ? `Refetched action ID ${actionId} (team ID ${teamId}) from DB`
-                    : `Fetched new action ID ${actionId} (team ID ${teamId}) from DB`
-            )
```
mariusandra, line 45: 💯

Twixes: I think we should leave this in, but at a reduced `debug` level, which would only be shown with `LogLevel.Debug` so that normally there's no spam

neilkakkar: Sounds good to me

neilkakkar: Actually, involves adding debug to status, so I'll pull it out into a separate PR instead.

Twixes (approved):

PR #470, opened 2021-06-10
Action matching polish

### LINKED ISSUE #467: Add more action matching tests
Action matching from #436 does have a bunch of tests, but they don't yet cover all options, let alone combinations of them. Coverage should be higher before production rollout.

Twixes:
## Changes

Part of #467.

## Checklist

-   [x] Jest tests


Twixes:
More tests to come before we leave the dry run phase, but this already improves a few cases and improves confidence a ton, so merging.

PR #472, opened 2021-06-11
Reduce `MAX_SETUP_RETRIES` from 15 to 10

Twixes:
## Changes

There is a slight problem with `setupPlugin` retries – the number of retries is not persisted across plugin server restarts, so when the plugin server is restarted, an erroneous plugin is loaded from retry 0 again… This can be noticed in the wild as below:
reretrying
That's extremely likely to happen currently, because with current `INITIALIZATION_RETRY_x` settings and `MAX_SETUP_RETRIES` set to 15, the retries go over 98 301 seconds – more than 27 hours. Any push to `posthog` master in these 27 hours effectively resets the retry counter, meaning the plugin never gets disabled (maybe on the weekend).
With `MAX_SETUP_RETRIES` set to 10, we'll only retry for 3069 seconds – less than an hour – which is much safer.
A _better_ solution would be to actually persist the number of retries across restarts.

## Checklist

-   [ ] Jest tests


yakkomajuri:
Yup, definitely agree with this!

yakkomajuri:
Tests need updating

PR #474, opened 2021-06-14
Handle race condition in merging person

### LINKED ISSUE #458: `NaN` issue with `posthog_person`
Sentry Issue: [PLUGIN-SERVER-2A](https://sentry.io/organizations/posthog/issues/2225546229/?referrer=github_integration)

```
error: update or delete on table "posthog_person" violates foreign key constraint "posthog_persondistin_person_id_5d655bba_fk_posthog_p" on table "posthog_persondistinctid"
  File "/code/plugins/node_modules/pg/node_modules/pg-protocol/dist/parser.js", line 278, in Parser.parseErrorMessage
    : new messages_1.DatabaseError(messageValue, length, name);
  File "/code/plugins/node_modules/pg/node_modules/pg-protocol/dist/parser.js", line 126, in Parser.handlePacket
    return this.parseErrorMessage(offset, length, bytes, "error" /* error */);
  File "/code/plugins/node_modules/pg/node_modules/pg-protocol/dist/parser.js", line 39, in Parser.parse
    const message = this.handlePacket(offset + HEADER_LENGTH, code, length, this.buffer);
  File "/code/plugins/node_modules/pg/node_modules/pg-protocol/dist/index.js", line 10, in Socket.<anonymous>
    stream.on('data', (buffer) => parser.parse(buffer, callback));
  File "events.js", line 376, in Socket.emit
...
(5 additional frame(s) were not displayed)
```

Twixes:
## Changes

This is a way to handle a race condition that we've been seeing in merging people, affecting data integrity - dangling persons being left instead of deleted after merging. Closes #458.

## Checklist

-   [ ] Jest tests


src/worker/ingestion/process-event.ts
```diff
@@ -343,26 +345,48 @@ export class EventsProcessor {
...
+            while (true) {
+                const otherPersonDistinctIds: PersonDistinctId[] = (
+                    await this.db.postgresQuery(
+                        'SELECT * FROM posthog_persondistinctid WHERE person_id = $1 AND team_id = $2',
+                        [otherPerson.id, mergeInto.team_id],
+                        'otherPersonDistinctIds'
+                    )
+                ).rows
+                for (const personDistinctId of otherPersonDistinctIds) {
+                    await this.db.moveDistinctId(otherPerson, personDistinctId, mergeInto)
```
neilkakkar, line 370: I'm a bit wary of this. With retries now, `moveDistinctId` is called multiple times with the same `otherPerson`, `personDistinctId`. (more plausible when messages are sent to kafka, instead of directly operating on the db, since then the first postgres query returns the same rows).

This might lead to new race conditions where follow-up moves are overridden by this extra move. I'm not sure if this can actually happen, and it seems less likely than the existing race condition, but still wanted to raise it.

Twixes: If I understood what you mean correctly, this shouldn't be a problem, because `moveDistinctId` wouldn't be called multiple times with the same `personDistinctId`  – the retry's `otherPersonDistinctIds` won't contain already moved IDs, as Postgres is the source of truth in all these queries, and we only need to move the IDs not moved on previous iteration due to being added after previous iteration started. Did I miss something?

neilkakkar: Ah, I see, I misread the code. When we send kafka messages, we update Postgres *before* sending these messages. This makes sense, there shouldn't be a problem here then!

neilkakkar (approved):

PR #476, opened 2021-06-14
Format CH person timestamp with only second precision everywhere

Twixes:
## Changes

https://github.com/PostHog/posthog/pull/4734 for the plugin server.

## Checklist

-   [ ] Jest tests


mariusandra:
I'd perhaps skip this...

If we insert directly into clickhouse, it has to be 1:1 what's in the DB, like in https://github.com/PostHog/posthog/pull/4734 with `sync_execute`.
If the data passes through Kafka, it strips out the milliseconds itself.

It's best if we could just fix the problem at the root: https://github.com/PostHog/posthog/issues/4260

Twixes:
Well, just for consistency good to have all `KAFKA_PERSON` produces here the same way (one uses second, two millisecond precision now). Overall I agree that this must be fixed at the root eventually, with the right precision in the CH schema.

PR #479, opened 2021-06-16
Fire webhooks from internal action matching results

### LINKED ISSUE #459: Fire core webhooks and Zapier triggers from plugin server
Direct followup to #235. Requirement of action matching in plugin server rollout.

Twixes:
## Changes

Resolves #459. Webhooks fired directly here instead of in Python with Celery.

## Checklist

-   [x] Jest tests

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -15,6 +15,7 @@ import { InternalMetrics } from './utils/internal-metrics'
 import { UUID } from './utils/utils'
 import { ActionManager } from './worker/ingestion/action-manager'
 import { ActionMatcher } from './worker/ingestion/action-matcher'
+import { OrganizationManager } from './worker/ingestion/organization-manager'
 import { EventsProcessor } from './worker/ingestion/process-event'
 import { TeamManager } from './worker/ingestion/team-manager'
 import { LazyPluginVM } from './worker/vm/lazy'
@@ -120,6 +121,7 @@ export interface Hub extends PluginsServerConfig {
     pluginConfigSecretLookup: Map<string, PluginConfigId>
     // tools
     teamManager: TeamManager
+    organizationManager: OrganizationManager
     actionManager: ActionManager
     actionMatcher: ActionMatcher
     eventsProcessor: EventsProcessor
@@ -182,12 +184,11 @@ export type JobQueueExport = {
     getQueue: (serverConfig: PluginsServerConfig) => JobQueue
 }
 
-export type PluginId = number
-export type PluginConfigId = number
-export type TeamId = number
-
+export type PluginId = Plugin['id']
+export type PluginConfigId = PluginConfig['id']
+export type TeamId = Team['id']
 export interface Plugin {
-    id: PluginId
+    id: number
     organization_id: string
     name: string
     plugin_type: 'local' | 'respository' | 'custom' | 'source'
@@ -214,7 +215,7 @@ export interface PluginCapabilities {
 }
 
 export interface PluginConfig {
-    id: PluginConfigId
+    id: number
     team_id: TeamId
     plugin?: Plugin
     plugin_id: PluginId
@@ -492,6 +493,18 @@ export interface CohortPeople {
     id: number
     cohort_id: number
     person_id: number
+}
+
+/** Usable Hook model. */
+export interface Hook {
+    id: number
+    team_id: number
+    user_id: number
+    resource_id: number | null
+    event: string
+    target: string
+    created: string
+    updated: string
 }
 
 /** Sync with posthog/frontend/src/types.ts */
--- a/src/types.ts
+++ b/src/types.ts
@@ -16,6 +16,7 @@ import { UUID } from './utils/utils'
 import { ActionManager } from './worker/ingestion/action-manager'
 import { ActionMatcher } from './worker/ingestion/action-matcher'
 import { EventsProcessor } from './worker/ingestion/process-event'
+import { TeamManager } from './worker/ingestion/team-manager'
 import { LazyPluginVM } from './worker/vm/lazy'
 
 export enum LogLevel {
@@ -118,6 +119,7 @@ export interface Hub extends PluginsServerConfig {
     pluginConfigSecrets: Map<PluginConfigId, string>
     pluginConfigSecretLookup: Map<string, PluginConfigId>
     // tools
+    teamManager: TeamManager
     actionManager: ActionManager
     actionMatcher: ActionMatcher
     eventsProcessor: EventsProcessor
--- a/src/types.ts
+++ b/src/types.ts
@@ -603,6 +603,13 @@ export interface Action extends RawAction {
     steps: ActionStep[]
 }
 
+/** Action<>Event mapping row. */
+export interface ActionEventPair {
+    id: number
+    action_id: Action['id']
+    event_id: Event['id']
+}
+
 export interface SessionRecordingEvent {
     uuid: string
     timestamp: string
--- a/src/types.ts
+++ b/src/types.ts
@@ -15,7 +15,7 @@ import { InternalMetrics } from './utils/internal-metrics'
 import { UUID } from './utils/utils'
 import { ActionManager } from './worker/ingestion/action-manager'
 import { ActionMatcher } from './worker/ingestion/action-matcher'
-import { HookCannon } from './worker/ingestion/hooks'
+import { HookCommander } from './worker/ingestion/hooks'
 import { OrganizationManager } from './worker/ingestion/organization-manager'
 import { EventsProcessor } from './worker/ingestion/process-event'
 import { TeamManager } from './worker/ingestion/team-manager'
@@ -125,7 +125,7 @@ export interface Hub extends PluginsServerConfig {
     organizationManager: OrganizationManager
     actionManager: ActionManager
     actionMatcher: ActionMatcher
-    hookCannon: HookCannon
+    hookCannon: HookCommander
     eventsProcessor: EventsProcessor
     jobQueueManager: JobQueueManager
     // diagnostics
--- a/src/types.ts
+++ b/src/types.ts
@@ -15,6 +15,7 @@ import { InternalMetrics } from './utils/internal-metrics'
 import { UUID } from './utils/utils'
 import { ActionManager } from './worker/ingestion/action-manager'
 import { ActionMatcher } from './worker/ingestion/action-matcher'
+import { HookCannon } from './worker/ingestion/hooks'
 import { OrganizationManager } from './worker/ingestion/organization-manager'
 import { EventsProcessor } from './worker/ingestion/process-event'
 import { TeamManager } from './worker/ingestion/team-manager'
@@ -124,6 +125,7 @@ export interface Hub extends PluginsServerConfig {
     organizationManager: OrganizationManager
     actionManager: ActionManager
     actionMatcher: ActionMatcher
+    hookCannon: HookCannon
     eventsProcessor: EventsProcessor
     jobQueueManager: JobQueueManager
     // diagnostics
--- a/src/utils/utils.ts
+++ b/src/utils/utils.ts
@@ -10,7 +10,7 @@ import { Readable } from 'stream'
 import * as tar from 'tar-stream'
 import * as zlib from 'zlib'
 
-import { Element, LogLevel, Plugin, PluginConfigId, PluginsServerConfig, TimestampFormat } from '../types'
+import { LogLevel, Plugin, PluginConfigId, PluginsServerConfig, TimestampFormat } from '../types'
 import { status } from './status'
 
 /** Time until autoexit (due to error) gives up on graceful exit and kills the process right away. */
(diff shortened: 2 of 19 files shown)


src/worker/ingestion/hooks.ts
```diff
@@ -0,0 +1,225 @@
...
+
+import { Action, Person } from '../../types'
+import { DB } from '../../utils/db/db'
+import { stringify } from '../../utils/utils'
+import { OrganizationManager } from './organization-manager'
+import { TeamManager } from './team-manager'
+
+export enum WebhookType {
+    Slack = 'slack',
+    Discord = 'discord',
```
neilkakkar, line 15: Is discord a new type? Don't see it here: https://github.com/PostHog/posthog/blob/master/posthog/tasks/webhooks.py#L110-L113

Twixes: Kind of, we don't need to explicitly handle it in control flows, but just to get things more in order (since we do mention it on the Project Settings page) I added detection of it. Also because that's what I use for testing all this webhook functionality :)

src/worker/ingestion/hooks.ts
```diff
@@ -0,0 +1,225 @@
...
+    person: Person | undefined,
+    siteUrl: string,
+    webhookType: WebhookType,
+    tokenParts: string[]
+): [string, string] {
+    let text = ''
+    let markdown = ''
+
+    if (tokenParts[0] === 'user') {
+        if (tokenParts[1] == 'name') {
```
neilkakkar, line 81:
```suggestion
        if (tokenParts[1] === 'name') {
```

and a few below, too! Python->JS conversion can get so annoying 😂

src/worker/ingestion/hooks.ts
```diff
@@ -0,0 +1,225 @@
...
+            ;[text, markdown] = getUserDetails(event, person, siteUrl, webhookType)
+        } else {
+            const propertyName = `$${tokenParts[1]}`
+            const property = event.properties?.[propertyName]
+            text = markdown = stringify(property)
```
neilkakkar, line 86: I'm not sure multiple assignment is equivalent in JS & Python 👀 : https://stackoverflow.com/questions/1758576/multiple-left-hand-assignment-with-javascript

But since `text` and `markdown` are already defined, this should be the same?

src/worker/ingestion/hooks.ts
```diff
@@ -0,0 +1,225 @@
...
+    } catch (error) {
+        const [actionName, actionMarkdown] = getActionDetails(action, siteUrl, webhookType)
+        messageText = `⚠ Error: There are one or more formatting errors in the message template for action "${actionName}".`
+        messageMarkdown = `*⚠ Error: There are one or more formatting errors in the message template for action "${actionMarkdown}".*`
+    }
+
+    return [messageText, messageMarkdown]
+}
+
+export class HookCannon {
```
neilkakkar, line 140: what is `Cannon` supposed to imply here? (a bit lost with the naming 😅 )

Is it because it finds **and fires** hooks?

src/worker/ingestion/hooks.ts
```diff
@@ -0,0 +1,225 @@
...
+        if (!team) {
+            return
+        }
+
+        const webhookUrl = team.slack_incoming_webhook
+        const organization = await this.organizationManager.fetchOrganization(team.organization_id)
+
+        if (webhookUrl) {
+            const webhookRequests = actionMatches
+                .filter((action) => action.post_to_slack)
```
neilkakkar, line 174: I'm a bit confused here. It seems like `slack_incoming_webhook` is used for teams and discord too?

But then, if I filter on `post_to_slack`, is that set to `True` for teams and discord too?

Because otherwise, we aren't posting webhooks for teams / discord ?

tests/shared/process-event.ts
```diff
@@ -235,9 +235,9 @@ export const createProcessEventTests = (
         )
 
         if (database === 'clickhouse') {
-            expect(queryCounter).toBe(11 + 14 /* event & prop definitions */)
+            expect(queryCounter).toBe(10 + 14 /* event & prop definitions */)
```
neilkakkar, line 238: 👀 

which query doesn't happen anymore?

src/worker/ingestion/hooks.ts
Twixes: Not sure I understand, do you mean the specific services Slack, Teams, Discord? You should completely disregard the `slack` part in `slack_incoming_webhook` and `post_to_slack` names, that's only left over from ole days when exclusively Slack was supported.

tests/worker/ingestion/organization-manager.test.ts
```diff
@@ -0,0 +1,55 @@
...
+            await hub.db.postgresQuery("UPDATE posthog_organization SET name = 'Updated Name!'", undefined, 'testTag')
+
+            mocked(hub.db.postgresQuery).mockClear()
+
+            organization = await organizationManager.fetchOrganization(commonOrganizationId)
+
+            expect(organization!.name).toEqual('TEST ORG')
+            expect(hub.db.postgresQuery).toHaveBeenCalledTimes(0)
+
+            jest.spyOn(global.Date, 'now').mockImplementation(() => new Date('2020-02-27 11:00:36').getTime())
```
neilkakkar, line 43: Is this testing the cache TTL ?

neilkakkar (commented):
A couple of questions, but otherwise looks good to me!

src/worker/ingestion/hooks.ts
Twixes: Literally yes, it definitely may be too "fun" LOL. Feel free to propose something clearer

tests/worker/ingestion/organization-manager.test.ts
Twixes: Yes, this whole `organization-manager` thing is almost totally copied from TeamManager code, because it's very much alike

src/worker/ingestion/hooks.ts
neilkakkar: As someone who called a thread-killer the `Terminator`, yes, I'm very happy with a hook cannon 😂 .

Although `hookCannon.findAndFireHooks` feels like it's a `HookCannonAndLogistician`.

To make things clearer, perhaps a `HookManager`? It seems to fit in well with our other managers, and performs a similar role to the others (although they fetch, while this one fires)

src/types.ts
```diff
@@ -15,7 +15,10 @@ import { InternalMetrics } from './utils/internal-metrics'
 import { UUID } from './utils/utils'
 import { ActionManager } from './worker/ingestion/action-manager'
 import { ActionMatcher } from './worker/ingestion/action-matcher'
+import { HookCannon } from './worker/ingestion/hooks'
```
yakkomajuri, line 18: :D

src/utils/utils.ts
```diff
@@ -620,3 +620,15 @@ export function stringClamp(value: string, def: number, min: number, max: number
     const nanToNull = (nr: number): null | number => (isNaN(nr) ? null : nr)
     return clamp(nanToNull(parseInt(value)) ?? def, min, max)
 }
+
+// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
+export function stringify(value: any): string {
+    switch (typeof value) {
+        case 'string':
+            return value
+        case 'undefined':
```
yakkomajuri, line 629: I think `JSON.stringify(undefined)` returns "undefined" already

src/worker/ingestion/hooks.ts
```diff
@@ -0,0 +1,225 @@
...
+
+export function determineWebhookType(url: string): WebhookType {
+    url = url.toLowerCase()
+    if (url.includes('slack.com')) {
+        return WebhookType.Slack
+    }
+    if (url.includes('discord.com')) {
+        return WebhookType.Discord
+    }
+    return WebhookType.Teams
```
yakkomajuri, line 27: Is this a good approach? Probably legacy so should be fine, but while this might not be harmful, I think I'd like some explicit Teams checking with an additional unknown/unspecified type, just for clarity.

I regularly use https://webhook.site/ for example

src/worker/ingestion/hooks.ts
```diff
@@ -0,0 +1,225 @@
...
+        userMarkdown = `[${userName}](${siteUrl}/person/${event.distinct_id})`
+    }
+    return [userName, userMarkdown]
+}
+
+export function getActionDetails(action: Action, siteUrl: string, webhookType: WebhookType): [string, string] {
+    const actionName = stringify(action.name)
+    let actionMarkdown: string
+    if (webhookType === WebhookType.Slack) {
+        actionMarkdown = `<${siteUrl}/action/${action.id}|${action.name}>`
```
yakkomajuri, line 53: maybe use `actionName` instead of `action.name` here too? (and below)

src/worker/ingestion/hooks.ts
```diff
@@ -0,0 +1,225 @@
...
+    } else {
+        actionMarkdown = `[${action.name}](${siteUrl}/action/${action.id})`
+    }
+    return [actionName, actionMarkdown]
+}
+
+export function getTokens(messageFormat: string): [string[], string] {
+    const matchedTokens = messageFormat.match(/(?<=\[)(.*?)(?=\])/g) || []
```
yakkomajuri, line 61: nit: as you know, I always appreciate a regex match example in a comment

src/worker/ingestion/hooks.ts
yakkomajuri: I was aware of this legacy naming, but I guess worth adding in a little comment for people who are new to the codebase so they don't get confused.

Even better would be refactoring it :D

src/worker/ingestion/hooks.ts
Twixes: :+1: Rearranged to avoid weird multiple assigment

src/worker/ingestion/organization-manager.ts
```diff
@@ -0,0 +1,47 @@
...
+
+    public async fetchOrganization(organizationId: RawOrganization['id']): Promise<RawOrganization | null> {
+        const cachedOrganization = this.getByAge(this.organizationCache, organizationId)
+        if (cachedOrganization) {
+            return cachedOrganization
+        }
+
+        const timeout = timeoutGuard(`Still running "fetchOrganization". Timeout warning after 30 sec!`)
+        try {
+            const organizationQueryResult = await this.db.postgresQuery(
```
yakkomajuri, line 24: genuine q: we have SQL queries spread in multiple areas of the codebase, `db`, `sql`, and individual queries in some files here and there. do we have any sort of convention for where they should be?

src/worker/ingestion/hooks.ts
Twixes: Renamed to, drum roll, `HookCommander`

yakkomajuri (commented):
didn't run it yet but mostly looks good - let me try it out

src/utils/utils.ts
Twixes: It in fact returns `undefined` instead of `"undefined"`, which caught me off guard for a bit 😅

src/worker/ingestion/hooks.ts
Twixes: Well, Teams is just the default behavior this way. I don't know what webhook.site accepts and I don't want to support it, so Teams format it is.

src/utils/utils.ts
yakkomajuri: ha, damn

src/worker/ingestion/hooks.ts
yakkomajuri: That makes sense yeah - webhook.site is just a webhook dump. 

When I added this comment I didn't see what the type was being used for, but yeah, if it's just formatting the message/payload, that's perfectly fine

src/worker/ingestion/organization-manager.ts
Twixes: I think for some more bespoke stuff (like `TeamManager.updateEventNamesAndProperties`), it doesn't make sense to make it common via `DB`, BUT thanks for bringing this line up – I updated it to use `DB.fetchOrganization` (and similarly in `TeamManager` with `fetchTeam`)

tests/shared/process-event.ts
Twixes: We don't do `shouldSendWebhooks` anymore 🎉

src/worker/ingestion/hooks.ts
neilkakkar: Ten-Hut!

yakkomajuri:
Ran this locally. Action calculated fine the first time and sent the relevant webhooks.

Second time though the UI told me the action was recalculated but the events didn't update and I didn't get the webhooks.

Got some errors for `posthog`:
```
ConnectionResetError: [Errno 54] Connection reset by peer
```

Don't have time to investigate, could be a fluke - just putting it out here for context.

Twixes:
What do you mean by "the events didn't update and I didn't get the webhooks"? You shouldn't get webhooks on action recalculation with this, the way Django does resend webhooks on action recalculation is unintended behavior.

yakkomajuri:
Right yeah, ofc - webhooks should fire as events come in. I guess the correlation I saw was what got me back into the old thinking model. Either way, the action didn't recalculate a second time and I didn't get webhooks for all events that were sent in :thinking:

yakkomajuri:
I'd say feel free to ignore though if this works for you - I did indeed get webhooks so the system is working

Twixes:
Hm, all OK for me, recalculation should definitely not be affected because that's Django and not touched by https://github.com/PostHog/posthog/pull/4794. I think we should be safe to go ahead

PR #480, opened 2021-06-17
Plugin Metrics

yakkomajuri:
# EDIT: Description outdated. See the comment below.

## Disclaimer

This is **very much a WIP**, which I'm only submitting based on our "principle" of a PR a day, as well as to support the discussion in https://github.com/PostHog/posthog/issues/4577. It also depends on https://github.com/PostHog/posthog/pull/4783

### Overview

Here's a graph built manually based on these changes (the goal is to build this automatically in a Plugin Metrics dashboard like in /instance/status):

Screenshot 2021-06-17 at 19 10 06

So instead of `metric_names` (see https://github.com/PostHog/posthog/issues/4577), I decided to go with `metrics`, which is an object exported by plugin devs containing a relation between a metric name and the type of mathematical operation PostHog should perform on that metric when displaying the graph. 

A bit verbose, yes - but quite powerful as a foundation for metrics that may come down the line. Either way, for functions like `exportEvents`, we'll be deriving the metrics anyway.

### API

Jobs-like. Here's what it might look like once everything is implemented:

```js
export const metrics = {
    eventsSeen: 'sum',
    somethingElse: 'max'
}

export function onEvent(event, { metrics }) {
    metrics.eventsSeen.increment(1)
    metrics['something-else'].update(100)
}
```

### Threads and metrics

As of now, this very WIP approach sets up a job in each thread to periodically send the metrics via an event. This is fine for increment operations using PostHog's `Sum` (which is what `exportEvents` needs), but is limiting for a lot of other things. Just as an experiment, to capture events from each thread, I actually used the dummy `posthog` we pass to the plugins.

Ultimately, if we want to provide people with the ability to do things like update a metric and have its latest value on a graph, we'll probably need to go through something like Redis and use a job in the main thread to regularly pull the values from there and capture an event. We could also consider sending an event per metric update, but I think that's just overkill.

### What now?

Next I'll give some thought/do some research into the types of graphs that plugin devs might want to display, as well as consider the Redis approach. Also need to update the metrics for `exportEvents` since right now each new `RetryError` will increment the counter whereas it still relates to only one event.


yakkomajuri:
Alright, so this should be reviewable now. There were quite a few tradeoffs to consider so I spent a reasonable amount of time thinking about them and optimizing things around that.

## What this allows for

This PR introduces both automatic metrics for `exportEvents` and a metrics API that plugin devs have access to (phases 1 and 2 from https://github.com/PostHog/posthog/issues/4577)

Here's how they work:

### exportEvents

```js
function exportEvents() {}
/* boom! you have the following metrics: 
 - events_seen
 - events_delivered_successfully 
 - other_errors
 - retry_errors
 - undelivered_events
*/
```

### Metrics API

```js
export const metrics = {
    'metric1': 'sum',
    'metric2': 'max',
    'metric3': 'min'
}

export function processEvent(event, { metrics }) {
    metrics.metric1.increment(100) // sum is now 100
    metrics.metric1.increment(10) // sum is now 110
    metrics.metric1.increment(-10) // sum is now 100
    metrics['metric2'].max(5) // max is now 5
    metrics['metric2'].max(10) // max is now 10
    metrics['metric3'].min(4) // min is now 4
    metrics['metric3'].min(1) // max is now 1
}
```

## Why only sum, max, and min?

I spent quite a bit of time thinking about metrics a plugin might want to set, and then how we could achieve them. The problem was that some metrics were undoable or would encourage plugin devs to shoot themselves in the foot.

Take `avg` for example - this would require one event per metric method call, so this is undoable with batched updates and events every 30min). 

Or a `metrics.myMetric.updateValue` method  - with the metrics being tallied up in the workers, the only way to get updates would be a system with some cross-thread source of truth, like a Redis-based one I mentioned. The problem is, events may arrive out of order, so even if we implemented a Redis system that ensured the thread-safe updates, the mental model we'd be passing down to plugin devs is that `update` would always give you "the latest state of affairs", which wouldn't be true if your metric relies on events and those come out of order. 

Hence, keeping it simple for now: sum, max, and min. This already allows for:

### sum

- Everything `exportEvents` does
- Counting failures and successes
- Counting "events not following schema" (schema enforcer plugin)
- Counting events processed or not processed by a given mechanism 

### min/max

- Time taken for request to third party API to complete (e.g. export plugins)
- Time taken to complete some processing (e.g. property flattener)
- Max/min length of some property being processed
- Max/min number of props in an event

## Implementation

Before `setupPlugin` runs, metrics are setup, so that they can be used in `setupPlugin`. This is needed so that some validation can be done on metric calls. I took this approach to ensure errors will be thrown during development if the metrics API is being used, to prevent plugins from being released with incorrect metrics. 

In the background, metrics specified are persisted on the database, allowing us to use those to create graphs (phase 3).

A job is scheduled per worker (at the `createHub` step) to send the metrics to PostHog via `DummyPosthog` (this was actually not a bad idea) every 30 min. This means the maximum possible number of events a plugin using metrics emits is 1 event per worker every 30min (note: it's not necessarily the case that an event will be emitted per worker every time).

src/utils/plugin-metrics.ts
```diff
@@ -0,0 +1,74 @@
...
+
+interface UpdateMetricPayload {
+    pluginConfig: PluginConfig
+    metricOperation: MetricMathOperations
+    metricName: string
+    value: number
+}
+
+export class PluginMetricsManager {
+    metricsPerPlugin: Record<PluginConfigId, PluginMetrics>
```
neilkakkar, line 18: judging from below, this seems like 
```suggestion
    metricsPerPlugin: Record<PluginId, PluginMetrics>
```

src/utils/plugin-metrics.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    constructor() {
+        this.metricsPerPlugin = {}
+    }
+
+    sendPluginMetrics(hub: Hub): void {
+        for (const plugin of Object.values(this.metricsPerPlugin)) {
```
neilkakkar, line 25: Small nit: calling this `plugin` is a bit confusing, when it's a `PluginMetrics` object

src/worker/vm/extensions/metrics.ts
```diff
@@ -0,0 +1,129 @@
+import equal from 'fast-deep-equal'
+
+import { Hub, MetricMathOperations, PluginConfig, PluginMetricsVmResponse, StoredPluginMetrics } from '../../../types'
+import { setPluginMetrics } from '../../../utils/db/sql'
+import { IllegalOperationError } from '../../../utils/utils'
+
+type MetricsOperations = {
+    increment: (metricName: string, value: number) => Promise<void>
```
neilkakkar, line 8: missing `max` and `min` ?

src/worker/vm/extensions/metrics.ts
```diff
@@ -0,0 +1,129 @@
...
+) {
+    if (!pluginConfig.plugin) {
+        return
+    }
+
+    let newMetrics: PluginMetricsVmResponse | StoredPluginMetrics = metrics
+    const oldMetrics = pluginConfig.plugin.metrics
+
+    if (!newMetrics) {
+        // if exportEvents exists, we'll automatically assingn metrics to it later
```
neilkakkar, line 83:
```suggestion
        // if exportEvents exists, we'll automatically assign metrics to it later
```

yakkomajuri:
Addressed all @neilkakkar !

yakkomajuri:
ping @neilkakkar

neilkakkar (approved):
Apologies I missed the ping, looks good - just some typing warnings to resolve :)

src/utils/plugin-metrics.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    constructor() {
+        this.metricsPerPluginConfig = {}
+    }
+
+    sendPluginMetrics(hub: Hub): void {
+        for (const pluginConfigMetrics of Object.values(this.metricsPerPluginConfig)) {
+            const config = pluginConfigMetrics.pluginConfig
+            const posthog = createPosthog(hub, config)
+            posthog.capture(`$plugin_metrics`, {
```
timgl, line 28: For the other 'internal' metrics that Karl did we used double $$, then also put it on a separate organisation (as those were for self hosted only). Maybe in this case we should filter them out in event_definition somehow.

yakkomajuri: we could just not create an event definition for it in the first place

yakkomajuri: Updated it here and in the UI PR to use `$$plugin_metrics`. Also added a list of events we want to skip creating definitions for - will probably be used for more events in the future. Would have filtered out everything with `$$` but maybe that's too much.

timgl: Yeah I wouldn't do that as someone might just created an event with that and then get very confused!

yakkomajuri: green light on the changes or?

PR #431, opened 2021-05-26
Buffer console writes every 100ms

mariusandra:
## Changes

This is an attempt to lazily solve https://github.com/PostHog/plugin-server/issues/424 by buffering `console.log` calls to run every 100ms. My theory is that writing ASAP causes strain (wildly fluctuating usage) on the postgres connection pool, which leads to deadlocks and those errors that I'm seeing.

Prior to this PR, when I ran this plugin locally:

```ts
export const jobs = {
    processEvent (event, meta) {
        console.log('retrying job!')
    }
}
export function processEvent (event, { jobs }) {
    console.log('in processEvent')
    jobs.processEvent(event).runIn(3, 'seconds')
    return event
}
```

... and browsed around app locally, I saw these timeouts in my console. 

After the changes in this PR I at first thought the timeouts got better, but not really. Soon after this happened:

```
[worker(worker-304ea0492a41cf15ba)] INFO: Completed task 928 (pluginJob) with success (0.43ms)
[worker(worker-0014bf26d851335d6e)] INFO: Completed task 929 (pluginJob) with success (0.47ms)
[MAIN] 12:02:05 🕒 Processed 5 events in 12s
[MAIN] 12:03:39 🕒 Processed 4 events in 94.43s
⌛⌛⌛ Postgres slow query warning after 30 sec {
  queryTextOrConfig: 'INSERT INTO posthog_pluginlogentry (id, team_id, plugin_id, plugin_config_id, timestamp, source,type, message, instance_id) VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9)',
  values: [
    '0179a81f-96e3-0000-3ac2-10e5c14925ad',
    1,
    3,
    10,
    '2021-05-26 10:01:53.123',
    'CONSOLE',
    'LOG',
    'retrying job!',
    '0179a81e-acb0-0000-9cb6-8e9003fb7044'
  ]
}
⌛⌛⌛ Postgres slow query warning after 30 sec {
  queryTextOrConfig: 'INSERT INTO posthog_pluginlogentry (id, team_id, plugin_id, plugin_config_id, timestamp, source,type, message, instance_id) VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9)',
  values: [
    '0179a81f-96e4-0000-cfe1-43b1c779d85c',
    1,
    3,
    10,
    '2021-05-26 10:01:53.124',
    'CONSOLE',
    'LOG',
    'retrying job!',
[...]

Issue #424, opened 2021-05-23
Console.log slow query warning


mariusandra:
When developing locally (with postgres), I see stuff like this in the logs way too often:

```
⌛⌛⌛ Postgres slow query warning after 30 sec {
  queryTextOrConfig: 'INSERT INTO posthog_pluginlogentry (id, team_id, plugin_id, plugin_config_id, timestamp, source,type, message, instance_id) VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9)',
  values: [
    '01799986-7949-0000-cccc-a3047fc031c1',
    1,
    3,
    10,
    '2021-05-23 13:59:57.513',
    'CONSOLE',
    'LOG',
    'retrying job!',
    '01799986-30dc-0000-cf62-e49fee9f4eff'
  ]
}
[___2] 16:07:49 🤮 Unhandled Promise Error!
```

The unhandled promise error is [from here](https://github.com/PostHog/plugin-server/blob/d86951642e9bbf9d05bcd875c6e39ce2252ea7be/src/worker/worker.ts#L69), and the "retrying job!" message was logged from within a simple plugin:

```js
export const jobs = {
    processEvent (event, meta) {
        console.log('retrying job!')
    }
}
export function processEvent (event, { jobs }) {
    console.log('in processEvent')
    jobs.processEvent(event).runIn(3, 'seconds')
    return event
}
```

I think it got worse with plugin jobs, but that might be a red herring. Possibly with just plugin logs? The `console.log` line in the plugins is not `await`-ed obviously, so perhaps that has something to do with it?

This happens often enough, but is hard to replicate reliably. Anyone have any ideas?

mariusandra:
For more context, I started exploring this when the bigquery export plugin started giving me lines like this:

![image](https://user-images.githubusercontent.com/53387/119263941-66b36280-bbe1-11eb-9828-2ddeee96cc6e.png)

It was thus hard to verify if the plugin itself works or not. I hope this happens just in local development, and not in production.

yakkomajuri:
Documenting here that we can probably have some batching mechanism to periodically insert batches into the DB at once. I think this was mentioned elsewhere too

neilkakkar:
Possibly related to issues with Bigquery plugin as well: https://github.com/PostHog/bigquery-plugin/pull/9

PR #486, opened 2021-06-24
batch plugin logs on postgres

yakkomajuri:
## Changes

Addresses #424 and supersedes #431.

This makes log inserts to Postgres from plugins happen in batches every second. Didn't implement batching for other logs but should be easy to do so later if we feel there's a need. Big improvements from this locally. No slow query warnings, and despite the buffer, the logs actually appear faster (since they're not "stuck in traffic").

## Checklist

-   [ ] Jest tests


neilkakkar (approved):
LGTM.

Surprised though that this solves the underlying issue.

In effect, were we making too many queries in succession to postgres? We're still making about 1/second. So does this imply earlier, we were even faster than that?

mariusandra (changes requested):
Very nice! Just two thoughts:
- even though we flush every second, perhaps for posterity it makes sense to also flush this buffer in teardown/shutdown mode? it's yet another stream/buffer/service/etc, so might make sense to control in a similar way...
- it looks like you're always inserting to postgres, even if kafka is enabled.

src/worker/vm/extensions/console.ts
```diff
@@ -17,18 +19,30 @@ function consoleFormat(...args: unknown[]): string {
...
 
 export function createConsole(server: Hub, pluginConfig: PluginConfig): ConsoleExtension {
-    async function consolePersist(type: PluginLogEntryType, ...args: unknown[]): Promise<void> {
+    const consoleBuffer = createBuffer({
+        limit: 100,
+        timeoutSeconds: 1,
+        onFlush: async (logEntries: LogEntryPayload[]) => {
+            await server.db.createPluginLogEntries(pluginConfig, logEntries)
+        },
+    })
```
Twixes, line 28: This solution seems very high-level, as it only covers `console` and in a per-plugin fashion – compared to Kafka message queuing, which covers the whole plugin server thread. I think if we do want to go for batching here, it may be cleaner and faster to have `createPluginLogEntry` append to a buffer, which would be flush for the whole the thread. That would avoid the situation where there are also _two_ buffer levels for Kafka-produced `console` logs (initial buffer here, then second buffer in `kafkaProducer.queueMessage`).

Twixes (changes requested):

src/utils/db/db.ts
```diff
@@ -698,52 +706,73 @@ export class DB {
...
+                        topic: KAFKA_PLUGIN_LOG_ENTRIES,
+                        messages: [{ key: parsedEntry.id, value: Buffer.from(JSON.stringify(parsedEntry)) }],
+                    })
+                } catch (e) {
+                    captureException(e)
+                    console.error(parsedEntry)
+                    console.error(e)
+                }
+            }
+            return
```
yakkomajuri, line 743: @mariusandra this should prevent the insert into postgres

src/worker/vm/extensions/console.ts
yakkomajuri: It is indeed. I originally came at this for writing a general batching system, but decided to start with this as the bulk of logs comes from plugins and I had to work on other things.

yakkomajuri:
@mariusandra re

```
even though we flush every second, perhaps for posterity it makes sense to also flush this buffer in teardown/shutdown mode? it's yet another stream/buffer/service/etc, so might make sense to control in a similar way...
```

yeah that'd make sense

src/worker/vm/extensions/console.ts
yakkomajuri: The system is there to take a batch of entries and insert them so it definitely isn't a major step to write worker-level memory batching. I just took over #431 to get this out quickly

src/utils/db/db.ts
mariusandra: Oh god, that's embarrassing :D. 

In my defence I have a strong cold, but still... :D

yakkomajuri:
huh, these passed locally 🤔 
Screenshot 2021-07-02 at 10 44 43

yakkomajuri:
@Twixes thread-level batching now

yakkomajuri:
friendly ping @Twixes

tests/postgres/e2e.test.ts
```diff
@@ -117,10 +117,10 @@ describe('e2e postgres ingestion', () => {
 
         posthog.capture('custom event', { name: 'hehe', uuid: new UUIDT().toString() })
 
+        // await delayUntilEventIngested(async () =>  )
```
Twixes, line 120: Why commented out?

tests/postgres/vm.test.ts
```diff
@@ -29,6 +30,7 @@ let closeHub: () => Promise<void>
 beforeEach(async () => {
     ;(Client as any).mockClear()
     ;[hub, closeHub] = await createHub()
+    // hub.db.createPluginLogEntries = jest.fn()
```
Twixes, line 33: Commented out

Twixes (commented):
Looks like this works so if unbatched Postgres log insertion is a hair-on-fire problem, it's okay, but still has this thing where Kafka logs are now batched two-fold – first via `addLog` and then via `queueMessage`. Not ideal since that only increases the time in which a log entry gets into ClickHouse, without benefits. Basically ideally `createPluginLogEntries` should queue a Kafka message (which it already does) _or_ queue a Postgres row, instead of `LogsBuffer` creating an additional buffer layer above `createPluginLogEntries`.

yakkomajuri:
That's a good point - `addLog` can just directly throw it in Kafka if that's available

yakkomajuri:
done @Twixes

src/utils/logs-buffer.ts
```diff
@@ -12,11 +12,19 @@ export class LogsBuffer {
         this.flushTimeout = null
     }
 
-    addLog(log: LogEntryPayload): void {
+    async addLog(log: LogEntryPayload): Promise<void> {
+        // drop into kafka queue immediately as that's aready a buffer
+        if (this.db.kafkaProducer) {
+            await this.db.createPluginLogEntries([log])
+            return
+        }
```
Twixes, line 20: This is kind of confusing. So this way in Kafka/CH mode, we have a `LogsBuffer`, which goes through all the motions as usual… but it's illusory because no data is buffered in it, as the _actual_ buffering is handed of to the Kafka-specific buffer immediately.

src/utils/db/db.ts
```diff
@@ -698,52 +707,73 @@ export class DB {
...
+                    ...[id, team_id, plugin_id, plugin_config_id, timestamp, source, type, message, instance_id],
+                ]
             }
+
+            await this.postgresQuery(
+                `INSERT INTO posthog_pluginlogentry (id, team_id, plugin_id, plugin_config_id, timestamp, source,type, message, instance_id) 
+                VALUES ${valuesString}`,
+                values,
+                'insertPluginLogEntries'
+            )
```
Twixes, line 772: I'd like to ask you to rearchitect this bit. Let's put Kafka and Postgres on equal buffering footing to make this as straightforward as possible – since `queueMessage` is already buffered within a `KafkaProducerWrapper`, we should have something like `queuePostgresInsert` buffered within a `PostgresWrapper`. This way there is each of Kafka and Postgres only ever has a single simple layer of buffering, and that layer is Kafka-/Postgres-specific.

Twixes (changes requested):
Sorry for dragging this on a bit, but my initial point still stands unfortunately 😅

yakkomajuri:
apologies for the lazy updates on the existing system instead of just rearchitecting in the first place :D

@Twixes

src/worker/plugins/teardown.ts
```diff
@@ -14,33 +14,33 @@ export async function teardownPlugins(server: Hub, pluginConfig?: PluginConfig):
...
-                            await server.db.createPluginLogEntry(
+                            await server.db.queuePluginLogEntry({
                                 pluginConfig,
-                                PluginLogEntrySource.System,
-                                PluginLogEntryType.Error,
-                                `Plugin failed to unload (instance ID ${server.instanceId}).`,
-                                server.instanceId
-                            )
+                                source: PluginLogEntrySource.System,
+                                type: PluginLogEntryType.Info,
```
Twixes, line 29: Why the change to `Info`?

yakkomajuri: great catch! probably lingered from a copy paste

Twixes (approved):
Benissimo

PR #495, opened 2021-07-05
Add support for @google-cloud/storage

yakkomajuri:
## Changes

Will need this for GCS uploads - both in the standalone plugin and in Snowflake.

## Checklist

-   [ ] Jest tests


yakkomajuri:
Also adds the harmless `stream.PassThrough`, needed because the GCS driver expects a access to the file system.

neilkakkar (approved):

PR #496, opened 2021-07-07
Action matching on static cohorts for CH

### LINKED ISSUE #484: Implement static cohort support in action matching
Currently static cohorts are not supported in action matching. This isn't a big deal, as the only result is that webhooks/Zapier won't fire in case of a static cohort match (which is a rare combination), but we should definitely improve this.

yakkomajuri:
## Changes

Closes #484 

Postgres is fine, everyone is in `posthog_cohortpeople`. But in CH we put people in `person_static_cohort`.

Given we don't know if the cohort is static or not when matching, I went the approach of just checking CH for a static cohort match, and if that's not the case, then check Postgres (where the "normal" cohorts are).

The other approach would be to check from Postgres if the cohort is static or not before hitting CH, but that'd result in more queries overall.

## Checklist

-   [ ] Jest tests

### DIFF
--- a/src/utils/db/db.ts
+++ b/src/utils/db/db.ts
@@ -569,12 +569,16 @@ export class DB {
     public async doesPersonBelongToCohort(cohortId: number, person: Person, teamId: Team['id']): Promise<boolean> {
         if (this.kafkaProducer) {
             const chResult = await this.clickhouseQuery(
-                `SELECT 1 FROM person_static_cohort WHERE team_id = ${teamId} AND cohort_id = ${cohortId} AND person_id = '${escapeClickHouseString(
-                    person.uuid
-                )}'`
+                `SELECT 1 FROM person_static_cohort
+                WHERE
+                    team_id = ${teamId}
+                    AND cohort_id = ${cohortId}
+                    AND person_id = '${escapeClickHouseString(person.uuid)}'
+                LIMIT 1`
             )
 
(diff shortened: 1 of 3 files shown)


Twixes:
Hm I thought we do know if a cohort is static, `posthog_cohort` column `is_static`?

yakkomajuri:
> Hm I thought we do know if a cohort is static, posthog_cohort column is_static?

I meant in memory at the time this happens. So the alternative approach (what I originally built) is to check Postgres for this, but we'll have fewer queries overall hitting CH `person_static_cohort` without knowing than checkin Postgres before.

@Twixes

Twixes:
The relevant `db.doesPersonBelongToCohort` call is within (4 levels of calls, as the doc comment signifies) `checkAction` (https://github.com/PostHog/plugin-server/blob/ch-static-cohort/src/worker/ingestion/action-matcher.ts#L136-L148), so I was thinking it should be easy to just pass the `action` down so that we do know if it `is_static` in `checkEventAgainstCohortFilter`.

yakkomajuri:
The action just has a cohort ID though - I'd still need to query Postgres to know if it's static, no?

Twixes (approved):
Ah, right, was being a bit dumb my mixing `Action` and `Cohort` data up. Okay, this doesn't have to super optimized anyway. LGTM!

PR #501, opened 2021-07-12
fix nan timestamp

yakkomajuri:
## Changes

Should solve [PLUGIN-SERVER-DN ](https://sentry.io/organizations/posthog/issues/2483414066/events/latest/?project=5592816&statsPeriod=14d). Seems for some reason we're getting timestamps in this format: 

```
MM/DD/YYYY HH:MM:SS AM|PM
```

Haven't managed to reliably pinpoint how we get here though. 

## Checklist

-   [ ] Jest tests

### DIFF
--- a/src/worker/ingestion/process-event.ts
+++ b/src/worker/ingestion/process-event.ts
@@ -166,7 +166,8 @@ export class EventsProcessor {
                 try {
                     // timestamp and sent_at must both be in the same format: either both with or both without timezones
                     // otherwise we can't get a diff to add to now
-                    return now.plus(DateTime.fromISO(data['timestamp']).diff(sentAt))
+                    const timestamp = new Date(data['timestamp']).toISOString()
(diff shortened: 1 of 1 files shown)


src/worker/ingestion/process-event.ts
```diff
@@ -166,7 +166,8 @@ export class EventsProcessor {
                 try {
                     // timestamp and sent_at must both be in the same format: either both with or both without timezones
                     // otherwise we can't get a diff to add to now
-                    return now.plus(DateTime.fromISO(data['timestamp']).diff(sentAt))
+                    const timestamp = new Date(data['timestamp']).toISOString()
+                    return now.plus(DateTime.fromISO(timestamp).diff(sentAt))
```
Twixes, line 170: Hint: Luxon `DateTime` has `DateTime.fromJSDate`

yakkomajuri:
I have seen some divergence between `Date` on newer Safari versions and other browsers. Haven't managed to establish that as the root though

PR #507, opened 2021-07-22
Save distinct ids on clickhouse person table

macobo:
This change adds support for populating `person.distinct_ids` column in clickhouse. 

Note that the implementation "seems" correct but is very susceptible to race conditions around fetching and then updating person entries. If the experiment is successful, the next step is to use locking to remove the raciness.

Issue: https://github.com/PostHog/posthog/issues/5167

Depends on https://github.com/PostHog/posthog/pull/5276

## Checklist

-   [ ] Jest tests

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -480,6 +480,7 @@ export interface ClickHousePerson {
     is_identified: number
     is_deleted: number
     timestamp: string
+    distinct_ids: string[]
 }
 
 /** Usable PersonDistinctId model. */
--- a/src/types.ts
+++ b/src/types.ts
@@ -466,6 +466,11 @@ export interface Person extends BasePerson {
     created_at: DateTime
 }
 
+/** Returned by fetchPerson */
+export interface PersonWithDistinctIds extends Person {
+    distinct_ids: string[]
+}
+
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -57,6 +57,7 @@ jobs:
               with:
                   repository: 'PostHog/posthog'
                   path: 'posthog/'
+                  ref: distinct-ids-on-person
```
macobo, line 60: This is needed for tests to run. Will remove before merging.

neilkakkar:
I’m assuming that if this experiment doesn’t give good results, we’ll revert it?

Do we have a safe revert path for these 2 dependent PRs?

macobo:
> I’m assuming that if this experiment doesn’t give good results, we’ll revert it?
> Do we have a safe revert path for these 2 dependent PRs?

Note that this needs to be live for several days before we can do an experiment (to have enough data).

Yep, we'll revert then by reverting this plugin-server code + creating a migration that removes the column.

neilkakkar (approved):
This looks good to me 👍

PR #508, opened 2021-07-22
Add faker.js

mariusandra:
## Changes

- Enables plugins like this:

```js
import faker from 'faker'

async function runEveryMinute(meta) {
    console.log(faker.name.findName())
}
```

Which then do this:

image

This is slightly easier than copier the 1.5MB `faker.min.js` file inside the plugin source editor.

**Small issue**

Faker.js requires node.js >= 14.17, which might not be what's installed everywhere. I'm not sure what it needs from such a new node version, and it might just work fine as is. Alternatively, people will have problems installing the plugin server if they are running an older node version. I'm not sure which will be the case.

## Checklist

-   [ ] Jest tests


mariusandra:
There's also a problem with running the plugin above inside a for loop: `posthog.capture` doesn't give us a promise to `await`, so we're stuck in the atomics wait loop. Hence we get timeouts and other fun.

Twixes:
_Should_ `posthog.capture` return a promise?

PR #518, opened 2021-08-03
Retry on every error in setupPlugin

neilkakkar:
## Changes

Seems like `test-postgres-1` succeeds in CI, even on failure

## Checklist

-   [ ] Jest tests

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -2,6 +2,7 @@ module.exports = {
     preset: 'ts-jest',
     testEnvironment: 'node',
     clearMocks: true,
+    bail: 1,
     coverageProvider: 'v8',
     setupFilesAfterEnv: ['./jest.setup.fetch-mock.js'],
     testMatch: ['<rootDir>/tests/**/*.test.ts', '<rootDir>/benchmarks/**/*.benchmark.ts'],
--- a/jest.config.js
+++ b/jest.config.js
@@ -2,7 +2,6 @@ module.exports = {
     preset: 'ts-jest',
     testEnvironment: 'node',
     clearMocks: true,
-    bail: 1,
     coverageProvider: 'v8',
     setupFilesAfterEnv: ['./jest.setup.fetch-mock.js'],
     testMatch: ['<rootDir>/tests/**/*.test.ts', '<rootDir>/benchmarks/**/*.benchmark.ts'],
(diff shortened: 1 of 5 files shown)


jest.config.js
```diff
@@ -2,6 +2,7 @@ module.exports = {
     preset: 'ts-jest',
     testEnvironment: 'node',
     clearMocks: true,
+    bail: 1,
```
neilkakkar, line 5: not sure why, but jest registers as a success even on failure without this: https://github.com/PostHog/plugin-server/runs/3231307579

Unsure if there's a better way to fix the problem

tests/plugins.test.ts
```diff
@@ -64,10 +64,8 @@ test('setupPlugins and runProcessEvent', async () => {
     expect(pluginConfig.config).toEqual(pluginConfig39.config)
     expect(pluginConfig.error).toEqual(pluginConfig39.error)
 
-    expect(pluginConfig.plugin).toEqual({
-        ...plugin60,
-        capabilities: { jobs: [], scheduled_tasks: [], methods: ['processEvent'] },
-    })
+    expect(pluginConfig.plugin).toEqual(plugin60)
```
neilkakkar, line 67: this seems like a regression introduced in https://github.com/PostHog/plugin-server/runs/3059157937 where these tests didn't run

jest.config.js
yakkomajuri: I'm guessing `bail` will produce the correct exit code but will not run any other files once a test in a file fails. While this is fine for CI, I wouldn't want this to be the case locally. We can maybe use `--bail` on the CI command rather than having this as a global config?

Also feels like this isn't the best way to fix this, but I'd be ok to merge as it's an improvement

yakkomajuri: More context: It seems the failure doesn't register because it's a TS error rather than an actual error encountered when running the test.

yakkomajuri: I believe the issue is `ts-jest`. We can try to compile before running tests and that might do it.

yakkomajuri:
Took the liberty of adding a commit here but feel free to push back.

The problem here seems to be that `ts-jest` doesn't produce a failing code on a compilation error for a test, it just skips to the next one. If we make sure everything compiles fine before running jest we should be getting these errors.

Twixes:
Now errors with this 👀 
Err

yakkomajuri:
typo, sorry

yakkomajuri (approved):

PR #519, opened 2021-08-03
Add API extension

yakkomajuri:
## Changes

This allows plugin devs to access the PostHog API with a lot more ease and security than right now.

This is now possible, for example:

```js
await posthog.api.get('/api/event')
```

Currently, devs have to do all the fetch requests themselves, meaning they also need to ask the user for a Personal API Key. We have a frontend mechanism to automatically provision one if you name your config option `posthogApiKey`, but this really isn't great. 

Not only that, but after https://github.com/PostHog/posthog/pull/5044, devs now should also specify a project to ensure they're accessing the right data. The token for this could be pulled from events, but what about plugins that only use e.g. `runEveryMinute` and access API, like the GitHub, GitLab, and Bitbucket annotators?

I originally set out to just build a simple way for devs to access the project token, but then I figured there's a lot more help we can give them here.

Admittedly, this isn't the nicest solution. What I'd like is a secret project token, but I have been convinced by others that dummy users might be the simplest way to go about this as of right now.

Hence, this mechanism provisions a "bot" user for the org in which the plugin using the API is enabled and a personal API key for this user, using this to execute any API requests. No need to ask users for personal API keys nor project tokens.

## Potential Additions

Having started to write this description, I realized it would probably be best to allow overriding the personal API key and the project token, so users can also access other PostHog instances/teams this way, but with the default behavior always being scoped to their specific team, to prevent unintended operations.

Also needs some good docs.

### DIFF
--- a/src/utils/utils.ts
+++ b/src/utils/utils.ts
@@ -649,3 +649,54 @@ export class IllegalOperationError extends Error {
         super(operation)
     }
 }
+
+export function getByAge<K, V>(cache: Map<K, [V, number]>, key: K, maxAgeMs = 30_000): V | undefined {
+    if (cache.has(key)) {
+        const [value, age] = cache.get(key)!
+        if (Date.now() - age <= maxAgeMs) {
+            return value
+        }
+    }
+    return undefined
+}
+
+// Equivalent of Python's string.ascii_letters
+export function getAsciiLetters() {
+    const LOWERCASE_START_POINT = 97 // ASCII 'a'
+    const UPPERCASE_START_POINT = 65 // ASCII 'A'
+
+    const lowercaseLetters = Array.from({ length: 26 }).map((_, i) => String.fromCharCode(LOWERCASE_START_POINT + i))
+    const uppercaseLetters = Array.from({ length: 26 }).map((_, i) => String.fromCharCode(UPPERCASE_START_POINT + i))
+
+    return `${lowercaseLetters.join('')}${uppercaseLetters.join('')}`
+}
+
+// Equivalent of Python's string.digits
+export function getAllDigits() {
+    return Array.from({ length: 10 })
+        .map((_, i) => i)
+        .join('')
+}
+
+export function generateRandomToken(nBytes: number) {
+    return intToBase(Number.parseInt(randomBytes(nBytes).toString('hex'), 16), 62)
+}
+
+export function intToBase(num: number, base: number): string {
+    if (base > 62) {
+        throw new IllegalOperationError('Cannot convert integer to base above 62')
+    }
+    const alphabet = getAllDigits() + getAsciiLetters()
+    if (num < 0) {
+        return '-' + intToBase(-num, base)
+    }
+    let value = ''
+    while (num != 0) {
+        const oldNum = num
+        num = Math.floor(oldNum / alphabet.length)
+        const index = oldNum % alphabet.length
+        value = alphabet[index] + value
+    }
+
+    return value || '0'
+}
--- a/src/utils/utils.ts
+++ b/src/utils/utils.ts
@@ -661,7 +661,7 @@ export function getByAge<K, V>(cache: Map<K, [V, number]>, key: K, maxAgeMs = 30
 }
 
 // Equivalent of Python's string.ascii_letters
-export function getAsciiLetters() {
+export function getAsciiLetters(): string {
     const LOWERCASE_START_POINT = 97 // ASCII 'a'
     const UPPERCASE_START_POINT = 65 // ASCII 'A'
 
@@ -672,13 +672,13 @@ export function getAsciiLetters() {
 }
 
 // Equivalent of Python's string.digits
-export function getAllDigits() {
+export function getAllDigits(): string {
     return Array.from({ length: 10 })
         .map((_, i) => i)
         .join('')
 }
 
-export function generateRandomToken(nBytes: number) {
+export function generateRandomToken(nBytes: number): string {
     return intToBase(Number.parseInt(randomBytes(nBytes).toString('hex'), 16), 62)
 }
 
@@ -699,10 +699,4 @@ export function intToBase(num: number, base: number): string {
     }
 
     return value || '0'
-}
-
-// helps make certain conditionals cleaner and more readable, like:
-// !!maybeTruthy === !!alsoMaybeTruthy
-export function isTruthy(value: any): boolean {
-    return !!value
 }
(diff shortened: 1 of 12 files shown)


yakkomajuri:
Reminder to self: Add ability to configure tokens

src/worker/vm/extensions/api.ts
```diff
@@ -0,0 +1,94 @@
...
+            tokenParam['token'] = team.api_token
+            apiKey = await server.pluginsApiKeyManager.fetchPluginsPersonalApiKey(team.organization_id)
+        }
+
+        const urlParams = new URLSearchParams(
+            method === (ApiMethod.Get || ApiMethod.Delete) && options && options.data
+                ? { ...options.data, ...tokenParam }
+                : tokenParam
+        )
+        const url = `${host}/${path}?${urlParams.toString()}`
```
Twixes, line 66: We should detect if `path` already contains a `?` and use `&` instead here if so.

Twixes (changes requested):
This automagical user solution is kinda dirty indeed. But there needs to be _some_ accountability for things happening via the API and all of it is based on the accounts system currently, so when I imagine the work needed to support properly-accounted-for internal service access, the automagical user seems alright.

Leaving a few notes regarding bits of code. BTW some new functions/methods here lack a return type, let's make sure we're all in on TS there. :)

src/utils/utils.ts
```diff
@@ -649,3 +649,60 @@ export class IllegalOperationError extends Error {
...
+        const index = oldNum % alphabet.length
+        value = alphabet[index] + value
+    }
+
+    return value || '0'
+}
+
+// helps make certain conditionals cleaner and more readable, like:
+// !!maybeTruthy === !!alsoMaybeTruthy
+export function isTruthy(value: any): boolean {
```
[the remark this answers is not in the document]
yakkomajuri: Fair point. I just thought `!!a !== !!b` was so unreadable

src/worker/vm/extensions/helpers/api-key-manager.ts
```diff
@@ -0,0 +1,91 @@
+import { DB } from '../../../../utils/db/db'
+import { timeoutGuard } from '../../../../utils/db/utils'
+import { generateRandomToken, getByAge, UUIDT } from '../../../../utils/utils'
+import { RawOrganization } from './../../../../types'
+
+type PluginsApiKeyCache<T> = Map<RawOrganization['id'], [T, number]>
+
+const PLUGINS_API_KEY_USER_EMAIL = '$plugins_personal_api_key'
```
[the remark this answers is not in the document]
yakkomajuri: Nothing seemed to pop up, but it's a valid point, I should confirm

src/worker/vm/extensions/helpers/api-key-manager.ts
```diff
@@ -0,0 +1,91 @@
...
+        const cachedKey = getByAge(this.pluginsApiKeyCache, organizationId)
+        if (cachedKey) {
+            return cachedKey
+        }
+
+        const timeout = timeoutGuard(`Still running "fetchPluginsPersonalApiKey". Timeout warning after 30 sec!`)
+        try {
+            let key: string | null = null
+            const userResult = await this.db.postgresQuery(
+                `SELECT id FROM posthog_user WHERE email = '${PLUGINS_API_KEY_USER_EMAIL}'`,
```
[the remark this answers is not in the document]
yakkomajuri: Yes sorry actually this is dev code I didn't update

src/utils/utils.ts
```diff
@@ -649,3 +649,60 @@ export class IllegalOperationError extends Error {
...
+
+// Equivalent of Python's string.digits
+export function getAllDigits() {
+    return Array.from({ length: 10 })
+        .map((_, i) => i)
+        .join('')
+}
+
+export function generateRandomToken(nBytes: number) {
+    return intToBase(Number.parseInt(randomBytes(nBytes).toString('hex'), 16), 62)
```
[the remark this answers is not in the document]
yakkomajuri: `toString` will return utf8 encoding by default, which `parseInt` will return `NaN` for

yakkomajuri:
Great points, thanks @Twixes

mariusandra (commented):
Looks good to me in general, though I wonder if we should create a new api key per project, and not one per team? Right now it won't make a difference, but if we add [project based permissioning](https://github.com/PostHog/posthog/issues/4263), should be improve something here for security?

Otherwise looks good to me, just 1) same caveats for security and storing plain text keys, and 2) will this change anything in the organisation's user's list? We now have a user with an e-mail that's not an e-mail. This may cause issues somewhere?

yakkomajuri:
@mariusandra I guess you mean one per project rather than one per org? Our terminology with project being the same as team makes things confusing 😅 

Either way, ultimately this just isn't super clean as a solution, and while a user/key per project would generally make more sense, this would lead to more dummy users. Keys aren't scoped by project anyway so a key scoped by project would still give access to everything else.

Then, regarding 1, there's nothing in this PR that messes with what we already have in that front. And the answer for 2 is I have to check the problems of this not being an email. Might as well give it an "email" I guess but thought that could be confusing too. Will look into options and circle back here.

src/worker/vm/extensions/helpers/api-key-manager.ts
```diff
@@ -38,19 +38,22 @@ export class PluginsApiKeyManager {
...
             const userResult = await this.db.postgresQuery(
-                `SELECT id FROM posthog_user WHERE email = $1`,
-                [PLUGINS_API_KEY_USER_EMAIL],
+                `SELECT id FROM posthog_user WHERE email LIKE $1`,
+                [`%@${POSTHOG_BOT_USER_EMAIL_DOMAIN}`],
                 'fetchPluginsUser'
             )
 
             if (userResult.rowCount < 1) {
+                const botUserEmailId = Math.round(Math.random() * 100000000)
```
Twixes, line 47: Why not `organizationId`?

yakkomajuri: I was thinking this is useful if we do later decide to add more bot users

Twixes: Most proper-domain-like-but-still-dummy would be something like below
```TypeScript
`${organizationId}@system.posthog`
```
 (and I _hope_ we don't add more bot users because that'd be too much 😅), but current scheme should be alright too.

Twixes (approved):

src/worker/vm/extensions/helpers/api-key-manager.ts
yakkomajuri: Well for example [this comment](https://github.com/PostHog/plugin-server/pull/519#pullrequestreview-730393631) brings up one approach that would lead to multiple bot users

PR #521, opened 2021-08-09
Capture handleIdentifyOrAlias exceptions in Sentry

yakkomajuri:
## Changes

I can confirm that `handleIdentifyOrAlias` **does** throw in certain conditions. We're just not doing anything when that happens. 

This will be important to more precisely identify the most common race conditions affecting us.

## Checklist

-   [ ] Jest tests

### DIFF
--- a/src/worker/ingestion/process-event.ts
+++ b/src/worker/ingestion/process-event.ts
@@ -100,6 +100,8 @@ export class EventsProcessor {
             })
             try {
                 await this.handleIdentifyOrAlias(data['event'], properties, distinctId, teamId)
+            } catch (e) {
+                Sentry.captureException(e)
             } finally {
                 clearTimeout(timeout1)
             }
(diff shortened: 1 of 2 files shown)


macobo (approved):

src/worker/ingestion/process-event.ts
```diff
@@ -100,6 +100,8 @@ export class EventsProcessor {
             })
             try {
                 await this.handleIdentifyOrAlias(data['event'], properties, distinctId, teamId)
+            } catch (e) {
+                Sentry.captureException(e, { extra: { event: data } })
```
macobo, line 104: Probably worth capturing teamId as well?

yakkomajuri: That should come along with `data`. `data` is of type `PluginEvent`, which includes:

```js
export interface PluginEvent {
    distinct_id: string;
    ip: string | null;
    site_url: string;
    team_id: number;
    now: string;
    event: string;
    sent_at?: string;
    properties?: Properties;
    timestamp?: string;
    offset?: number;
    $set?: Properties;
    $set_once?: Properties;
    kafka_offset?: string;
    uuid?: string;
}
```

PR #523, opened 2021-08-10
Handle elements conflict gracefully

### LINKED ISSUE #522: Repeated error in RDS logs
## In what situation are you experiencing subpar performance?

Our RDS logs are getting hammered by the same message every few seconds, not sure if this is a bug, or peformance related, but we're churning through logs and our metrics don't look great.

```
2021-07-30 11:00:06 UTC:10.0.10.6(54102):[9249]:ERROR: duplicate key value violates unique constraint "unique hash for each team"
2021-07-30 11:00:06 UTC:10.0.10.6(54102):[9249]:DETAIL: Key (team_id, hash)=(1, <HASH>) already exists.
2021-07-30 11:00:06 UTC:10.0.10.6(54102) [9249]:STATEMENT: INSERT INTO posthog_elementgroup (hash, team_id) VALUES ($1, $2) RETURNING *
```

## How to reproduce

1. Fire up a self hosted PostHog in AWS, on 1.27.0?

## Environment

- [ ] PostHog Cloud
- [x] self-hosted PostHog, version/commit: 1.27.0

## Additional context

N/A

yakkomajuri:
## Changes

Fixes #522 

Prevents us from erroring (and thus log spamming) on a pretty natural and harmless race condition where we try to insert the same elements twice.

## Checklist

-   [ ] Jest tests

### DIFF
--- a/src/utils/db/db.ts
+++ b/src/utils/db/db.ts
@@ -745,27 +745,30 @@ export class DB {
         try {
             await this.postgresTransaction(async (client) => {
                 const insertResult = await client.query(
-                    'INSERT INTO posthog_elementgroup (hash, team_id) VALUES ($1, $2) RETURNING *',
+                    'INSERT INTO posthog_elementgroup (hash, team_id) VALUES ($1, $2) ON CONFLICT DO NOTHING RETURNING *',
                     [hash, teamId]
                 )
-                const elementGroup = insertResult.rows[0] as ElementGroup
-                for (const element of cleanedElements) {
-                    await client.query(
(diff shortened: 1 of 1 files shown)


src/utils/db/db.ts
```diff
@@ -745,27 +745,30 @@ export class DB {
...
+                                element.nth_child,
+                                element.nth_of_type,
+                                element.attributes || '{}',
+                                element.order,
+                                element.event_id,
+                                element.attr_class,
+                                elementGroup.id,
+                            ]
+                        )
+                    }
```
Twixes, line 771: BTW this looks like a good opportunity for replacing multiple insertions (slowed down by `await`) with a single `INSERT INTO foo (a, b) VALUES ($1, $2), ($3, $4), ...`

Twixes (approved):
This looks good

src/utils/db/db.ts
yakkomajuri: I'll get this in and do another PR just so I'm conceptually getting through my to-do list as it pertains to race conditions and all

PR #524, opened 2021-08-10
Handle moveDistinctIds race conditions

yakkomajuri:
## Changes

Me and @macobo will probably discuss this a bit further tomorrow.

Handles 2 types of race conditions that may arise when running `mergeDistinctIds`:

- `mergeInto` person no longer exists: this triggers [an error](https://sentry.io/organizations/posthog/issues/2454986618/?project=5592816&query=is%3Aunresolved+pg&statsPeriod=14d) and leaves the persons unmerged
- `otherPerson` no longer exists: this currently does not trigger an error but leaves persons unmerged

These happen because in between the time the persons are fetched and `moveDistinctIds` is called, another thread may have deleted either of the persons. Happy to explain further. 

## Checklist

-   [ ] Jest tests

### DIFF
--- a/src/utils/db/db.ts
+++ b/src/utils/db/db.ts
@@ -44,6 +44,7 @@ import {
     castTimestampOrNow,
     clickHouseTimestampToISO,
     escapeClickHouseString,
+    RaceConditionError,
     sanitizeSqlIdentifier,
     tryTwice,
     UUID,
@@ -605,7 +606,7 @@ export class DB {
         // this is caused by a race condition and will trigger a retry with
         // updated persons
         if (movedDistinctIdResult.rows.length === 0) {
-            throw new Error(`Failed trying to move distinct IDs because otherPerson doesn't exist.`)
+            throw new RaceConditionError(`Failed trying to move distinct IDs because otherPerson doesn't exist.`)
         }
 
         if (this.kafkaProducer) {
--- a/src/utils/db/db.ts
+++ b/src/utils/db/db.ts
@@ -604,7 +604,7 @@ export class DB {
 
         // this is caused by a race condition and will trigger a retry
         if (movedDistinctIdResult.rows.length === 0) {
-            throw new Error('Tried to move ditinct IDs from a non-existent person')
+            throw new Error('Tried to move ditinct IDs from an unreferenced person')
         }
 
         if (this.kafkaProducer) {
--- a/src/utils/db/db.ts
+++ b/src/utils/db/db.ts
@@ -602,9 +602,10 @@ export class DB {
             'updateDistinctIdPerson'
         )
 
-        // this is caused by a race condition and will trigger a retry
+        // this is caused by a race condition and will trigger a retry with
+        // updated persons
         if (movedDistinctIdResult.rows.length === 0) {
-            throw new Error('Tried to move ditinct IDs from an unreferenced person')
+            throw new Error(`Failed trying to move distinct IDs because otherPerson doesn't exist.`)
         }
 
         if (this.kafkaProducer) {
--- a/src/utils/db/db.ts
+++ b/src/utils/db/db.ts
@@ -591,22 +591,40 @@ export class DB {
     }
 
     public async moveDistinctIds(source: Person, target: Person): Promise<void> {
-        const movedDistinctIdResult = await this.postgresQuery(
-            `
-                UPDATE posthog_persondistinctid
-                SET person_id = $1
-                WHERE person_id = $2
-                  AND team_id = $3
-                RETURNING *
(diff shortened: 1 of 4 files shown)


src/utils/db/db.ts
```diff
@@ -602,6 +602,11 @@ export class DB {
             'updateDistinctIdPerson'
         )
 
+        // this is caused by a race condition and will trigger a retry
+        if (movedDistinctIdResult.rows.length === 0) {
+            throw new Error('Tried to move ditinct IDs from an unreferenced person')
```
macobo, line 607: Nit: Typo. :)

yakkomajuri:
Changing the approach from the "inline alias" required adding `totalMergeAttempts` to `alias` so we can bail out of what could under very rare and specific circumstances be an infinite loop.

Also had moved towards `moveDistinctIds` returning a success status but I think the  explicit `RaceConditionError` makes sense here

yakkomajuri:
@macobo this should be good for a second look

src/utils/db/db.ts
```diff
@@ -591,22 +591,40 @@ export class DB {
...
+                [target.id, source.id, target.team_id],
+                'updateDistinctIdPerson'
+            )
+        } catch (error) {
+            if (
+                error.message.includes(
+                    'insert or update on table "posthog_persondistinctid" violates foreign key constraint'
+                )
+            ) {
+                operationFailedDueToRaceCondition = true
```
macobo, line 614: Q: Why not `throw new RaceConditionError` here? Would simplify the code somewhat, also the two error messages can be different/better.

src/utils/db/db.ts
```diff
@@ -591,22 +591,40 @@ export class DB {
...
+            } else {
+                throw error
+            }
+        }
+
+        // this should never (?) happen, but needed to satisfy the TS compiler
```
macobo, line 620: Let's instead add a type-level assert - this is noise.

You can do this by appending `!` at the end of every `movedDistinctIdResult`

src/worker/ingestion/process-event.ts
```diff
@@ -331,17 +331,29 @@ export class EventsProcessor {
...
     }
 
-    public async mergePeople(
-        mergeIntoPayload: [Person, string],
-        otherPersonPayload: [Person, string],
-        totalMergeAttempts = 0
-    ): Promise<void> {
-        const [mergeInto, mergeIntoDistinctId] = mergeIntoPayload
-        const [otherPerson, otherPersonDistinctId] = otherPersonPayload
+    public async mergePeople({
```
macobo, line 344: Q: Why struct over just 5 params?

src/worker/ingestion/process-event.ts
```diff
@@ -376,17 +390,19 @@ export class EventsProcessor {
...
+                if (error instanceof RaceConditionError && failedAttempts < MAX_FAILED_PERSON_MERGE_ATTEMPTS) {
+                    shouldRetryAliasOperation = true
+                    break
                 }
 
-                // If a person was deleted in between fetching and moveDistinctId,
-                // re-run alias to ensure the updated persons are fetched and
-                // merged safely
-                return await this.alias(otherPersonDistinctId, mergeIntoDistinctId, teamId, false, failedAttempts)
+                throw error
```
macobo, line 405: Q: Is the other try-catch still valid? if not, it makes sense to perhaps remove the loop alltogether.

macobo (commented):
Some more notes. I think this can and should be simplified.

src/utils/db/db.ts
yakkomajuri: Yeah I'm aware of this, was just wondering what would be cleaner. But you're right.

src/worker/ingestion/process-event.ts
yakkomajuri: Honestly, I find it cleaner here. Had quite a bit of back and forth in my head trying to figure out what ID and person mapped to what ("oldPerson", "newPerson", "mergeInto", "otherPerson", "previousDistinctId", "distinctId"), so I like the explicitness in the call

src/worker/ingestion/process-event.ts
yakkomajuri: Should still be valid, as a distinct ID can be added in between `moveDistinctIds` and `deletePerson`, in which case we need to move the IDs again so we can delete without an error.

macobo (approved):
Haven't tested this but code-wise this makes sense and no bit-flipping happening as far as I can tell :)

yakkomajuri:
I really should put some effort into deflaking these tests 😠

Issue #526, opened 2021-08-13
Insert distinct ID error


yakkomajuri:
Seems we fail multiple times a second on Cloud to insert a distinct ID into Postgres. 

This makes sense, given a user that triggers an event is likely to trigger multiple and we see a ton of new users (distinct IDs) every day, meaning it's quite likely two events for the same non-existent ID will end up on different threads at the same time. 

We do [handle this properly](https://github.com/PostHog/plugin-server/blob/8e9d0f0155dcb4ea457861dabed9a1c8399abb56/src/worker/ingestion/process-event.ts#L291) but:

a) we swallow errors 
b) we're doing a lot of supposedly unnecessary queries and relying heavily on erroring as default behavior

I believe there should be ways to improve this. Have a couple ideas but mostly just documenting for now.

yakkomajuri:
Somewhat handled in #527

PR #527, opened 2021-08-13
nicer handling of insert distinct id failure

yakkomajuri:
## Changes

I did some digging and turns out Postgres errors like 20-40 times per minute trying to insert a distinct ID that already exists. More context on #526 

The plugin server handles this properly yet it swallows the errors, while Postgres creates a bit of logspam for what is an expected error for our system. This is similar to #523

Ultimately, this means that on Cloud we're doing tens of thousands of unnecessary queries a day just for distinct ID inserts, but I think any potential solution at this stage might just be over-engineering and would still be prone to race conditons, even if it shortens the window for them to happen.

Anyway, so this just silences unnecessary Postgres errors while making sure the errors we do care about are actually captured in Sentry.

## Checklist

-   [ ] Jest tests

### DIFF
--- a/src/utils/db/db.ts
+++ b/src/utils/db/db.ts
@@ -570,9 +570,14 @@ export class DB {
         distinctId: string
     ): Promise<ProducerRecord | void> {
         const insertResult = await client.query(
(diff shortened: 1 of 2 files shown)


Twixes (approved):
Makes sense to me

mariusandra (approved):

PR #530, opened 2021-08-17
create person from session recording events

yakkomajuri:
## Changes

One of the data issues identified last sprint.

## Checklist

-   [ ] Jest tests

### DIFF
--- a/src/worker/ingestion/process-event.ts
+++ b/src/worker/ingestion/process-event.ts
@@ -119,7 +119,8 @@ export class EventsProcessor {
                         distinctId,
                         properties['$session_id'],
                         ts,
-                        properties['$snapshot_data']
+                        properties['$snapshot_data'],
+                        personUuid
                     )
                     this.pluginsServer.statsd?.timing('kafka_queue.single_save.snapshot', singleSaveTimer, {
                         team_id: teamId.toString(),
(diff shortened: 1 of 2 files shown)


yakkomajuri:
friendly ping @macobo

macobo:
LGTM. Worth pulling in someone from #team-platform as well to spread context? (cc @tiina303?)

src/worker/ingestion/process-event.ts
```diff
@@ -607,4 +604,21 @@ export class EventsProcessor {
...
+        sentAt: DateTime,
+        personUuid: string
+    ): Promise<void> {
+        const isNewPerson = await this.personManager.isNewPerson(this.db, teamId, distinctId)
+        if (isNewPerson) {
+            // Catch race condition where in between getting and creating, another request already created this user
+            try {
+                await this.db.createPerson(sentAt, {}, teamId, null, false, personUuid.toString(), [distinctId])
+            } catch (error) {
+                Sentry.captureException(error, { extra: { teamId, distinctId, sentAt, personUuid } })
```
macobo, line 620: Is this worth notifying sentry of?

macobo (approved):

src/worker/ingestion/process-event.ts
yakkomajuri: It's probably a bit of noise, but I'd like to have at least a bit of context on Sentry about all these race conditions we handle. This context could be helpful the next time we have a race condition issue to debug.

macobo: Ack. Worth keeping this in mind though since sentry does have an error rate limit, very frequent errors will cause us other issues :)

tiina303 (approved):

Issue #371, opened 2021-05-10
Ingest events backwards


mariusandra:
Whenever we ingest events, we end up with entries in 3 tables: `events`, `persons` and `person_distinct_ids`.

Excluding plugins for now, if we ingest the same events again in the same order, we will (most likely) end up with the same entries in those tables.

However, I think we should also get exactly the same entries in those tables if we ingest events... backwards. If we do, it would make ingestion, retries and exports a lot simpler. For example, we could export multiple months in parallel from one posthog instance to another and know that all persons were transferred as they are.

This task touches two things: 
- Store person property updated_at timestamps... and then compare them before making any changes (in [denormalized properties](https://github.com/PostHog/posthog/issues/4272)?). We need this anyway for $set_once issues, but this could also be added to $set (if a property has been set by an event in the future, don't reset it to something if this event is in the past).
- Store some metadata on when users were merged and use that to prevent past changes from overriding future changes (e.g. somehow prevent renaming the new user to the old user)

The details and feasability of the above points need to be specced out.

yakkomajuri:
@tiina303

tiina303:
A case for importance (being able to replay events later is already called out and we saw a need for that this week), but additionally network is unpredictable - sometimes a packet sent later will arrive earlier, sometimes an earlier packet fails and will be retried later.

For `set` we can just look at timestamps, however for `set_once` we can't as it shouldn't overwrite anything that was earlier, so imagine:
```
Timestamp | function | value
1         | set      |  1
2         | set_once |  2
```
Regardless of the order of arrival we want the value to be 1 in the end. If 2 is sent first, we need to know it was a `set_once` call that set it to know that we should overwrite it with the earlier value.

Furthermore this needs to also end up with val = 1 even if packet 1 is sent later.
```
Timestamp | function | value
1         | set_once |  1
2         | set_once |  2
```
I propose we keep an additional column `set_once:bool` that tells us if it was set by `set_once`, if it was and we see an earlier `set` or an earlier `set_once`, we need to overwrite it and update the timestamp to earlier.

[...]

PR #535, opened 2021-08-19
Export historical events

yakkomajuri:
## WIP

Everything's subject to change, the approach, the names, the defaults, etc.

## To Do

- [x] Validate approach
- [x] Settle on naming
- [x] Implement job triggering from the interface
- [x] Implement kill switch
- [x] Add tests

## How this works

So this is a bit complicated as it makes heavy use of Redis and Postgres (with respect to queries, not memory/storage size), to ensure that:

- Every event gets exported (even if the server restarts)
- We export in a performant way using all threads efficiently

Approach is yet to be validated/refined but this works and works quite well as it stands.

I left a lot of comments in to explain how this works but either way, here's the basics:

1. An export is triggered
2. We fetch the min and max timestamps for the team's events. We use these to fetch and export events in intervals, like so:

Screenshot 2021-08-19 at 14 20 45

3. We trigger a job (currently one per thread, actually) to kick start things. This job tells the thread who picks it up that it's time to figure out an interval to start processing events from.
4. The job is picked up by a thread which sees it doesn't have an interval, so it should get one. It sends a `INCR` command to Redis and that specifies to it the lower boundary of the timestamp interval it should cover.
5. The thread runs a query to get events starting at the lower boundary it determined via Redis and orders them by `id` in Postgres or `_offset` in CH. The query also comes with `LIMIT` and `OFFSET` clauses, so the thread is processing e.g. the first 100 events.
6. All goes well, so the thread pops a job in the queue to process the next 100 events in this timestamp interval. 
7. Another thread picks up this job and finds that for timestamp interval X->X+10, there are no events with `OFFSET 100`. It then queues a job specifying that this interval is done, and it's time to pick up a new one. 
8. Everything starts again from 3.

<details>

<summary>Hidden section as this might not be relevant anymore </summary>

<br />
Beyond the above flow, there are other things handled, like plugin server restarts. Here's a TL;DR of how those work:

1. We start processing timestamp interval X->X+10, which has 1000 events. We add this interval to a Redis list of "pending intervals" - intervals we've started on but haven't finished.
2. Every time we're done with a batch of 100 events **within that interval**, we store the next offset we will process (in Postgres). 
3. If we're done with the interval as a whole, we get rid of the offset stored in Postgres.
4. If the plugin server restarts, during `teardownPlugin` we pull that list of unfinished intervals from Redis and add it to Postgres. 
5. When the plugin server comes back up, we tell it to start processing from the last cursor we stored. **However,** we  also take the unfinished list that was in Postgres and put it in Redis (so we can do things like `pop` elements) and we will **always prioritize unfinished intervals** from earlier to picking up a new interval from Redis. Here are the possible decisions:

Screenshot 2021-08-19 at 14 55 43

So ultimately, here's what's stored where:

**Redis**

| What | Type | Purpose |
| :----: |  :----: | :--------: |
| Interval Counter | Integer | Determining what's the next interval to process |
| Intervals being processed | List | Will be stored in Postgres if the server restarts so we can still process them |
| Unfinished intervals | List | Left over intervals from before the restart |

**Postgres**

| What | Type (Conceptually, not literally) | Purpose |
| :----: |  :----: | :--------: |
| Timestamp Cursor | UNIX Timestamp | Last interval we started to process |
| Unfinished intervals | List | Intervals we didn't finish processing. Only kept until the plugin starts running again, at which point we keep this in Redis. |
| Offset (one per interval in the list above) | Integer | Offset to start processing from for intervals we didn't finish. |
| Max Timestamp | UNIX Timestamp | Set once on the first run and persists across restarts. Ensures we only export historical data, since the plugin will probably start exporting live events once it's on. |

</details>


fuziontech:
This is really solid but I'm wondering why we didn't just use Graphile for this since it has so much already baked in for managing task executions already.
https://github.com/graphile/worker#features

A simple way to leverage this would be to:
- Have whatever is responsible for triggering this break the job into chunks much like you are doing
- Enqueue each chunk onto graphile
- Have a worker pool on graphile that is just responsible for chewing through the export queue. That way you can scale that up and and down

If you do that if things are shut down the workers just boot back up and start pulling from the queue of work. Exponential backoff is handled for poison tasks. You don't have one bad payload block the entire flow potentially. It would make this a lot more vanilla in implementation too without having to have these new data structures in redis and in pg.

yakkomajuri:
All of this is essentially Graphile under the hood :D @fuziontech

yakkomajuri:
I think if there was a guarantee that for each currently running interval there would be a job stored (via Graphile) by the time the plugin is tore down then we could do away with some of the unfinished intervals lists, but I'm not sure that's certain.

yakkomajuri:
So exponential backoff and poisoned payloads are indeed handled here. But actually I think you're right and there are some properties of our job queue that could be leveraged to do away with the unfinished lists at least.

yakkomajuri:
@fuziontech haven't tested yet but indeed I believe we might be able to [do this](https://github.com/PostHog/plugin-server/pull/535/commits/73b71ceff2b1919dc0a1a9b1bb635ea053cb78e8) and the queue we already have in place will handle the rest.

A job always queues another job so the job queue should always have a job for each interval. There's a possible scenario where we quit piscina mid-job so we never enqueue another but this should be no worse than what `teardownPlugin` would be subject to in the previous approach

mariusandra:
Hey, solid work indeed! Commenting a few higher level points in no specific order:

1) I see you're currently exporting a `exportFromTheBegining` function from a plugin. Is this just a stopgap? I've always seen it such that you still only write the same `exportEvents = (events) => fetch('/api', events)` function as now, and then the posthog lets you tick a "also export from the beginning" checkbox when setting up the plugin. This would then trigger whatever is needed in the plugin server to do the job.

2) "Export from the beginning" is actually an import plugin! We're just importing from our own database. I see a lot of parallels with both tasks: both get a range of data to query, split that range up between workers, and then querying the data and send it somewhere. Thus I think the right level of abstraction is that of an import/export jobs. Said differently, `create table plugin_importexportjobs`, which somehow keeps track of the progress of all import or export jobs, including how many jobs are done and how many events already exported/imported. 

3) @fuziontech The reason why we took retries away from graphile and write them ourselves, is because jobs can have multiple backends. In case graphile is down, we want to use the slow backup S3 queue. Or a backup graphile queue. Data consistency and integrity first and foremost. 

4) That said, maybe the same considerations do not apply here. I definitely don't want someone's export jobs clogging up graphile and forcing all my plugins onto the S3 queue. I also don't think it's efficient to use S3 as a backend for this job queue. So having a separate pool/group/queue on graphile definitely makes sense. Retries can then be performed by graphile as well, though we might want to record the status of retries ourselves if we go for the suggestion in point 2.

5) Redis's `incr` is awesome and easy to explain in plugins, but if you want you can build something similar with just postgres as well. `UPDATE plugin_importexportjobs SET interval_number = interval_number + 1 WHERE id = 43 RETURNING interval_number`. Redis is still a cache that can go `poof` at any moment and shouldn't be considered a database for storing data for long running operations.

fuziontech:
Re: 3 - I totally agree. You all did amazing work making plugin server resilient to backups and failures. This workflow is just a bit different as it's not part of the ingestion pipeline. The events are already persisted. We are just processing them from one backend to another.

Re: 4 - Exactly. This builds a lot of queuing functionality on top of what is (a really fantastic) queue. I just don't think it belongs on the queue when it could just be enqueued on a separate, resource isolated and monitored queue. Not a huge deal, but something that seems like it would be easier to manage if there was a fire.

Re 5: I don't even think you need to have an offset persisted. If you have tasks with the signature `function export_block(offset, limit, order_by)` you can enqueue a bunch of blocks from start to end by calling that on a loop for each block to ship. The workers will then just chip away at that. The delta between offset and limit would be determined by the batch size. The risk there would be that a batch may be retried which means that this would only guarantee at least once delivery of the messages. If you wanted exactly once guarantees you'd need to have the offset per batch recorded somewhere. I think that should be ok to live in redis since it is ephemeral to the batch. A failure of redis wouldn't impact the processing of these events (if protected from redis failure), but a failure of the batch could pick up from the last spot pulling from redis. Avoiding PG would be nice since I don't think it's worth hitting it for every event emitted just to set a checkpoint, but I'm not 100% against it.

mariusandra:
Just enqueueing all ranges when starting an export makes sense, though sounds a bit risky. We need to make sure we don't get `len($events_table)` errors immediately after someone misconfigures their export source. Though, it would make many things easier.

yakkomajuri:
@mariusandra 

> I see you're currently exporting a exportFromTheBegining function from a plugin

I was thinking if these events should just be piped into `exportEvents` but wondered if a plugin might not want to do some different processing to historical data? Maybe add a prop? Might actually just be unnecessary complexity, so should cast the events to `PluginEvent` instead `Event` to do this.

> Thus I think the right level of abstraction is that of an import/export jobs.

👍 

That said, maybe the same considerations do not apply here. I definitely don't want someone's export jobs clogging up graphile and forcing all my plugins onto the S3 queue. I also don't think it's efficient to use S3 as a backend for this job queue. So having a separate pool/group/queue on graphile definitely makes sense. Retries can then be performed by graphile as well, though we might want to record the status of retries ourselves if we go for the suggestion in point 2.

> Redis's incr is awesome and easy to explain in plugins, but if you want you can build something similar with just postgres as well.

Yeah I'm aware of this. We still do store the cursor in Postgres and it just felt natural to do increment in Redis. But yes, could consolidate into one.

yakkomajuri:
> Just enqueueing all ranges when starting an export makes sense, though sounds a bit risky. 

No, not all ranges! In the current `setupPlugin` trigger we are enqueueing one range per thread. So the ranges in graphile will be between 1 and N threads depending on the final implementation. What @fuziontech made me realize is that this is enough to pick up from where we left off, as long as the cursor is stored in Postgres. No need for keeping manually keeping track of lists with unfinished ranges as I was doing before.

yakkomajuri:
Still untested for now, but we now do all the cursor operations directly in Postgres and we leverage the existing `exportEvents` for this:

https://github.com/PostHog/plugin-server/pull/535/commits/ee2efb4b1025c76ccb1888dc139a8adb8ae74782

neilkakkar:
Maybe I'm just completely out of the loop: At this stage, do all UI operations / API calls that operate on events go through the plugin server? This wasn't the case until a while ago, which was a big blocker for exports from the beginning.

yakkomajuri:
@neilkakkar I'm confused as to what you mean? What would block us from exporting events from the plugin server?

neilkakkar:
Let's say `mergePersons` or `alias` can be done without generating an event for plugin server (I think these were moved to events being handled by plugin server recently - but bear with the example).

So, now events having distinctIDs as the old `alias` values become orphaned, since they don't know about the `alias` event happening (since there was no event published for it), which means the exported events are inconsistent, so if we were to pipe those into a new PostHog instance, we'd get the wrong results.

Kinda ties into concerns with ImportEvents as well.

yakkomajuri:
@neilkakkar I see what you mean. Indeed this _was_ a problem (meaning it still is for historical data), but this implementation doesn't concern itself with this. This helps you get events out and that's about it.

What you're talking about is part of a larger issue that we need to take various steps to fix: #371

yakkomajuri:
I'll be writing tests for this today and then we can get it in and iterate from there. The interface is live but hidden behind feature flag so there's no explicit way to trigger this

yakkomajuri:
ping @tiina303 I think we should get this in before it gets stale

fuziontech (approved):
:shipit:

PR #537, opened 2021-08-23
Trigger plugin jobs via Celery

yakkomajuri:
## Changes

Allows plugin jobs to be triggered via the PostHog UI. This needs a supporting PR to `posthog` that's coming soon.

How it works:

- An endpoint in `posthog` receives the job specs (job name, payload) via a POST request
- The request triggers sending a task to Celery which is picked up by the plugin server
- The plugin server enqueues the job to be run as soon as possible, but still going through the job queues (instead of skipping to `runJob` directly)

A few things were built in on the `posthog` side as well as here that aren't yet natively supported but will help us going forward. The most important of these is the job operation, which is currently just being dumped in the payload if devs want to do anything with it. In the future this will allow us to do things like start and stop longer running jobs like imports and exports. 

## Checklist

-   [ ] Jest tests

### DIFF
--- a/src/main/ingestion-queues/queue.ts
+++ b/src/main/ingestion-queues/queue.ts
@@ -11,7 +11,7 @@ import { KafkaQueue } from './kafka-queue'
 
 interface Queues {
     ingestion: Queue
-    redis: Queue
+    auxiliary: Queue
 }
 
 export function pauseQueueIfWorkerFull(
@@ -57,7 +57,7 @@ export async function startQueues(
         const redisQueue = startQueueRedis(server, piscina, mergedWorkerMethods)
         const queues = {
             ingestion: redisQueue,
-            redis: redisQueue,
+            auxiliary: redisQueue,
         }
         if (server.KAFKA_ENABLED) {
             queues.ingestion = await startQueueKafka(server, piscina, mergedWorkerMethods)
--- a/src/main/ingestion-queues/queue.ts
+++ b/src/main/ingestion-queues/queue.ts
@@ -2,7 +2,7 @@ import Piscina from '@posthog/piscina'
 import { PluginEvent } from '@posthog/plugin-scaffold'
 import * as Sentry from '@sentry/node'
 
-import { Hub, Queue, WorkerMethods } from '../../types'
+import { CeleryTriggeredJobOperation, Hub, PluginConfig, Queue, Team, WorkerMethods } from '../../types'
 import { status } from '../../utils/status'
 import { sanitizeEvent, UUIDT } from '../../utils/utils'
 import { CeleryQueue } from './celery-queue'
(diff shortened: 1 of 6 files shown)


src/main/ingestion-queues/queue.ts
```diff
@@ -51,48 +51,76 @@ export async function startQueue(
...
             try {
-                const checkAndPause = () => pauseQueueIfWorkerFull(() => celeryQueue.pause(), server, piscina)
-                await ingestEvent(server, workerMethods, event, checkAndPause)
+                payload['$operation'] = jobOp
+                const job = {
+                    type,
+                    payload,
+                    pluginConfigId,
+                    pluginConfigTeam,
+                    timestamp: new Date().getTime(),
```
Twixes, line 82: Nit! Isn't `Date.now()` more direct?

Twixes (commented):
Substance looks good to me.
Tests failing, but these seem unrelated and should be fixed by #538.

src/types.ts
```diff
@@ -686,3 +686,5 @@ export interface PropertyDefinitionType {
 }
 
 export type PluginFunction = 'onEvent' | 'processEvent' | 'onSnapshot' | 'pluginTask'
+
+export type CeleryTriggeredJobOperation = 'start' // new ops go here
```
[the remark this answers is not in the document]
yakkomajuri: Yeah enum makes sense. I guess there's just one op for now but will add more

src/main/ingestion-queues/queue.ts
```diff
@@ -51,48 +51,76 @@ export async function startQueue(
...
 }
 
-function startQueueRedis(server: Hub, piscina: Piscina | undefined, workerMethods: WorkerMethods): Queue {
+function startQueueRedis(server: Hub, piscina: Piscina, workerMethods: WorkerMethods): Queue {
     const celeryQueue = new CeleryQueue(server.db, server.PLUGINS_CELERY_QUEUE)
 
+    // this queue is for triggering plugin jobs from the PostHog UI
     celeryQueue.register(
-        'posthog.tasks.process_event.process_event_with_plugins',
+        'posthog.tasks.plugins.plugin_job',
```
neilkakkar, line 67: Doesn't this now imply that `process_event_with_plugins` tasks on a non-kafka installation won't work?

Twixes: It does, but they are never sent in such a case. I think we only accepted them out of laziness/migration-tolerance, but it doesn't seem that ever was useful in practice.

neilkakkar: Ah, nevermind, it's **!**`!server.KAFKA_ENABLED` below.

src/main/ingestion-queues/queue.ts
```diff
@@ -51,48 +51,76 @@ export async function startQueue(
     try {
         if (server.KAFKA_ENABLED) {
             return await startQueueKafka(server, piscina, mergedWorkerMethods)
```
neilkakkar, line 53: The `return` doesn't make sense here anymore, does it? I think you want `startQueueRedis` always?

src/main/ingestion-queues/queue.ts
yakkomajuri: Yeah so before we just didn't run the celery queue at all if you had Kafka, as the plugin server only used it for events. 

Now we always run it, but don't listen to the events queue if Kafka is on.

This is git playing tricks on you as I didn't edit that line, I added a whole block of code that was based on the one that was there :D

src/main/ingestion-queues/queue.ts
```diff
@@ -49,50 +49,79 @@ export async function startQueue(
     }
 
     try {
+        const redisQueue = startQueueRedis(server, piscina, mergedWorkerMethods)
         if (server.KAFKA_ENABLED) {
             return await startQueueKafka(server, piscina, mergedWorkerMethods)
-        } else {
-            return startQueueRedis(server, piscina, mergedWorkerMethods)
         }
+        return redisQueue
```
neilkakkar, line 56: Since we want to do operations on drain: https://github.com/PostHog/plugin-server/blob/master/src/main/pluginsServer.ts#L138

Maybe makes sense to return a list of queues?

yakkomajuri: This is actually a valid comment, but oh damn it's gonna be quite a mess

src/main/ingestion-queues/queue.ts
```diff
@@ -49,50 +49,79 @@ export async function startQueue(
...
-            distinct_id: string,
-            ip: string | null,
-            site_url: string,
-            data: Record<string, unknown>,
-            team_id: number,
-            now: string,
-            sent_at?: string
+            pluginConfigTeam: Team['id'],
+            pluginConfigId: PluginConfig['id'],
+            type: string,
```
Twixes, line 72: This is called `name` in https://github.com/PostHog/posthog/pull/5687, so maybe here as well?

yakkomajuri: Yeah probably the other way around. I wrote the endpoint there first and then here I was trying to be consistent with the terminology in place for jobs. Will update there.

src/main/ingestion-queues/queue.ts
yakkomajuri: I opted for keeping `queue` what it was: the ingestion queue. This is important because while we can do things like:

```
for (const queue of queues) {
    queue.resume()
}
```

... we can't really do the same for `isPaused`. So `queue` remains what it was + we now have a `redisQueueForKafkaIngestion` which is the jobs queue and is only needed for kafka given on celery ingestion the plugin jobs and ingestion queues are the same.

neilkakkar: 🙈 . Thanks!

src/main/pluginsServer.ts
```diff
@@ -134,9 +136,19 @@ export async function startPluginsServer(
...
-        queue = await startQueue(hub, piscina)
+        const queues = await startQueues(hub, piscina)
+
+        // `queue` refers to the ingestion queue. With Celery ingestion, we only
+        // have one queue for plugin jobs and ingestion. With Kafka ingestion, we
+        // use Kafka for events but still start Redis for plugin jobs.
+        // Thus, if Kafka is disabled, we don't need to call anything on
+        // redisQueueForKafkaIngestion, as that will also be the ingestion queue.
+        queue = queues.ingestion
+        redisQueueForKafkaIngestion = config.KAFKA_ENABLED ? queues.redis : null
```
neilkakkar, line 147: re: naming, isn't this `redisQueueForPluginJobs` ?

yakkomajuri: Well that's the mess here. The redis queue is both the ingestion queue and the plugin jobs queue. But when we don't have Kafka enabled, the ingestion queue === the redis queue, so we shouldn't be calling methods on it twice.

When Kafka is enabled, on the other hand, we set this queue, which will like you said only be for plugin jobs. So I wanted the naming to make this explicit but not sure if it helped.

neilkakkar: Ahhhh. `redisQueueForPluginJobsOnKafka` ? 😂 

Hmmm, thinking a bit more, I think the confusion stems from the two queues being called `ingestion` and `redis`? The former name is "what is the queues fuction" and the latter name is "what type of queue it is".

Since on kafka we have two queues, and only one queue on non-Kafka, how about calling the queues `main` and `auxillary` ? and then the auxillary queue name can be `redisQueueForPluginJobs` ? It becomes clear, thanks to Line 147, that the auxillary queue is only set on `config.KAFKA_ENABLED`.

neilkakkar: .. This might feel pretty trivial, but I think it's important to get this right: we have a million queues already, and it can get confusing for new-comers to understand what goes where if we don't name things properly.

Apologies for the pedantry 😅 .

yakkomajuri: Not a problem at all! That's what reviewers are for. Thanks for upholding this

src/main/ingestion-queues/queue.ts
yakkomajuri: I call the variable type over there now, but kept the API as `name` since that's friendlier. "Job type" is confusing

yakkomajuri:
Addressed @neilkakkar

neilkakkar (approved):
Thanks!

PR #543, opened 2021-08-24
Add meta utils, cursor ops

yakkomajuri:
## Changes

Tests incoming.

Adds a new object to the plugin meta: `utils`. This way meta doesn't need to grow out of control and we can more neatly let `utils` grow as much as it needs to adding helper functions for plugin devs.

This also adds the first layer of generalized utils stemming from #535: cursors.

## Checklist

-   [ ] Jest tests


yakkomajuri:
I'd get this in without a version bump, then let the update in `plugin-scaffold` trigger the update, after which I'll remove the type here.

PR #545, opened 2021-08-24
Expose `@google-cloud/pubsub` to plugins

jredl-va:
## Changes

Adds support for the google cloud pub/sub package to continue work in flight of publishing events to a google cloud pub/sub topic. 

That said work is currently located here: https://github.com/vendasta/pubsub-plugin


fuziontech (approved):
Looks good to me!

PR #555, opened 2021-09-01
WIP Group Analytics

yakkomajuri:
## Changes

Support `$group` events from `posthog.group` and `$groups` in events 

## Checklist

-   [ ] Jest tests


yakkomajuri:
Should probably be fine to get this in and get the ball rolling on group analytics.

Maybe with some additional guard?

Twixes:
I would say yes, but only once we revive a group analytics branch in the main repo, since otherwise `posthog_grouptypemapping` etc. don't existing making this pointless.

PR #558, opened 2021-09-05
Fix intitial params

timgl:
## Changes

- There was a bug where `$initial_current_url` (and browser, os etc) would only be set if a `utm_` tag was present.
- I'm now checking which properties we are _actually_ updating and only storing those against the event. Should _massively_ reduce the amount of data we need to store in events, plus the amount of data we're sending to Postgres to update the user
- **Open question:** if we merge this _everyone_ will start getting initial_current_url, but of the page that they visited most recently after merging this. Maybe we should hardcode in a date after which the user needs to be created or something?
- Alternatively, given no-one had noticed this was broken for six months, should we just remove inititial_current_url etc, given that it'll be in a broken state no matter what? we'll keep utm_campaign

## Checklist

-   [ ] Jest tests


zhicong:
Thanks for fixing this! The ability to track `$initial_current_url` automatically was a huge draw for us in installing Posthog, so please don't remove it. It helps us understand where our organic signups are coming from, and perhaps you might want to publicize this feature more for greater usage.

src/worker/ingestion/process-event.ts
```diff
@@ -268,14 +288,15 @@ export class EventsProcessor {
...
-            return personFound
+            return returnedProps
         }
 
         if (incrementedPropertiesQueryResult && incrementedPropertiesQueryResult.rows.length > 0) {
             updatedProperties = { ...updatedProperties, ...incrementedPropertiesQueryResult.rows[0].properties }
         }
 
-        return await this.db.updatePerson(personFound, { properties: updatedProperties })
+        await this.db.updatePerson(personFound, { properties: updatedProperties })
```
mariusandra, line 298: If I read this right, then calling `await this.db.updatePerson(personFound, { properties: updatedProperties })` ... will store as the person's properties only the updated properties, overridding all existing properties... unless an `$increment` property set.

src/worker/ingestion/process-event.ts
```diff
@@ -254,7 +254,27 @@ export class EventsProcessor {
             )
         }
 
-        let updatedProperties: Properties = { ...propertiesOnce, ...personFound.properties, ...properties }
+        // Figure out which properties we are actually setting
+        const returnedProps: Properties = {}
+        let updatedProperties: Properties = {}
```
mariusandra, line 259: The `returnedProps` logic is solid, but `updatedProperties` should stay how it was and contain all the properties, not just the changed ones. Otherwise `arePersonsEqualExcludingIncrement` will always be false below and we will run into data loss :party_parrot:

src/worker/ingestion/process-event.ts
```diff
@@ -472,14 +493,17 @@ export class EventsProcessor {
...
                 teamId,
                 distinctId,
                 properties['$set'] || {},
                 properties['$set_once'] || {},
                 filteredIncrementProperties
             )
+
+            delete properties['$set']
+            delete properties['$set_once']
+            properties = { ...properties, ...updatedProperties }
```
mariusandra, line 506: I'd also add a code comment here, that `$set` and `$set_once` come back inside `updatedProperties`, otherwise it reads like they are removed from being stored on an event.

Alternatively, rename `updatedProperties` to `updatedSetAndSetOnce` and let the code speak for itself 🎶

mariusandra (commented):
> I'm now checking which properties we are actually updating and only storing those against the event. Should massively reduce the amount of data we need to store in events, 

This makes sense!

> plus the amount of data we're sending to Postgres to update the user

This introduces a data loss bug :). No need to optimise here. Comments inline.

> Open question: if we merge this everyone will start getting initial_current_url, but of the page that they visited most recently after merging this. Maybe we should hardcode in a date after which the user needs to be created or something?

I'm in favour of merging it in as is.

> Alternatively, given no-one had noticed this was broken for six months, should we just remove inititial_current_url etc, given that it'll be in a broken state no matter what? we'll keep utm_campaign

I think these are very useful for web analytics.

mariusandra (changes requested):
Nice find, though I think the implementation will

timgl:
@mariusandra good to merge? These tests seem broken generally

src/worker/ingestion/process-event.ts
```diff
@@ -254,7 +254,28 @@ export class EventsProcessor {
             )
         }
 
-        let updatedProperties: Properties = { ...propertiesOnce, ...personFound.properties, ...properties }
+        // Figure out which properties we are actually setting
+        const returnedProps: Properties = {}
+        let updatedProperties: Properties = {}
```
mariusandra, line 259: I'd simplify this to:
```js
updatedProperties = { ...personFound.properties } // copy props, will mutate below
```
because then I could remove this more confusing line below:
```js
updatedProperties = { ...personFound.properties, ...updatedProperties }
```

tests/shared/process-event.ts
```diff
@@ -222,7 +222,7 @@ export const createProcessEventTests = (
                     $browser: 'Chrome',
                     $current_url: 'https://test.com',
                     $os: 'Mac OS X',
-                    $browser_version: 80,
+                    $browser_version: false,
```
mariusandra, line 225: Why this change?

mariusandra (commented):
Few more comments. I'm not sure what's the deal with the tests though... :/

tests/shared/process-event.ts
timgl: This tests the case you highlighted where $set_once would be set again if the property was false. It also tests the other issue you highlighted with partial saves.

src/worker/ingestion/process-event.ts
timgl: Done

timgl:
@mariusandra tests pass, are you happy with this?

PR #559, opened 2021-09-06
Fix tests by :any-ing catch errors

timgl:
## Changes

There was a change in typescript that means all catch error are type 'unkown' https://github.com/Microsoft/TypeScript/issues/20024

## Checklist

-   [ ] Jest tests

### DIFF
--- a/src/healthcheck.ts
+++ b/src/healthcheck.ts
@@ -13,7 +13,7 @@ export async function healthcheck(): Promise<boolean> {
             status.error('💔', 'Redis key @posthog-plugin-server/ping not found! Plugin server seems to be offline')
             return false
         }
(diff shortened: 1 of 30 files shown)


Twixes (commented):
Good catch, we saw these failures with @yakkomajuri during the hackathon as well, but didn't have time to investigate. This is a valid hotfix.
However this should also not been a problem with CI, because CI _should_ be using `tsc` from `package.json`, not what the GH Actions runner has globally. That part of the problem is fixed by https://github.com/PostHog/plugin-server/pull/560.
Both this and that PR get rid of the errors.

timgl:
Yes lets close this in favour of #560

PR #561, opened 2021-09-07
fix($identify): do not alias same anon id to multiple identified users

hazzadous:
This arises when we have this sequence of events:

```
posthog.capture('event1')
posthog.identify('user1')
posthog.capture('event2')
posthog.identify('user2')
posthog.capture('event3')
```

We want to ensure we associate the anonymous 'event1' with 'user1' when
they are identified, but we do not want to alias 'user1' and 'user2' as
these are separate users. With the previous logic we would associate
both 'user1' and 'user2' with the anonymous id associated with 'event1'
and therefore by transitivity, 'user1' and 'user2' would be associated.

Closes https://github.com/PostHog/posthog/issues/5527

### DIFF
--- a/src/worker/ingestion/process-event.ts
+++ b/src/worker/ingestion/process-event.ts
@@ -192,9 +192,8 @@ export class EventsProcessor {
         if (event === '$create_alias') {
             await this.alias(properties['alias'], distinctId, teamId)
         } else if (event === '$identify') {
-            const anonymousDistinctId = properties['$anon_distinct_id']
-            if (anonymousDistinctId) {
-                await this.alias(anonymousDistinctId, distinctId, teamId)
+            if (properties['$anon_distinct_id']) {
+                await this.alias(properties['$anon_distinct_id'], distinctId, teamId)
             }
             await this.setIsIdentified(teamId, distinctId)
         }
--- a/src/worker/ingestion/process-event.ts
+++ b/src/worker/ingestion/process-event.ts
@@ -310,7 +310,10 @@ export class EventsProcessor {
         const newPerson = await this.db.fetchPerson(teamId, distinctId)
 
         if (oldPerson?.is_identified) {
-            console.log('Person associated with previousDistinctId is identified, refusing to alias new distinctId')
+            console.log('Person associated with previousDistinctId is identified, refusing to alias new distinctId', {
+                distinctId,
+                previousDistinctId,
+            })
             return
         }
 
--- a/src/worker/ingestion/process-event.ts
+++ b/src/worker/ingestion/process-event.ts
@@ -192,46 +192,11 @@ export class EventsProcessor {
         if (event === '$create_alias') {
             await this.alias(properties['alias'], distinctId, teamId)
         } else if (event === '$identify') {
-            await this.handleIdentify(event, properties, distinctId, teamId)
-        }
-    }
-
-    private async handleIdentify(
-        event: string,
-        properties: Properties,
-        distinctId: string,
-        teamId: number
-    ): Promise<void> {
-        // Ensure that if the person identified by "$anon_distinct_id" is
-        // identified, that we do not alias to `distinctId`, as this would
-        // be merging two distinct identified persons. See
-        // https://github.com/PostHog/posthog/issues/5527 for discussion on
-        // the details of this guard is protecting against
-        const previousDistinctId = properties['$anon_distinct_id']
-        if (previousDistinctId) {
-            const previousPerson = await this.db.fetchPerson(teamId, previousDistinctId)
-            if (previousPerson?.is_identified) {
-                console.log(
-                    'Person associated with $anon_distinct_id is identified, refusing to alias new distinctId',
-                    {
-                        distinctId,
-                        previousDistinctId,
-                    }
-                )
-            } else {
-                await this.alias(
-                    previousDistinctId,
-                    distinctId,
-                    teamId,
-                    true,
-                    0,
-                    // Ensure that any persons created by alias are created as
-                    // identified, after all, this is the identify event handling
-                    true
-                )
+            if (properties['$anon_distinct_id']) {
+                await this.alias(properties['$anon_distinct_id'], distinctId, teamId)
             }
+            await this.setIsIdentified(teamId, distinctId)
         }
-        await this.setIsIdentified(teamId, distinctId)
     }
 
     private async setIsIdentified(teamId: number, distinctId: string, isIdentified = true): Promise<void> {
@@ -339,11 +304,19 @@ export class EventsProcessor {
         distinctId: string,
         teamId: number,
         retryIfFailed = true,
-        totalMergeAttempts = 0,
-        isIdentified = false
+        totalMergeAttempts = 0
     ): Promise<void> {
         const oldPerson = await this.db.fetchPerson(teamId, previousDistinctId)
         const newPerson = await this.db.fetchPerson(teamId, distinctId)
+
+        if (oldPerson?.is_identified) {
+            console.log('Person associated with `previousDistinctId` is identified, refusing to alias new distinctId', {
+                distinctId,
+                previousDistinctId,
+            })
+
+            return
+        }
 
         if (oldPerson && !newPerson) {
             try {
@@ -354,7 +327,7 @@ export class EventsProcessor {
                 // integrity error
                 if (retryIfFailed) {
                     // run everything again to merge the users if needed
-                    await this.alias(previousDistinctId, distinctId, teamId, false, 0, isIdentified)
+                    await this.alias(previousDistinctId, distinctId, teamId, false)
                 }
             }
             return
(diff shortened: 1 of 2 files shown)


src/worker/ingestion/process-event.ts
```diff
@@ -192,8 +192,15 @@ export class EventsProcessor {
         if (event === '$create_alias') {
             await this.alias(properties['alias'], distinctId, teamId)
         } else if (event === '$identify') {
```
macobo, line 194: Couple of headscratchers:

- We have $create_alias which goes through the same path above. Thoughts on what to do with that?
- We are increasing the number of times we're fetching the person here. It's slightly expensive.
- We're asserting that person with anonymousDistinctId exists in this code path before aliasing, not sure if it's actually correct.

Given that `.alias` already looks up people, it seems instead worth moving this check to `.alias` into the branch that calls mergePeople.

Also let's log something (for now) when going non-anonymous -> non-anonymous.

tests/shared/process-event.ts
```diff
@@ -1273,6 +1273,80 @@ export const createProcessEventTests = (
         expect(person2.is_identified).toBe(true)
     })
 
+    test('we do not alias users if distinct id changes but we are already identified', async () => {
```
macobo, line 1276: It's also worth testing manually with stateful client libraries whether this behavior works.

E.g. Given:

event1
Identify to userA
event2
Identify to userB
event3

What will the result be, which events will userA and userB own?

src/worker/ingestion/process-event.ts
hazzadous: * For `$create_alias` this seems like an explicit request to alias, hence I left this as is. I can look into the expected usage a little more to see if there is anything we should do with this but it seems like this logic should stay as is.
* Yes it is making it more expensive. I considered moving this check into `.alias` where we're already making the db requests, but it seems like if you ask to alias two ids then it should alias them unconditionally.
* We presumably have to check the anonymous user before aliasing as this is what we want to act conditionally on, but perhaps I'm missing something?

I'll move the check into `.alias` and add some logging.

tests/shared/process-event.ts
hazzadous: Will add some events into the test and assert they are associated with the right Persons

src/worker/ingestion/process-event.ts
```diff
@@ -288,6 +289,11 @@ export class EventsProcessor {
         const oldPerson = await this.db.fetchPerson(teamId, previousDistinctId)
         const newPerson = await this.db.fetchPerson(teamId, distinctId)
 
+        if (oldPerson?.is_identified) {
```
macobo, line 292: I think this case might be too early in this chain but it might be worth it to chat through the consequences of this.

src/worker/ingestion/process-event.ts
```diff
@@ -288,6 +289,11 @@ export class EventsProcessor {
...
+            console.log('Person associated with previousDistinctId is identified, refusing to alias new distinctId')
```
macobo, line 293: Include distinct ids in here as well. You don't need to use interpolation, just add `{ distinctId, previousDistinctId }` as the last argument

tests/shared/process-event.ts
macobo: On the plugin server level the tests you added are :ok_hand: - we should still 100% check and verify the client libraries behavior as well though. 

This is a huge behavioral change and we want to be sure we understand it fully :)

src/worker/ingestion/process-event.ts
```diff
@@ -192,8 +192,9 @@ export class EventsProcessor {
         if (event === '$create_alias') {
             await this.alias(properties['alias'], distinctId, teamId)
         } else if (event === '$identify') {
-            if (properties['$anon_distinct_id']) {
-                await this.alias(properties['$anon_distinct_id'], distinctId, teamId)
+            const anonymousDistinctId = properties['$anon_distinct_id']
+            if (anonymousDistinctId) {
+                await this.alias(anonymousDistinctId, distinctId, teamId)
```
macobo, line 197: Nitpick: You can revert this part of the change :)

tests/shared/process-event.ts
hazzadous: Yup, I had a play with trying to get the plugin-server and posthog-js running locally but wasn't convinced it was on configured correctly.

One of the other libraries would be ideal as well as posthog-js already has the guard against the issue behaviour. I guess rudderstack would be a good candidate here?

hazzadous: Also, are there any integration tests I should add to also?

macobo: Other libraries are stateless, though I haven't checked out e.g. react native/ios libraries.

Perhaps the thing to do is temporarily remove the guard from posthog-js and try this out without it?

Re integration test suite: https://github.com/PostHog/posthog-js/blob/master/testcafe/e2e.spec.js does kind of this already, though it might be a bit hard to modify.

hazzadous: yup I'll have a go with disabling the guard and see if I can get it running against local.

src/worker/ingestion/process-event.ts
hazzadous: test: timing/concurrency issues

src/worker/ingestion/process-event.ts
```diff
@@ -192,11 +192,46 @@ export class EventsProcessor {
...
+        teamId: number
+    ): Promise<void> {
+        // Ensure that if the person identified by "$anon_distinct_id" is
+        // identified, that we do not alias to `distinctId`, as this would
+        // be merging two distinct identified persons. See
+        // https://github.com/PostHog/posthog/issues/5527 for discussion on
+        // the details of this guard is protecting against
+        const previousDistinctId = properties['$anon_distinct_id']
+        if (previousDistinctId) {
+            const previousPerson = await this.db.fetchPerson(teamId, previousDistinctId)
```
macobo, line 212: Why did we go back down to this route? :thinking: 

Couple of issues:
1. We're losing the property that "only anonymous users get merged"
2. We're increasing load on the DB due to repeated lookups

Worth discussing face-to-face?

tests/shared/process-event.ts
```diff
@@ -1273,6 +1336,212 @@ export const createProcessEventTests = (
...
+
+        //     // Now let's make sure there are no persons. In a production environment
+        //     // these identify tasks should be retried, or send to a "dead letter
+        //     // queue" or similar for inspection.
+        //     expect(await hub.db.fetchPersons()).toEqual([])
+        // })
+    })
+
+    describe('when handling $create_alias', () => {
+        test('we can alias two existing identified users', async () => {
```
hazzadous, line 1508: Update to verify create_alias doesn't merge in this case

src/worker/ingestion/process-event.ts
hazzadous: I wasn't sure that "only anonymous users get merged" was a requirement but will update accordingly.

src/worker/ingestion/process-event.ts
yakkomajuri: also think alias should be left alone. It's an explicit choice to merge.

tests/shared/process-event.ts
```diff
@@ -1358,6 +1421,316 @@ export const createProcessEventTests = (
...
+            // depending on specific internals, it would be ideal if we could avoid
+            // this but I'm not sure how right now.
+            const originalCreatePerson = hub.db.createPerson.bind(hub.db)
+            const createPersonMock = jest.fn(async (...args) => {
+                // eslint-disable-next-line
+                // @ts-ignore
+                const result = await originalCreatePerson(...args)
+                if (createPersonMock.mock.calls.length === 1) {
+                    // On second invocation, make another identify call
+                    await identify(hub, newDistinctId)
```
hazzadous, line 1545: USe initial distinct Id

tests/shared/process-event.ts
```diff
@@ -1358,6 +1421,316 @@ export const createProcessEventTests = (
...
+
+        //     // Now let's make sure there are no persons. In a production environment
+        //     // these identify tasks should be retried, or send to a "dead letter
+        //     // queue" or similar for inspection.
+        //     expect(await hub.db.fetchPersons()).toEqual([])
+        // })
+    })
+
+    describe('when handling $create_alias', () => {
+        test('we do not alias two existing identified persons', async () => {
```
hazzadous, line 1593: Update to allow aliasing of identified persons

yakkomajuri:
I'm taking over here. Also making a judgment call that `$create_alias` **should** allow merging identified people.

I believe that is an explicit decision, and there's no other scenario I can really see creating an alias being used for beyond identified -> identified. 

This is also important to us as we allow people to merge whoever they want on the interface, and so it's not up to us to tell users who they can and cannot merge **manually and after the fact**. That merging is done via `$create_alias`

yakkomajuri:
I started messing with this but realized I'd end up with some not-so-nice conflicts with something else I'm doing. Will join the work up

PR #564, opened 2021-09-13
Disallow generic ids

yakkomajuri:
## Changes

We shouldn't allow people to identify users with generic IDs as this can lead to massive problems. 

## Checklist

-   [ ] Jest tests

### DIFF
--- a/src/worker/ingestion/process-event.ts
+++ b/src/worker/ingestion/process-event.ts
@@ -39,29 +39,24 @@ const EVENTS_WITHOUT_EVENT_DEFINITION = ['$$plugin_metrics']
 
 // used to prevent identify from being used with generic IDs
 // that we can safely assume stem from a bug or mistake
-const DISALLOWED_IDS = new Set([
+const CASE_INSENSITIVE_ILLEGAL_IDS = new Set([
     'anonymous',
     'guest',
-    'not_authenticated',
-    'distinctId',
+    'distinctid',
     'distinct_id',
     'id',
-    'ID',
-    '',
-    ' ',
-    0,
-    '[object Object]',
-    'Email',
+    'not_authenticated',
     'email',
     'undefined',
-    'null',
     'true',
     'false',
-    'NaN',
-    'Infinity',
-    'None',
-    'none',
 ])
+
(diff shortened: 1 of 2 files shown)


timgl:
Test? Also I think this is something we should add to the client libraries + documentation otherwise this will cause a lot of confusion for users.

src/worker/ingestion/process-event.ts
```diff
@@ -37,6 +37,32 @@ const MAX_FAILED_PERSON_MERGE_ATTEMPTS = 3
...
+const DISALLOWED_IDS = new Set([
+    'anonymous',
+    'guest',
+    'not_authenticated',
+    'distinctId',
+    'distinct_id',
+    'id',
+    'ID',
+    '',
+    ' ',
```
Twixes, line 51: Maybe we should just disallow any all-whitespace IDs?

src/worker/ingestion/process-event.ts
```diff
@@ -37,6 +37,32 @@ const MAX_FAILED_PERSON_MERGE_ATTEMPTS = 3
...
+    0,
```
Twixes, line 52: If the type of `distinctId` is right in `EventsProcessor` – `string`, then this will never match, as `0` is a `number`.

yakkomajuri: ah yes sorry accidental typo

src/worker/ingestion/process-event.ts
```diff
@@ -189,6 +210,9 @@ export class EventsProcessor {
         distinctId: string,
         teamId: number
     ): Promise<void> {
+        if (isDistinctIdIllegal(distinctId)) {
+            return
```
Twixes, line 214: Can we also add `statsd` instrumentation (sending `distinctId` too) here to have at least an internal metric of the scale of the issue?

yakkomajuri:
@Twixes how's this lookin'

yakkomajuri:
I'll add Docs now btw

yakkomajuri:
Docs are ready https://github.com/PostHog/posthog.com/pull/2001

Twixes (approved):
Lookin' good

PR #570, opened 2021-09-14
Add VS Code debugger configurations

Twixes:
## Changes

Needed to debug various things while working on #569 and decided to try out VS Code's capabilities of integrating with Node.js, turns out this is works super nicely. Made two launch configs: Postgres and ClickHouse.

How to use:


yakkomajuri (approved):
haven't tested but should be good to get in. cheers for this

PR #576, opened 2021-09-16
Safer people merging

yakkomajuri:
## Changes

~~**A LOT** more info coming soon, hold tight~~

Still some things to fix here.

Also thanks a lot to @hazzadous for the initial work here and a very solid foundation for the bulk of the tests here. The tests you wrote were very good.

So this PR batches 2 changes in 1. This is unfortunate but was a decision made to ship faster, given they both made significant changes to the same area of the codebase, looking to fix the same issue, just taking different approaches.

The 2 changes are:

1. A refactor of `mergePeople`. Doing away with the sketchy `while (true)` loop and logic that was very prone to leaving things out of sync. The big problem here is that we were enqueueing kafka messages for updates to CH directly at each update to Postgres, which is _bad_. Really we want to only enqueue all the messages when we know everything we should do in Postgres has succeeded, which we can accomplish with a transaction.
2. Fixing https://github.com/PostHog/posthog/issues/5527 by not allowing a merge from an identified user.

1 is reasonably self-explanatory. 2 needs some context and I had to think quite a bit about how to conceptually approach it.

### $create_alias

This has been left alone. If someone explicitly asks us to merge 2 users, we _will_ do it. It's much harder to shoot yourself in the foot here, and we shouldn't infer for users that they don't want to merge 2 users they explicitly chose to merge. Plus `$create_alias` underpins our person merge modal, where users can manually pick in a very explicit way what users they want to merge into one. We should let them do that, identified or not.

### $identify

Here's where things get tricky. Here's a table explaining all the paths and their results:

> Note: Distinct ID 1 refers to the person we're merging **from**, and Distinct ID 2 is the person we're merging **into**

| Distinct ID 1 exists | Person for Distinct ID 1 is identified | Distinct ID 2 exists | Person for Distinct ID 2 is identified  | Outcome |
| :-------------: | :-------------: | :-------------: | :-------------: | :-------------: |
| 🚫 | 🚫 | 🚫 | 🚫 | Create person with both distinct IDs |
| 🚫 | 🚫 | ✅  | 🚫 | Add Distinct ID 1 to the person for Distinct ID 2 |
| 🚫 | 🚫 | ✅  | ✅  | Add Distinct ID 1 to the person for Distinct ID 2 |
| ✅ | 🚫 | 🚫 | 🚫 | Add Distinct ID 2 to the person for Distinct ID 1 |
| ✅ | ✅ | 🚫 | 🚫 | Add Distinct ID 2 to the person for Distinct ID 1 |
| ✅ | 🚫 | ✅  | 🚫 | Merge the two people |
| ✅ | 🚫 | ✅  | ✅ | Merge the person for distinct ID 1 (anonymous) into the one for distinct ID 2 (identified) |
| ✅ | ✅ | ✅  | ✅ | Do not merge!! Leave the 2 people alone. |
| ✅ | ✅ | ✅  | 🚫 |  Do not merge!! We will not merge an identified person into an anonymous one |

## Checklist

-   [ ] Jest tests

### DIFF
--- a/src/worker/ingestion/process-event.ts
+++ b/src/worker/ingestion/process-event.ts
@@ -313,8 +313,7 @@ export class EventsProcessor {
             try {
                 await this.db.addDistinctId(oldPerson, distinctId)
                 // Catch race case when somebody already added this distinct_id between .get and .addDistinctId
-            } catch (error) {
-                Sentry.captureException(error)
+            } catch {
                 // integrity error
                 if (retryIfFailed) {
                     // run everything again to merge the users if needed
@@ -328,8 +327,7 @@ export class EventsProcessor {
             try {
                 await this.db.addDistinctId(newPerson, previousDistinctId)
                 // Catch race case when somebody already added this distinct_id between .get and .addDistinctId
-            } catch (error) {
-                Sentry.captureException(error)
+            } catch {
                 // integrity error
                 if (retryIfFailed) {
                     // run everything again to merge the users if needed
@@ -345,8 +343,7 @@ export class EventsProcessor {
                     distinctId,
                     previousDistinctId,
                 ])
-            } catch (error) {
-                Sentry.captureException(error)
+            } catch {
                 // Catch race condition where in between getting and creating,
                 // another request already created this person
                 if (retryIfFailed) {
--- a/src/worker/ingestion/process-event.ts
+++ b/src/worker/ingestion/process-event.ts
@@ -38,11 +38,6 @@ const MAX_FAILED_PERSON_MERGE_ATTEMPTS = 3
 // for e.g. internal events we don't want to be available for users in the UI
 const EVENTS_WITHOUT_EVENT_DEFINITION = ['$$plugin_metrics']
 
-export enum AliasTriggerEvent {
-    Identify = '$identify',
-    CreateAlias = '$create_alias',
-}
-
 // used to prevent identify from being used with generic IDs
 // that we can safely assume stem from a bug or mistake
 const CASE_INSENSITIVE_ILLEGAL_IDS = new Set([
@@ -321,7 +316,7 @@ export class EventsProcessor {
             return
         }
         if (event === '$create_alias') {
-            await this.alias(properties['alias'], distinctId, teamId)
+            await this.alias(properties['alias'], distinctId, teamId, false)
         } else if (event === '$identify') {
             if (properties['$anon_distinct_id']) {
                 await this.alias(properties['$anon_distinct_id'], distinctId, teamId)
@@ -334,7 +329,7 @@ export class EventsProcessor {
         previousDistinctId: string,
         distinctId: string,
         teamId: number,
-        aliasTriggerEvent = AliasTriggerEvent.Identify,
+        shouldIdentifyPerson = true,
         retryIfFailed = true,
         totalMergeAttempts = 0
     ): Promise<void> {
@@ -349,7 +344,7 @@ export class EventsProcessor {
                 // integrity error
                 if (retryIfFailed) {
                     // run everything again to merge the users if needed
-                    await this.alias(previousDistinctId, distinctId, teamId, aliasTriggerEvent, false)
+                    await this.alias(previousDistinctId, distinctId, teamId, shouldIdentifyPerson, false)
                 }
             }
             return
@@ -363,7 +358,7 @@ export class EventsProcessor {
                 // integrity error
                 if (retryIfFailed) {
                     // run everything again to merge the users if needed
-                    await this.alias(previousDistinctId, distinctId, teamId, aliasTriggerEvent, false)
+                    await this.alias(previousDistinctId, distinctId, teamId, shouldIdentifyPerson, false)
                 }
             }
             return
@@ -385,7 +380,7 @@ export class EventsProcessor {
                 // another request already created this person
                 if (retryIfFailed) {
                     // Try once more, probably one of the two persons exists now
-                    await this.alias(previousDistinctId, distinctId, teamId, aliasTriggerEvent, false)
+                    await this.alias(previousDistinctId, distinctId, teamId, shouldIdentifyPerson, false)
                 }
             }
             return
@@ -394,15 +389,14 @@ export class EventsProcessor {
         if (oldPerson && newPerson && oldPerson.id !== newPerson.id) {
             // $create_alias is an explicit call to merge 2 users, so we'll merge anything
             // for $identify, we'll not merge a user who's already identified into anyone else
-            const isIdentifyCallToMergeAnIdentifiedUser =
-                aliasTriggerEvent === AliasTriggerEvent.Identify && oldPerson.is_identified
+            const isIdentifyCallToMergeAnIdentifiedUser = shouldIdentifyPerson && oldPerson.is_identified
 
             if (isIdentifyCallToMergeAnIdentifiedUser) {
                 status.warn('🤔', 'refused to merge an already identified user via an $identify call')
             } else {
                 await this.mergePeople({
                     totalMergeAttempts,
-                    aliasTriggerEvent,
+                    shouldIdentifyPerson,
                     mergeInto: newPerson,
                     mergeIntoDistinctId: distinctId,
                     otherPerson: oldPerson,
@@ -418,14 +412,14 @@ export class EventsProcessor {
         otherPerson,
         otherPersonDistinctId,
         totalMergeAttempts = 0,
-        aliasTriggerEvent = AliasTriggerEvent.Identify,
+        shouldIdentifyPerson = true,
     }: {
         mergeInto: Person
         mergeIntoDistinctId: string
         otherPerson: Person
         otherPersonDistinctId: string
         totalMergeAttempts: number
-        aliasTriggerEvent?: AliasTriggerEvent
+        shouldIdentifyPerson?: boolean
     }): Promise<void> {
         const teamId = mergeInto.team_id
 
@@ -489,7 +483,7 @@ export class EventsProcessor {
                     otherPersonDistinctId,
                     mergeIntoDistinctId,
                     teamId,
-                    aliasTriggerEvent,
+                    shouldIdentifyPerson,
                     false,
                     failedAttempts
                 )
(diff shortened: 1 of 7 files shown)


tests/shared/process-event.ts
```diff
@@ -47,7 +48,31 @@ async function createPerson(
...
+                distinctIds,
+                (events as Event[])
+                    .filter((event) => distinctIds.includes(event.distinct_id))
+                    .map((event) => event.event),
+            ] as const
+        })
+    )
+}
+
+export const createIdentifiedPerson = async (hub: Hub, teamId: number, distinctId: string): Promise<void> => {
```
hazzadous, line 73: Is there any reason to reach into the internals here? It would seem that we can get to the same state just with a sequence of events and avoid artificially updating the db.

src/worker/ingestion/process-event.ts
```diff
@@ -299,10 +264,51 @@ export class EventsProcessor {
...
+            }
+            await this.setIsIdentified(teamId, distinctId)
+        }
+    }
+
     private async alias(
         previousDistinctId: string,
         distinctId: string,
         teamId: number,
+        aliasTriggerEvent = AliasTriggerEvent.Identify,
```
hazzadous, line 311: It seems strange to require `alias` to understand how it was triggered. If we add another way to trigger it, does this mean we would have to update the enum and decide on the desired behaviour. Might it be better to just name it something like `shouldBeMergeIdentifiedPersons` ?

tests/shared/process-event.ts
yakkomajuri: The description is coming soon and I was going to comment on this specific line.  

The tests you wrote were an amazing foundation here, so big thanks for that.

However, there was a conceptual mistake with the $identify tests.

The following:

```js
posthog.identify('first_id')
posthog.capture('some event')
posthog.identify('second_id')
posthog.capture('some other event')
```

Leads to and should lead to **one** person in the DB. Reason being that we've never seen `second_id` before, so we actually just pop that ID into the person from `first_id`, we never actually merge _persons_.

So to actually test that we're not allowing the merging of persons here, we need a person to exist, and it's just cleaner/easier to tap into the internals for it.

I completely get the sentiment of mimicking the libs here, but this does in fact do that. It works with the (very reasonable) assumption that by the time this happens:

```js
posthog.identify('first_id')
posthog.capture('some event')
posthog.identify('second_id')
posthog.capture('some other event')
```

somewhere else `second_id` has already been created. The assumption is actually more reasonable than doing it all without internals because it's much more likely that someone ends up trying to merge into a person that already exists than they create 2 new persons in a session (this would require some odd circumstances).

Hope this makes sense?

src/worker/ingestion/process-event.ts
yakkomajuri: Yup, that's valid. Originally had a boolean here but thought this was more explicit. Your comment makes sense though

src/worker/ingestion/process-event.ts
```diff
@@ -389,63 +412,68 @@ export class EventsProcessor {
...
         // – in such a case a distinct ID may be assigned to the person in the database
         // AFTER `otherPersonDistinctIds` was fetched, so this function is not aware of it and doesn't merge it.
-        // That then causeds `deletePerson` to fail, because of foreign key constraints –
+        // That then causes `deletePerson` to fail, because of foreign key constraints –
         // the dangling distinct ID added elsewhere prevents the person from being deleted!
         // This is low-probability so likely won't occur on second retry of this block.
         // In the rare case of the person changing VERY often however, it may happen even a few times,
         // in which case we'll bail and rethrow the error.
-        while (true) {
+        await this.db.postgresTransaction(async (client) => {
```
hazzadous, line 429: Is it possible we can put this around the whole event handling code instead? i.e. we really want handleIdentify and handleAlias to be atomic

yakkomajuri: It's reasonable to consider expanding how much the transaction wraps in a further PR. 

However, I'd push back against wrapping the entire event handling code on it. We can discuss this further elsewhere as it is a bit outside the scope here but for one @fuziontech was already raising concerns about performance and Postgres becoming a bottleneck if we use transactions too liberally. 

Let's have this discussion elsewhere though as I am indeed planning on pushing more refactors, and we should:

a) ship this PR fast
b) it already includes quite a few changes

hazzadous: Yup, for sure it's an improvement already, happy to get in without.

@fuziontech do we have any stats on the number of open transactions we have at the moment? I suspect even wrapping all event handlers in a transactions they will still be very sort lived. If we want to avoid transactions then maybe we can refactor to not require transactions but still maintain integrity?

tests/shared/process-event.ts
hazzadous: If it's not a valid sequence of events, then I would suggest we get rid and replace with one that is valid.

Point being, it's very easy to test how the system behaves in response to events, and these are going to be pretty stable and representative of actual interactions (as this is what we tell others to use), where as reaching in we are making it harder to refactor.

yakkomajuri: Right, but the thing is - some scenarios are more likely given the current state of our data than through a sequence in a session.

I'm happy to make some changes here but ultimately I see it as valid to have the DB in a given state and trigger events from there to see how we react given the current state.

And what we generally see is:

User A opens up their app and a person is created for them.

User B opens up their app and a person is created for them, which is then asked to be merged into User A.

These are not a sequence of events happening at once.

yakkomajuri: But yes indeed they are a sequence of events - they're just not very compatible with the stateful mock client that's there, but I can certainly re-write/circumvent that too

hazzadous: I think where ever we can do a valid sequence of events in and validate that the state of the system is correct that should be the preference. For the case you're mentioned, that seems like you've basically got a valid sequence of events there.

I guess there are cases where we can't get the db into a specific state through events so I guess in these cases we don't have much choice.

yakkomajuri: updated to reflect this - thanks for pushing me here!

yakkomajuri:
sorry for the commit spam here - CH tests take **extremely long** for me locally so it's faster to push and run them remotely.

hazzadous:
> sorry for the commit spam here - CH tests take **extremely long** for me locally so it's faster to push and run them remotely.

Hah, are you using a remove CH?

fuziontech:
you should rebase and try running them on a big CodeSpaces instance 😉

timgl:
Re tests being slow, one change that I made on the main posthog repo is to _not_ empty/truncate clickhouse tables between every test run, and instead just filter every database call by team ID. That cut execution time in half so might be worth trying here.

yakkomajuri:
Tests cover this but if you want to "see it with your own eyes"

https://user-images.githubusercontent.com/38760734/133766133-7233aa60-642c-48e5-8da9-a2c50c18a1df.mov

hazzadous (approved):
Really handy to see the `posthog.capture("$identify", ...)` trick in the vid.

Looks like there are plenty of other issues regarding, atomicity of postgres changes, error handling for, e.g. if we fail to send messages to kafka do we retry processing the event etc. but at least gets the identify issue mostly out of the way.

yakkomajuri:
yeah that merge people test seems like it was already a bit flaky before, passed locally before pushing, will fix

src/worker/ingestion/process-event.ts
```diff
@@ -324,10 +281,55 @@ export class EventsProcessor {
...
+                // another request already created this person
+                personFound = await this.db.fetchPerson(teamId, distinctId)
+            }
+        }
+        if (personFound && !personFound.is_identified) {
+            await this.db.updatePerson(personFound, { is_identified: isIdentified })
+        }
+    }
+
+    private async handleIdentifyOrAlias(
```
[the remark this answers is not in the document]
yakkomajuri: potentially yes. thought the same, will leave for another refactor

src/worker/ingestion/process-event.ts
```diff
@@ -324,10 +281,55 @@ export class EventsProcessor {
...
+        properties: Properties,
+        distinctId: string,
+        teamId: number
+    ): Promise<void> {
+        if (isDistinctIdIllegal(distinctId)) {
+            this.pluginsServer.statsd?.increment(`illegal_distinct_ids.total`, { distinctId })
+            return
+        }
+        if (event === '$create_alias') {
+            await this.alias(properties['alias'], distinctId, teamId, false)
```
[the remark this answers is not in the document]
yakkomajuri: ah yes there's a simple way to do this, will update

src/worker/ingestion/process-event.ts
```diff
@@ -324,10 +281,55 @@ export class EventsProcessor {
...
+                    [distinctId]
+                )
+            } catch {
+                // Catch race condition where in between getting and creating,
+                // another request already created this person
+                personFound = await this.db.fetchPerson(teamId, distinctId)
+            }
+        }
+        if (personFound && !personFound.is_identified) {
+            await this.db.updatePerson(personFound, { is_identified: isIdentified })
```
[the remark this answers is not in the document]
yakkomajuri: this function was just moved on this PR, not updated/changed.

good point though, but out of scope here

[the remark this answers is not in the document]
tiina303: Agreed it's not part of this, could you put up a follow-up PR for this afterwards?

src/worker/ingestion/process-event.ts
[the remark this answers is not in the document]
tiina303: & thoughts on the testing part?

> Maybe it would be good if we had tests for all 8 cases that you outlined awesomely for both functions (create_alias and identify) separately & checking that we end up with the right people & right identified state. It will be a bit of duplication in tests, but I think it'll be harder to break things in the future.

yakkomajuri:
Friday evening, I've been working to get tests to pass and this happens:

Screenshot 2021-09-17 at 14 47 48

I hate life

fuziontech:
NoooOOOOOOoooooo!

yakkomajuri:
Updated the query counter. That's a horrible test because it will flake with all sorts of changes. I'll get rid of it somewhere else. This PR has shown me I really need to spend some time refactoring plugin server stuff

PR #578, opened 2021-09-16
truncate test tables

yakkomajuri:
- truncate test tables instead of delete
- also truncate ee_hook

## Changes

_Please describe._

## Checklist

-   [ ] Jest tests


Twixes:
Curious why?

yakkomajuri:
gotta handle foreign key constraints actually.

@Twixes maybe a minor improvement given the test db is not massive but this should be faster.

hazzadous (approved):

PR #579, opened 2021-09-17
do not swallow errors on setupPlugin

yakkomajuri:
## Changes

_Please describe._

we were swallowing errors on setup fails

lets not do that :D

## Checklist

-   [ ] Jest tests


mariusandra (approved):
Looks as good as the PR description reads.

PR #582, opened 2021-09-20
chore(test): reduce the test timeouts to make tests fail faster

hazzadous:
There are some really large timeouts in here, which makes tests appear
to hang. It's not a wonderful dev experience. It looks to me like these
tests should be running much faster than the specified timeouts.


guidoiaquinti:
Hey @hazzadous , is this PR still relevant/in progress? Any chance we can push it to the finish line or close it? 

Thank you! 🙇

PR #585, opened 2021-09-22
Split identification flow

yakkomajuri:
## Changes

Splits identifying a person away from `updatePerson` and makes it more robust. Now we don't update Kafka if we don't need to and conceptually it makes sense to have this separate.

This will also make even more sense once I completely re-do `updatePerson`, as I want to have that focus solely on properties.

## Checklist

-   [ ] Jest tests

### DIFF
--- a/src/utils/db/db.ts
+++ b/src/utils/db/db.ts
@@ -496,6 +496,50 @@ export class DB {
         return person
     }
 
+    public async setPersonAsIdentified(person: Person): Promise<void> {
+        let shouldUpdateClickhouse = false
+
+        await this.postgresTransaction(async (client) => {
+            const result = await client.query(
+                `
+                UPDATE posthog_person 
+                SET is_identified=$1 
+                WHERE id = $2 
+                AND is_identified = $3
+                RETURNING is_identified
+            `,
+                [true, person.id, false]
+            )
+
+            if (result.rows.length > 0) {
+                shouldUpdateClickhouse = true
+            }
+        })
+
+        if (this.kafkaProducer && shouldUpdateClickhouse) {
+            const message = {
+                topic: KAFKA_PERSON,
+                messages: [
+                    {
+                        value: Buffer.from(
+                            JSON.stringify({
+                                created_at: castTimestampOrNow(
+                                    person.created_at,
+                                    TimestampFormat.ClickHouseSecondPrecision
+                                ),
+                                properties: JSON.stringify(person.properties),
+                                team_id: person.team_id,
+                                is_identified: true,
+                                id: person.uuid,
+                            })
+                        ),
+                    },
+                ],
+            }
+            await this.kafkaProducer.queueMessage(message)
+        }
+    }
+
     public async updatePerson(person: Person, update: Partial<Person>, client: PoolClient): Promise<ProducerRecord[]>
     public async updatePerson(person: Person, update: Partial<Person>): Promise<Person>
     public async updatePerson(
--- a/package.json
+++ b/package.json
@@ -6,9 +6,9 @@
     "main": "dist/index.js",
     "scripts": {
         "test": "jest --runInBand --forceExit tests/**/*.test.ts",
-        "test:postgres:1": "jest --runInBand --forceExit tests/*.test.ts",
+        "test:postgres:1": "jest --runInBand --forceExit tests/*.test.ts tests/shared/db.test.ts",
         "test:postgres:2": "jest --runInBand --forceExit tests/postgres/*.test.ts",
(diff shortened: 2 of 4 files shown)


yakkomajuri:
Unrelated test failures. See #586

src/utils/db/db.ts
```diff
@@ -496,6 +496,49 @@ export class DB {
...
+                AND is_identified = $3
+                RETURNING is_identified
+            `,
+            [true, person.id, false],
+            'setPersonAsIdentified'
+        )
+
+        if (result.rows.length > 0) {
+            shouldUpdateClickhouse = true
+        }
```
tiina303, line 516: nit in python I'd do (& drop it from the top) not sure if valid ts
```
let shouldUpdateClickhouse = result.rows.length > 0
```

src/utils/db/db.ts
```diff
@@ -496,6 +496,49 @@ export class DB {
...
+
+        if (this.kafkaProducer && shouldUpdateClickhouse) {
```
tiina303, line 518: What if we crashed just before this line last time, how can I now setIdentified for the clickhouse person (the postgres person is already identified, so calling twice won't help) && in general what happens if we have people identified in Postgres but not in CH?

src/utils/db/db.ts
```diff
@@ -496,6 +496,49 @@ export class DB {
...
+                                ),
+                                properties: JSON.stringify(person.properties),
+                                team_id: person.team_id,
+                                is_identified: true,
+                                id: person.uuid,
+                            })
+                        ),
+                    },
+                ],
+            }
```
tiina303, line 537: I assume you'll deal with this duplication as you 
> This will also make even more sense once I completely re-do updatePerson, as I want to have that focus solely on properties.

tiina303 (approved):

src/utils/db/db.ts
yakkomajuri: ah yes ofc this lingered from before I had the const

src/utils/db/db.ts
yakkomajuri: I have an idea

yakkomajuri: Wrapped it in a transaction, so that if we don't queue the message in Kafka, we'll rollback on Postgres too, making sure we keep both in the same state and a subsequent identify call will get them both back where they should be.

src/utils/db/db.ts
yakkomajuri: addressed, cheers

src/utils/db/db.ts
```diff
@@ -496,6 +496,44 @@ export class DB {
         return person
     }
 
+    public async setPersonAsIdentified(person: Person): Promise<void> {
+        await this.postgresTransaction(async (client) => {
```
tiina303, line 500: Not sure if this is the right tradeoff here:
problems:
* we're more likely to now not write to Postgres (e.g. Kafka is down, or there was a problem with the write), which we treat as source of truth for people stuff & where we can recover stuff from (because everything is async user called identify and assumed things are good, but nothing happened is more likely now)
* number of active transactions increases (perf decrease)

How immediately do Postgres & ClickHouse need to be in sync here? An alternative to consider could be to have some cron job after some delay (significantly bigger than Kafka normal delay), that updates the identified for CH (in case we failed here). It's quite easy as we only ever set it to true (so we know exactly when it needs to happen & don't need to worry about anyone overriding the value to something else).

yakkomajuri: Honestly, I don't think this is that important to warrant a cron job.

The chances of us failing in between those lines is minimal and very low-impact. So I'd just advocate for reverting the transaction change and leaving as is.

tiina303: sound good to me.

yakkomajuri: After thinking for a while I went back to prioritizing integrity (even with low probability scenarios) over performance here.

We do a check not long before this runs to see if the person was already identified, so we also shouldn't be doubling too many messages to Kafka. Doesn't hurt to be cautious - thanks for pushing here.

src/utils/db/db.ts
```diff
@@ -496,6 +496,43 @@ export class DB {
...
     }
 
+    public async setPersonAsIdentified(person: Person): Promise<void> {
+        const result = await this.postgresQuery(
+            `
+                UPDATE posthog_person 
+                SET is_identified=$1 
+                WHERE id = $2 
+                AND is_identified = $3
+                RETURNING is_identified
```
macobo, line 506: Returning result is not used

PR #586, opened 2021-09-22
fix tests after posthog#5976

yakkomajuri:
## Changes

_Please describe._

posthog#5976

## Checklist

-   [ ] Jest tests


yakkomajuri:
Actually this is not only a tests fix anymore, it's an important fix for `createPerson`

PR #589, opened 2021-09-24
Adjust selector matching for property values being in an array

Twixes:
## Changes

Fixes https://github.com/PostHog/plugin-server/issues/566 for good. Property filters save values into an array these days, which was accounted for in other matching paths, but not in selector matching.

## Checklist

-   [x] Jest tests


yakkomajuri (approved):
thanks!

PR #591, opened 2021-09-28
Set isIdentified only when aliasing actually occurs

macobo:
## Changes

Closes https://github.com/PostHog/posthog/issues/6116 by not setting `is_identified` when not updating identities (e.g. when updating properties)

This was causing issues after https://github.com/PostHog/plugin-server/pull/576 as users who updated props on anonymous users would run into issues when logging in/signing up creating new users.

## Checklist

-   [x] Jest tests

### DIFF
--- a/src/worker/ingestion/process-event.ts
+++ b/src/worker/ingestion/process-event.ts
@@ -317,11 +317,8 @@ export class EventsProcessor {
         }
         if (event === '$create_alias') {
             await this.alias(properties['alias'], distinctId, teamId, false)
-        } else if (event === '$identify') {
-            if (properties['$anon_distinct_id']) {
-                await this.alias(properties['$anon_distinct_id'], distinctId, teamId)
-            }
-            await this.setIsIdentified(teamId, distinctId)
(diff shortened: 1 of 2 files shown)


macobo:
We only rely on this in the app and it seems we just change the UI a bit based on is_identified in person view. I _think_ this is fine - we expect distinct_id to change when people log in/sign in posthog-js, but not 100% sure on e.g. server-side libraries. But it's a minor enough thing we can remedy later if needed.

yakkomajuri:
Yeah it's not a very important setting

hazzadous (approved):
Looks ok, does this introduce a race between merging and setting not happening atomically?

Minor: perhaps person2distinctid would be a more obvious name?

macobo:
Sorry @hazzadous for not waiting for your review

> Looks ok, does this introduce a race between merging and setting not happening atomically?

No more than it did previously - note the call happens in the same place in most of the cases, except not when no anon_distinct_id or the two distinct_ids are the same.

PR #597, opened 2021-10-12
WIP create person with properties early when processing event

yakkomajuri:
## Changes

Playing with the idea of removing all the `createPerson` calls deeper down in `process-event` in favor of creating a person with some initial properties early on to make sure it exists.

Putting it out as a WIP PR just to run the entire test stack more easily

## Checklist

-   [ ] Jest tests


tiina303 (changes requested):
This looks great, the code looks simpler and clearer. I'm not sure on the perf cost of the merge, but we definitely need to ship the changes to how we update props & merge people before this pr

tiina303:
will break into small prs in the merged repo

PR #601, opened 2021-10-15
Add simple ingestion support for groups

macobo:
## Changes

Initial support for groups added here. Very much inspired by the original PR.

Some gotchas: 
- Group rows are created on $groupidentify events, not when $group keys are passed. We'll iterate this as needed.
- Depends on https://github.com/PostHog/posthog/pull/6462 for tests as well

## Checklist

-   [x] Jest tests

### DIFF
--- a/src/utils/db/db.ts
+++ b/src/utils/db/db.ts
@@ -1170,12 +1170,19 @@ export class DB {
         }
 
         const insertGroupTypeResult = await this.postgresQuery(
-            `INSERT INTO posthog_grouptypemapping (group_type, group_type_index, team_id)
-            VALUES ($1, $2, $3)
-            ON CONFLICT DO NOTHING
-            RETURNING group_type_index`,
-            [groupType, index, teamId],
-            'upsertGroupType'
+            `
+            WITH insert_result AS (
+                INSERT INTO posthog_grouptypemapping (team_id, group_type, group_type_index)
+                VALUES ($1, $2, $3)
+                ON CONFLICT DO NOTHING
+                RETURNING group_type_index
+            )
+            SELECT * FROM insert_result
+            UNION
+            SELECT group_type_index FROM posthog_grouptypemapping WHERE team_id = $1 AND group_type = $2;
+            `,
+            [teamId, groupType, index],
+            'insertGroupType'
         )
 
         if (insertGroupTypeResult.rows.length == 0) {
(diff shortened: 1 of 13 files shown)


yakkomajuri (commented):
initial first pass

src/worker/ingestion/group-type-manager.ts
```diff
@@ -0,0 +1,44 @@
...
+        try {
+            const teamGroupTypes: GroupTypeToColumnIndex = await this.db.fetchGroupTypes(teamId)
+            this.groupTypesCache.set(teamId, [teamGroupTypes, Date.now()])
+            return teamGroupTypes
+        } finally {
+            clearTimeout(timeout)
+        }
+    }
+
+    public async fetchGroupTypeIndex(teamId: TeamId, groupType: string): Promise<number | null> {
```
yakkomajuri, line 31: nit: this inserts, not only fetches

macobo:
@yakkomajuri ready for another pass

src/utils/db/db.ts
```diff
@@ -1141,4 +1147,85 @@ export class DB {
...
+                messages: [
+                    {
+                        value: Buffer.from(
+                            JSON.stringify({
+                                group_type_index: groupTypeIndex,
+                                group_key: groupKey,
+                                team_id: teamId,
+                                group_properties: JSON.stringify(properties),
+                                created_at: castTimestampOrNow(
+                                    DateTime.utc(),
```
yakkomajuri, line 1213: nit and not important to fix now but if we're doing this users will end up seeing wrong created_at dates, given they will change with each update, and rows will be collapsed so we can't do `argMin`.

not important to fix, just leaving the comment

src/utils/db/db.ts
```diff
@@ -1141,4 +1147,85 @@ export class DB {
...
+
+        return result
+    }
+
+    public async insertGroupType(teamId: TeamId, groupType: string, index: number): Promise<number | null> {
+        if (index >= this.MAX_GROUP_TYPES_PER_TEAM) {
+            return null
+        }
+
+        const insertGroupTypeResult = await this.postgresQuery(
```
yakkomajuri, line 1172: nice

yakkomajuri (approved):
some tests seem to be failing. but, given our conversation today about approaching this iteratively, lgtm

PR #607, opened 2021-10-20
Refactor Kafka person update message creation into a function

tiina303:
## Changes

Refactoring, this will make it easier to later change the message to have timestamps and last operation.
2 no-op changes.

I'm assuming we don't need to provide the `is_deleted: 0` at all if it's 0, which I assume is the default?

Relying on unittests didn't run anything locally, should I?

## Checklist

-   [ ] Jest tests


yakkomajuri (approved):
lgtm

PR #608, opened 2021-10-21
Remove unused person creation path

tiina303:
## Changes

Goal: making future changes / refactor easier.
We always have a person at this point, because the private `updatePersonProperties` is only called in `capture` where we have `createPersonIfDistinctIdIsNew` beforehand.

Same story for `setIdentified` at that point we should have a person already and if we don't throwing might be better than creating a person with no props nor other info, to make debugging easier.

## Checklist

-   [ ] Jest tests

### DIFF
--- a/src/worker/ingestion/process-event.ts
+++ b/src/worker/ingestion/process-event.ts
@@ -211,28 +211,9 @@ export class EventsProcessor {
         properties: Properties,
         propertiesOnce: Properties
     ): Promise<Properties> {
-        let personFound = await this.db.fetchPerson(teamId, distinctId)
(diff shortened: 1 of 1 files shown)


tiina303:
or is that there because someone could delete the distinct ID completely in the middle of the creation and update call?
Based on my reading of the code delete only happens in a transaction, where we move the distinct ID first. Can a distinct ID be completely deleted somehow? What happens if someone deletes a person in the UI?

yakkomajuri (approved):
Yeah, I think this is fine. Let's keep an eye when we merge but indeed _some_ person should exist by the time we get to these calls

PR #609, opened 2021-10-25
Silence distinct ID conflict Sentry errors

yakkomajuri:
## Changes

I added this bit of error logging to help debug some things in the past but now it leads to Sentry spamming due to a very normal event (distinct id conflicts).

https://sentry.io/organizations/posthog/issues/2454986618/?project=5592816&query=is%3Aunresolved&statsPeriod=14d

## Checklist

-   [ ] Jest tests

### DIFF
--- a/src/worker/ingestion/process-event.ts
+++ b/src/worker/ingestion/process-event.ts
@@ -670,7 +670,9 @@ export class EventsProcessor {
             try {
                 await this.db.createPerson(sentAt, {}, teamId, null, false, personUuid.toString(), [distinctId])
             } catch (error) {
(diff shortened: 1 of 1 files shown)


tiina303:
Could you elaborate on why distinct ID conflicts are normal and should be ignored?

yakkomajuri:
More often than not a person's events come in batches. Most people use PostHog on the frontend, and so in a second I'm already getting pageviews, snapshots, etc from a person.

Thus, we're very likely to be processing an event from a person that does not exist in parallel. As such, one thread creates the distinct ID mapping, and the other then fails to insert because of that constraint.

PR #610, opened 2021-10-26
fetchPerson to be able to use in a transaction

tiina303:
## Changes

Need to use fetchPerson in a transaction to be sure the person wasn't merged in the middle of fetching and updating or any other action.

Not sure if we need to `SELECT FOR UPDATE` in this case too or not.

## Checklist

-   [ ] Jest tests


yakkomajuri (approved):

tiina303:
Will move to the new merged repo

PR #611, opened 2021-10-26
Create new update properties function

tiina303:
## Changes

Add function to update user properties predictably, when ingestion can occur in any order.

Next up:
1. gating to be able to roll this out for our team only first

## Checklist

-   [ ] Jest tests

### DIFF
--- a/src/utils/db/db.ts
+++ b/src/utils/db/db.ts
@@ -1,3 +1,4 @@
+import { expressionStatement } from '@babel/types'
 import ClickHouse from '@posthog/clickhouse'
 import { CacheOptions, Properties } from '@posthog/plugin-scaffold'
 import { captureException } from '@sentry/node'
@@ -29,6 +30,7 @@ import {
     Hook,
     Person,
     PersonDistinctId,
+    PersonPropertyUpdateOperation,
     PluginConfig,
     PluginLogEntry,
     PluginLogEntrySource,
@@ -60,6 +62,8 @@ import {
     chainToElements,
     generateKafkaPersonUpdateMessage,
     generatePostgresValuesString,
+    getPersonPropertyUpdateExpressions,
+    getPersonPropertyUpdateValues,
     hashElements,
     timeoutGuard,
     unparsePersonPartial,
@@ -492,21 +496,6 @@ export class DB {
         return person
     }
 
-    private addPropsUpdateValues(
-        properties: Properties,
-        op: PersonPropertyUpdateOperation,
-        values: Array<any>,
-        expressions: Array<string>
-    ) {
-        for (const [key, value] of Object.entries(properties)) {
-            const parsedValue = typeof value === 'string' ? `"${value}"` : value
-            values.push(op, key, parsedValue)
-            expressions.push(
-                `row($${values.length - 2}, $${values.length - 1}, $${values.length}::jsonb)::person_property_update`
-            )
-        }
-    }
-
     public async updatePersonProperties(
         teamId: number,
         distinctId: string,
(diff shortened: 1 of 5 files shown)


src/utils/db/db.ts
```diff
@@ -486,6 +486,68 @@ export class DB {
         return person
     }
 
+    private addPropsUpdateValues(
```
yakkomajuri: let's extract this to utils

src/utils/db/db.ts
```diff
@@ -486,6 +486,68 @@ export class DB {
...
+    public async updatePersonProperties(
+        teamId: number,
+        distinctId: string,
+        properties: Properties,
+        propertiesOnce: Properties,
+        timestamp: DateTime
+    ): Promise<void> {
+        const values: Array<any> = ['<id>', timestamp.toISO()]
+
+        const propertyUpdateExpressionsArr: any[] = []
```
yakkomajuri: I guess this can be called a nit but I'm much more in favor of returning this than modifying in place. Makes things cleaner in my opinion. Leveraging calling by sharing with simpler objects like arrays can just get confusing IMO

src/utils/db/db.ts
```diff
@@ -486,6 +486,68 @@ export class DB {
...
+        const propertyUpdateExpressions = propertyUpdateExpressionsArr.join(',')
+
+        const query = `SELECT update_person_props($1, $2, ARRAY[ ${propertyUpdateExpressions} ] );`
+
+        let props = {}
+        let personFound: Person | undefined
+        await this.postgresTransaction(async (client) => {
+            personFound = await this.fetchPerson(teamId, distinctId)
+            if (!personFound) {
+                // can I throw in a transaction
```
yakkomajuri: sure, yeah. take a look at how the `postgresTransaction` method handles this. IIRC it will catch, log to Sentry, and ROLLBACK.

src/utils/db/db.ts
```diff
@@ -486,6 +486,68 @@ export class DB {
...
+                throw new Error(
+                    `Could not find person with distinct id "${distinctId}" in team "${teamId}" to update props`
+                )
+            }
+            values[0] = personFound.id
+            props = await client.query(query, values)
+        })
+
+        if (this.kafkaProducer && personFound) {
```
yakkomajuri: This is a comment for us, not you specifically, as I contributed to this implementation.

But, effectively, for every event with `$set`, we'll add a new row to CH (which will hopefully be collapsed, but still). Could be worth doing some checking on if we actually updated something before producing to Kafka.

Given you now fetch the person before updating props, should be fine to just check the returned props against the persons props and deciding to issue a CH insert or not

src/utils/db/db.ts
```diff
@@ -486,6 +486,68 @@ export class DB {
...
+            const kafkaMessage = generateKafkaPersonUpdateMessage(
+                timestamp,
+                props, // todo: right type
```
yakkomajuri: what do you mean with right type?

src/utils/db/db.ts
```diff
@@ -486,6 +486,68 @@ export class DB {
...
+                teamId,
+                personFound.is_identified, // is this safe to use this way
```
yakkomajuri: yup

src/utils/db/db.ts
```diff
@@ -486,6 +486,68 @@ export class DB {
...
+            }
+            values[0] = personFound.id
+            props = await client.query(query, values)
+        })
+
+        if (this.kafkaProducer && personFound) {
+            const kafkaMessage = generateKafkaPersonUpdateMessage(
+                timestamp,
+                props, // todo: right type
+                teamId,
```
yakkomajuri: this is ultimately the same thing, but for clarity, let's do `personFound.team_id`. Just so the reader doesn't need to consider where else this ID might be coming from

yakkomajuri (changes requested):
looking good, a few nits.

also, let's have the tests in this PR?

src/utils/db/db.ts
```diff
@@ -492,6 +496,45 @@ export class DB {
...
+            if (!personFound) {
+                throw new Error(
+                    `Could not find person with distinct id "${distinctId}" in team "${teamId}" to update props`
+                )
+            }
+            values[0] = personFound.id
+            props = await client.query(query, values)
+        })
+
+        if (this.kafkaProducer && personFound && personFound.properties != props) {
```
yakkomajuri, line 526: JS is nice but not _that_ nice. This object comparison will always return false.

We probably already have a lib or util somewhere in this codebase to compare objects so let's use that.

Some reading:

https://stackoverflow.com/questions/1068834/object-comparison-in-javascript

yakkomajuri: I think `JSON.stringify` should be enough for us here btw

src/worker/ingestion/process-event.ts
```diff
@@ -254,6 +255,25 @@ export class EventsProcessor {
         return returnedProps
     }
 
+    private sanitizePropertiesOnce(properties: Properties, propertiesOnce: Properties): void {
```
yakkomajuri, line 258: nice

yakkomajuri (changes requested):
some more tests needed but we're moving forward

tiina303:
Will move to the new merged repo

PR #613, opened 2021-10-26
add benchmark for new person props update

yakkomajuri:
## Changes

_Please describe._

Just moving things along / reducing the amount of things @tiina303 has to do/think about.

We don't run these anywhere anymore so it's fine to merge even with errors.

## Checklist

-   [ ] Jest tests


tiina303 (approved):
You can land or you can wait until I've shipped the new update function & then updated this accordingly. Maybe even run the benchmark tests to move things along faster, so I don't need to :)

